PR #3, opened 2018-08-21
Implement support for hydration transformation.

tsroka:


src/main/java/graphql/nadel/Nadel.java
```diff
@@ -97,47 +88,67 @@ public TypeDefinitionRegistry buildRegistry(ServiceDefinition serviceDefinition)
...
+        }
+        return links;
+    }
+
+    private Link createHydrationLink(SchemaNamespace schemaNamespace, FieldDefinitionWithParentType definition,
+                                     InnerServiceHydration hydration) {
+        SchemaNamespace targetService = assertNotNull(this.namespaceByService.get(hydration.getServiceName()));
+        final FieldDefinition targetField = findTargetFieldForHydration(hydration);
+        //TODO: will not work for lists or non nullable types. does braid support this at all?
+        String targetTypeName = TypeInfo.typeInfo(targetField.getType()).getName();
```
tsroka: @andimarek do we need to change braid in order to support lists? or did i miss something?

src/main/java/graphql/nadel/Nadel.java
```diff
@@ -97,47 +88,67 @@ public TypeDefinitionRegistry buildRegistry(ServiceDefinition serviceDefinition)
...
+
+        //TODO: braid does not support multiple arguments, so we just take first
+        final String fromField = hydration.getArguments().entrySet().stream()
+                .map(e -> e.getValue().getInputName())
+                .findFirst()
+                .orElse(definition.field().getName());
+
+        return Link.from(schemaNamespace, definition.parentType(), definition.field().getName(), fromField)
```
tsroka: @andimarek fromField can only be a single field. Should we make changes in braid or maybe restrict DSL so that it only allows passing one field?

src/main/java/graphql/nadel/Parser.java
```diff
@@ -22,8 +25,9 @@ public StitchingDsl parseDSL(String input) {
         CommonTokenStream tokens = new CommonTokenStream(lexer);
         StitchingDSLParser parser = new StitchingDSLParser(tokens);
         parser.removeErrorListeners();
+        parser.addErrorListener(ThrowingErrorListener.INSTANCE);
         parser.getInterpreter().setPredictionMode(PredictionMode.SLL);
-        parser.setErrorHandler(new BailErrorStrategy());
+        parser.setErrorHandler(new DefaultErrorStrategy());
```
tsroka: this will allow to pass through exception from ThrowingErrorListener, which generates detailed info about what went wrong.

src/main/java/graphql/nadel/Nadel.java
andimarek: lets leave it for now for one field and we will improve braid in the future

src/test/groovy/graphql/nadel/NadelTest.groovy
```diff
@@ -63,8 +61,6 @@ class NadelTest extends Specification {
         0 * graphqlRemoteRetriever1.queryGraphQL(*_) >> completedFuture([data: []])
     }
 
-    //fime: get this up and running again once DSL refactoring is over
-    @Ignore
     def "stitching with transformation"() {
```
andimarek: we should also use hydration here

src/main/java/graphql/nadel/TransformationUtils.java
```diff
@@ -0,0 +1,87 @@
...
+import graphql.nadel.dsl.FieldDefinitionWithTransformation;
+import graphql.nadel.dsl.ServiceDefinition;
+import graphql.util.TraversalControl;
+import graphql.util.TraverserContext;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Stack;
+
+public final class TransformationUtils {
```
andimarek: we should mark it as `@Internal`

src/main/java/graphql/nadel/SchemaSourceFactory.java
```diff
@@ -0,0 +1,18 @@
...
+import com.atlassian.braid.SchemaSource;
+import graphql.nadel.dsl.ServiceDefinition;
+import graphql.schema.idl.TypeDefinitionRegistry;
+
+import java.util.List;
+
+/**
+ * Factory for schema source.
+ */
+public interface SchemaSourceFactory {
```
andimarek: I am not sure this should be part of the Nadel API: I think we should hide Braid Impl Details from the client

src/main/java/graphql/nadel/GraphQLRemoteSchemaSourceFactory.java
```diff
@@ -0,0 +1,35 @@
...
+import graphql.schema.idl.TypeDefinitionRegistry;
+
+import java.util.List;
+
+/**
+ * Schema source factory that uses remote graphql endpoint.
+ *
+ * @param <C> type of passed graphql context.
+ */
+public class GraphQLRemoteSchemaSourceFactory<C> implements SchemaSourceFactory {
```
andimarek: same as the other interface: Braid impl details

andimarek: I don't think we should expose braid types to the public API

src/main/java/graphql/nadel/SchemaSourceFactory.java
andimarek: same as above: I don't think we should expose Braid types, rather some Nadel specifc "SchemaSource"

tsroka: I do not agree, if you see Nadel constructor it already takes GraphQLRemoteRetrieverFactory that provides GraphQLRemoteRetriever which is braid type (and it was here already).  It is not enough however for stitching service and we need more general way of creating schema source that's why I've added SchemaSourceFactory interface. 
I do not like adding some abstractions over braid types as they will be mapped one to one internally to braid types, this is additional cost at no benefits at all (we do not plan to move away from braid in foreseeable future), also I believe that we will have to expose internals of braid to allow support for transformations that are not likely to be implemented here as they will be to specific.

src/test/groovy/graphql/nadel/NadelTest.groovy
tsroka: what do you mean? hydration is tested here (tbh it is the only thing that is tested :) )

src/main/java/graphql/nadel/GraphQLRemoteSchemaSourceFactory.java
tsroka: will make it package private, it is only needed for backward compatibility with old Nadel constructor

src/test/groovy/graphql/nadel/NadelTest.groovy
andimarek: I mean the name of the test :-)

PR #4, opened 2018-08-31
Pass argument name of the target link query into braid. 

tsroka:
also: 
- Use latest version of braid.
- do not replace source fields

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,7 +28,7 @@ targetCompatibility = 1.8
 
 def slf4jVersion = '1.7.25'
 def graphqlJavaVersion = '2018-07-30T00-23-12-bfee885'
-def braidVersion = '0.11.3-graphqljava10-SNAPSHOT'
+def braidVersion = '0.12.2-graphqljava10-SNAPSHOT'
 
 def releaseVersion = System.env.RELEASE_VERSION
 version = releaseVersion ? releaseVersion : getDevelopmentVersion()
@@ -50,7 +50,7 @@ jar {
 }
 
 task extractGraphqlGrammar(type: Copy) {
-    from({ zipTree({ configurations.compile.files.find({ it.path.contains('/com.graphql-java/graphql-java/') }) }) }) {
+    from({ zipTree({ configurations.compile.files.find({ it.path.contains('com.graphql-java/graphql-java/') }) }()) }) {
         include '*.g4'
     }
     into 'src/main/antlr'
--- a/src/main/java/graphql/nadel/Nadel.java
+++ b/src/main/java/graphql/nadel/Nadel.java
@@ -16,6 +16,7 @@ import graphql.language.SDLDefinition;
 import graphql.nadel.TransformationUtils.FieldDefinitionWithParentType;
 import graphql.nadel.dsl.FieldTransformation;
 import graphql.nadel.dsl.InnerServiceHydration;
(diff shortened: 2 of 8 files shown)


build.gradle
```diff
@@ -50,7 +50,7 @@ jar {
 }
 
 task extractGraphqlGrammar(type: Copy) {
-    from({ zipTree({ configurations.compile.files.find({ it.path.contains('/com.graphql-java/graphql-java/') }) }) }) {
+    from({ zipTree({ configurations.compile.files.find({ it.path.contains('com.graphql-java/graphql-java/') }) }()) }) {
```
andimarek: why changing that?

src/main/java/graphql/nadel/Nadel.java
```diff
@@ -117,16 +118,18 @@ private Link createHydrationLink(SchemaNamespace schemaNamespace, FieldDefinitio
...
+                        hydration.getSourceLocation()));
 
-        return Link.from(schemaNamespace, definition.parentType(), definition.field().getName(), fromField)
+        return Link.from(schemaNamespace, definition.parentType(), definition.field().getName(),
+                argument.getFieldMappingDefinition().getInputName())
                 //TODO: we need to add something to DSL to support 'queryVariableArgument' parameter of .to
                 // by default it is targetField name which is not always correct.
                 .to(targetService, targetTypeName, targetField.getName())
-                .replaceFromField()
+                .argument(argument.getName())
```
andimarek: why did you remove 'replaceFromField'?

build.gradle
tsroka: When there is an issue (e.g. 'com.graphql-java/graphql-java/' is not found) the error message is very cryptic bc zipTree treats the argument as a closure (it says something like closure cannot be converted to path). Here we are executing the closure so that argument passed to zipTree will be always a path

src/main/java/graphql/nadel/Nadel.java
tsroka: In nadel DSL you define public schema, so it is up to you if you want to include from field or not(you either put it there or not, it is not needed for hydration as long as it is defined in private schema). If replaceFromField will be here there will be no way of including source field AND hydrated field that uses it.

andimarek: 👍

src/test/groovy/graphql/nadel/NadelTest.groovy
```diff
@@ -72,13 +72,13 @@ class NadelTest extends Specification {
...
                 barId: ID
-                bar : String <= \$innerQueries.BarService.bar(id: \$source.barId)
+                bar : String <= \$innerQueries.BarService.bar(barId: \$source.barId)
             }
         }
         
         service BarService {
             type Query {
-                bar(id: ID): Bar
+                bar(barId: ID): Bar
```
andimarek: We don't really have a test, that barId is actually is used as Argument for `bar`: the mocks respond in a very generic way (may fault). Do you think we can make it more strict to actually test that the right argument is used?

build.gradle
andimarek: sounds good

src/test/groovy/graphql/nadel/NadelTest.groovy
tsroka: change test so that i really checks the argument

PR #5, opened 2018-09-04
add the ability to add custom DataFetcher for certain fields

andimarek:
this gives the user of Nadel the ability to inject basically custom code for DataFetchers which is not expressed in the DSL

### DIFF
--- a/src/main/java/graphql/nadel/Nadel.java
+++ b/src/main/java/graphql/nadel/Nadel.java
@@ -80,9 +80,14 @@ public class Nadel {
         }
 
         AsyncExecutionStrategy asyncExecutionStrategy = new AsyncExecutionStrategy();
-        this.braid = Braid.builder()
-                .typeDefinitionRegistry(baseTypeDefinitionRegistry)
-                .withRuntimeWiring(runtimeWiringConsumer)
+        Braid.BraidBuilder braidBuilder = Braid.builder();
+        if (baseTypeDefinitionRegistry != null) {
+            braidBuilder.typeDefinitionRegistry(baseTypeDefinitionRegistry);
+        }
+        if (runtimeWiringConsumer != null) {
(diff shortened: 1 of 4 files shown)


andimarek:
the test is actually green: see https://travis-ci.org/graphql-java/nadel/builds/424219774

src/main/java/graphql/nadel/Nadel.java
```diff
@@ -40,11 +42,22 @@
     private final Map<String, TypeDefinitionRegistry> typesByService = new LinkedHashMap<>();
     private final Map<String, SchemaNamespace> namespaceByService = new LinkedHashMap<>();
 
+    public Nadel(String dsl,
```
tsroka: we need javadoc here. Or maybe make constructor private and use Builder and document each of the fields there?

src/test/groovy/graphql/nadel/NadelTest.groovy
```diff
@@ -107,4 +115,54 @@ class NadelTest extends Specification {
...
+
+        TypeDefinitionRegistry typeDefinitionRegistry = new TypeDefinitionRegistry();
+
+
+        def fieldDefinition = FieldDefinition
+                .newFieldDefinition()
+                .name("specialField")
+                .type(new TypeName("String"))
+                .build()
+        ObjectTypeDefinition objectTypeDefinition = ObjectTypeDefinition.newObjectTypeDefinition()
```
tsroka: I am thinking that we can do better: field definition and object type definition should be built from DSL, not manually only dataFetcher should be supplied. We just need something to tell nadel that this fetcher should be used, how about:
`type Foo {
specialField: String <= $innerFetcher.specialFieldFetcher
}`
and then we could just do with only providing fetchers  -> no need to build types again etc
what do you think?

andimarek:
will not be merged

PR #7, opened 2018-09-06
Custom data fetcher 2

tsroka:
This is continuation of Andi's last PR, however with some support from DSL. It allows to define base service (that does not have backing schema source) and fetchers for it. 

Let me know what you think

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -50,7 +50,7 @@ jar {
 }
 
 task extractGraphqlGrammar(type: Copy) {
-    from({ zipTree({ configurations.compile.files.find({ it.path.contains('/com.graphql-java/graphql-java/') }) }) }) {
+    from({ zipTree({ configurations.compile.files.find({ it.path.contains('com.graphql-java/graphql-java/') }) }()) }) {
         include '*.g4'
     }
     into 'src/main/antlr'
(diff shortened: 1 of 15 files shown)


andimarek:
I am not sure about the general design idea to register special DataFetcher via DSL: anybody who contributes this fields must also contribute the factory method, which means you need to write code, right? 

Not sure this is the way the DSL should go ...

tsroka:
Hey @andimarek, 
I think merging it will be not to bad, I believe that fetchers can become very nice feature of braid. Additionally this PR provides a way of defining directives - in master Nadel you cannot define new directives (or define directives at all) as they are not added to merged schema and error is generated. With this PR new directive can be added in base service definition and used across all services.
What do you think, can we merge it?

andimarek:
I can see how custom datafetcher can be useful of course, but they are a bit magic and the challenge is actually where the implementation comes from.

I don't like the idea of having a base service: it just doesn't map to anything in real life and seems to me like a not optimal solution.  I get the idea of defining directives or scalars outside of services, but I think for example a special `common` area which can have certain restrictions would make more sense.

andimarek:
as discussed we will not do this now, maybe later

PR #9, opened 2018-09-13
Add type definition factory and use it in Nadel constructor

tsroka:


src/main/java/graphql/nadel/Nadel.java
```diff
@@ -42,10 +43,27 @@
...
+     *
+     * @param dsl                 string containing Nadel DSL.
+     * @param schemaSourceFactory schema source factory that provide {@link SchemaSource} for each service defined in
+     *                            DSL.
+     * @param registryFactory     provides additional type definitions that will be added to the stitched schema. If no
+     *                            additional types are needed {@link TypeDefinitionRegistryFactory#DEFAULT} can be used.
+     *
+     * @throws InvalidDslException in case there is an issue with DSL.
+     */
+    public Nadel(String dsl,
```
johnliucn925: Should we have another constructor without TypeDefinitionRegistryFactory?  I guess most time beginner user will not pass it. no need for them to understand what is TypeDefinitionRegistryFactory.default since it is kind of advanced feature.

tsroka: agreed, let's introduce builder in the future PR

PR #10, opened 2018-09-17
Support for providing TypeRuntimeWirings

tsroka:
I forgot in the last PR that TypeDefinitionRegistry itself is not enough to provide data fetchers. This PR fixes that and also adds test.

### DIFF
--- a/src/main/java/graphql/nadel/TypeDefinitionsWithRuntimeWiring.java
+++ b/src/main/java/graphql/nadel/TypeDefinitionsWithRuntimeWiring.java
@@ -1,0 +1,56 @@
+package graphql.nadel;
+
+import graphql.schema.idl.TypeDefinitionRegistry;
+import graphql.schema.idl.TypeRuntimeWiring;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Stores {@link TypeDefinitionRegistry} and collection of {@link TypeRuntimeWiring}s.
+ */
+public class TypeDefinitionsWithRuntimeWiring {
+    private final TypeDefinitionRegistry typeDefinitionRegistry;
+    private final List<TypeRuntimeWiring> typeRuntimeWirings;
+
+    private TypeDefinitionsWithRuntimeWiring(TypeDefinitionRegistry typeDefinitionRegistry,
+                                             List<TypeRuntimeWiring> typeRuntimeWirings) {
+        this.typeDefinitionRegistry = typeDefinitionRegistry;
+        this.typeRuntimeWirings = typeRuntimeWirings;
+    }
+
+    public static Builder newTypeDefinitionWithRuntimeWiring() {
+        return new Builder();
+    }
+
+    public TypeDefinitionRegistry typeDefinitionRegistry() {
+        return typeDefinitionRegistry;
+    }
+
+    public List<TypeRuntimeWiring> typeRuntimeWirings() {
+        return typeRuntimeWirings;
+    }
+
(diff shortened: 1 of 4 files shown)


src/main/java/graphql/nadel/TypeDefinitionsWithRuntimeWiring.java
```diff
@@ -0,0 +1,56 @@
...
+
+        private Builder() {
+        }
+
+        public Builder withTypeDefinitionRegistry(TypeDefinitionRegistry registry) {
+            this.typeDefinitionRegistry = Objects.requireNonNull(registry, "registry");
+            return this;
+        }
+
+        public Builder withTypeRuntimeWiring(TypeRuntimeWiring wiring) {
```
andimarek: just a side note: the naming of the builder is of: graphql-java names it differently a bit from braid and nadel :-(

src/main/java/graphql/nadel/TypeDefinitionsWithRuntimeWiring.java
```diff
@@ -0,0 +1,56 @@
...
+import graphql.schema.idl.TypeRuntimeWiring;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Stores {@link TypeDefinitionRegistry} and collection of {@link TypeRuntimeWiring}s.
+ */
+public class TypeDefinitionsWithRuntimeWiring {
```
andimarek: We should all classes annotate with PublicAPI or Internal (from graphql-java).

src/main/java/graphql/nadel/TypeDefinitionsWithRuntimeWiring.java
```diff
@@ -0,0 +1,56 @@
...
+    private final TypeDefinitionRegistry typeDefinitionRegistry;
+    private final List<TypeRuntimeWiring> typeRuntimeWirings;
```
andimarek: why using this List of TypeRuntimeWiring here? 
Braid expects `Consumer<RuntimeWiring.Builder>` which I think we should also use ....

src/main/java/graphql/nadel/TypeDefinitionsWithRuntimeWiring.java
tsroka: will fix

src/main/java/graphql/nadel/TypeDefinitionsWithRuntimeWiring.java
tsroka: will do

src/main/java/graphql/nadel/TypeDefinitionsWithRuntimeWiring.java
andimarek: I agree that it is nice not to have Consumer but just data,
but the `RuntimeWiring` is not just `TypeRuntimeWiring`: it is also scalars, directivesWiring and more.
By restricting it to `TypeRuntimeWiring` you are restricting the whole wiring from the outside approach: You need `RuntimeWiring` to be able to do everything.

The reason why it is actually a `Consumer<Builder>` and not just `RuntimeWiring` is a braid/ graphql-java shortcoming, because Braid wants to combine different `RuntimeWiring` and there is no easy way to combine them atm, so they give you the Builder.

tsroka: understood. Fixed to use Consumer<builder> as well

PR #23, opened 2019-01-09
 Initial implementation of source query transformation

tsroka:


src/main/java/graphql/nadel/FieldInfo.java
```diff
@@ -5,9 +5,7 @@
 public class FieldInfo {
 
     public enum FieldKind {
```
tsroka: I am not sure if we need field info at all, we can change parser such that FieldDefinition will contain service name(e.g. NadelFieldDefinition extends FieldDefinition). For example, for hydration it is not needed at all cause I can get this information from FieldDefinitionWithTransformation (see SourceQueryTransformer)

src/main/java/graphql/nadel/engine/SourceQueryTransformer.java
```diff
@@ -0,0 +1,124 @@
...
+    private OperationDefinition.Builder operationDefinition = newOperationDefinition();
+
+    private SourceQueryTransformer() {
+    }
+
+    public static Document transform(List<MergedField> fields, ExecutionContext executionContext, ExecutionStepInfo executionStepInfo) {
+        return new SourceQueryTransformer().transformInternal(fields, executionContext, executionStepInfo);
+    }
+
+    private Document transformInternal(List<MergedField> fields, ExecutionContext executionContext, ExecutionStepInfo executionStepInfo) {
```
tsroka: I feel that here we can do much better than just transforming to a target service query. We can also generate reverse transformation function for each field, such that when we apply it to the result from the delegate (Map<String,Object>) we will get a ObjectExecutionResultNode. I feel that it will be the easiest way to do as:
- you will not have to traverse types again and apply reverse transformations in a separate class  (ie. currently in DelegatedResultToResultNode) 
- all the transformation logic will be in one place,
- the only thing that strategy will have to do will be to apply reverse function,  collect  Unresolved nodes (hydration fields) group by service, and execute them again

@andimarek what do you think?

src/main/java/graphql/nadel/engine/NadelExecutionStrategy.java
```diff
@@ -84,7 +81,8 @@ private DelegatedExecution getDelegatedExecutionForFieldDefinition(GraphQLFieldD
                                                                 List<MergedField> mergedFields,
                                                                 DelegatedExecution delegatedExecution,
                                                                 ExecutionStepInfo executionStepInfo) {
-        Document query = mergedFieldsToDocument.mergedSelectionSetToDocument(mergedFields);
+        Document query = SourceQueryTransformer.transform(mergedFields, context, executionStepInfo);
```
andimarek: no static calls please: they are impossible to mock. Just declare an instance variable

src/main/java/graphql/nadel/engine/SourceQueryTransformer.java
```diff
@@ -0,0 +1,124 @@
...
+import static java.util.Collections.singletonList;
+
+public class SourceQueryTransformer {
+    private final ExecutionStepInfoFactory executionStepInfoFactory = new ExecutionStepInfoFactory();
+    private OperationDefinition.Builder operationDefinition = newOperationDefinition();
+
+    private SourceQueryTransformer() {
+    }
+
+    public static Document transform(List<MergedField> fields, ExecutionContext executionContext, ExecutionStepInfo executionStepInfo) {
```
andimarek: same as above: no static methods please.

src/main/java/graphql/nadel/engine/SourceQueryTransformer.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.stream.Collectors;
+
+import static graphql.execution.MergedField.newMergedField;
+import static graphql.language.OperationDefinition.newOperationDefinition;
+import static graphql.language.SelectionSet.newSelectionSet;
+import static java.util.Collections.singletonList;
+
+public class SourceQueryTransformer {
+    private final ExecutionStepInfoFactory executionStepInfoFactory = new ExecutionStepInfoFactory();
+    private OperationDefinition.Builder operationDefinition = newOperationDefinition();
```
andimarek: this is quite confusing tbh and should go away when there is no static call anymore

src/main/java/graphql/nadel/engine/SourceQueryTransformer.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static graphql.execution.MergedField.newMergedField;
+import static graphql.language.OperationDefinition.newOperationDefinition;
+import static graphql.language.SelectionSet.newSelectionSet;
+import static java.util.Collections.singletonList;
+
+public class SourceQueryTransformer {
```
andimarek: this whole class should take advantage of `AstTransformer` or maybe `QueryVisitor`: both traverse the query for you. Especially `QueryVisitor` follows Fragments.

andimarek: We need to also add unit tests at some point. Could be later if we decide so.

src/main/java/graphql/nadel/engine/NadelExecutionStrategy.java
tsroka: I have different view on that. Why would you mock a thing that takes data and just transforms it  - in other words  SourceQueryTransformer.transform is a pure function, without any ext deps or side effects? What would be the value of testing NadelExecutionStrategy that mocks this out? Of course it is a matter of balance (what to mock and what not) but I think in case of pure functions mocking should not be used (in similar fashion you normally will not mock things like StringUtils.isBlank or others). Happy to have a discussion about that, but generally too much mocking is considered harmful by some :)
https://www.thoughtworks.com/insights/blog/mockists-are-dead-long-live-classicists

src/main/java/graphql/nadel/engine/SourceQueryTransformer.java
tsroka: why? I put in there so that referenced fragments can be added while traversing a source query

src/main/java/graphql/nadel/engine/SourceQueryTransformer.java
tsroka: yup, unit testing is very important here, just wanted to get initial feedback

src/main/java/graphql/nadel/engine/NadelExecutionStrategy.java
andimarek: Yes: I definitely wanna test it with mocked out dependencies. the strategy is complex enough on its own.
please have a look here for the general thinking: https://github.com/graphql-java/graphql-java/blob/master/coding-guidelines.md

andimarek: these static calls are a bad code smell and you box yourself in a corner with that: even if you are not sure when you will mock it out you still can do it.
By using an instance variable you also make clear that you have a dependency on a non trivial piece of code. A static call just hides it.

tsroka: Well, what I am trying to say is that you should never mock it cause this code depends on ourceQueryTransformer and will not work without it.  If I am doing refactor of SourceQueryTransformer that changed the behavior of it, and it was also mocked I will never see that something is wrong here. I take your point that it should not be a static method (as per the codestyle) cause it might hide the fact that it is non trivial, but disagree that it should be mocked. As long as it is pure function it should not be mocked as there are not benefits to it.

PR #127, opened 2020-02-03
Fix batch hydration error forwarding

gnawf:
Previously the error would be swallowed in batch hydration results.


andimarek:
great work

PR #136, opened 2020-02-20
Fix hydration inside rename transformation

gnawf:
Found this bug while creating a test. Reproduction case below:

```
service Foo {
    type Query {
        foo: Foo => renamed from fooOriginal
    }
    type Foo {
        id: ID!
        fooBar: Bar => hydrated from Bar.barById(id: $source.fooBarId)
    }
}
service Bar {
    type Query {
        barById(id: ID!): Bar
    }
    type Bar {
        id: ID!
    }
}
```

Foo Schema

```
type Query {
    fooOriginal: Foo
}
type Foo {
    id: ID!
    fooBarId: ID
}
```

Bar Schema

```
type Query {
    barById(id: ID!): Bar
}
type Bar {
    id: ID
}
```

Query

```
{
    foo {
        id
        fooBar {
            id
        }
    }
}
```

Before this patch, the hydration transformation would create a new `Field` object for `fooBarId` and discard any previous field metadata entries including the rename metadata. As such, when `fooOriginal` is processed for the rename transformation, the `fooBarId` field is discarded and thus not hydrated.


src/main/java/graphql/nadel/engine/FieldMetadataUtil.java
```diff
@@ -135,6 +138,20 @@ private static String writeMetadata(List<FieldMetadata> fieldMetadata) {
...
+    private static Field writeMetadata(Field field, List<FieldMetadata> newMetadata) {
+        return field.transform(builder -> writeMetadata(builder, newMetadata));
+    }
+
+    private static Field.Builder writeMetadata(Field.Builder builder, List<FieldMetadata> newMetadata) {
+        String newSerializedValue = writeMetadata(newMetadata);
+        return builder.additionalData(NADEL_FIELD_METADATA, newSerializedValue);
+    }
+
+    private static List<FieldMetadata> readMetadata(Field field) {
```
andimarek: it is a good practice never to return `null` from a method with List/Collection. This simplifies the client code.

src/test/groovy/graphql/nadel/NadelE2ETest.groovy
```diff
@@ -680,4 +680,84 @@ class NadelE2ETest extends Specification {
 
     }
 
+    def "test hydration inside a renamed field"() {
```
andimarek: I think this test should be in NadelExecutionStrategyTest and be more specific: see the other tests there

src/test/groovy/graphql/nadel/engine/NadelExecutionStrategyTest.groovy
```diff
@@ -2579,4 +2579,82 @@ fragment F1 on TestingCharacter {
...
+        def fieldInfos = topLevelFieldInfo(fooField, fooService)
+
+        def nadelExecutionStrategy = new NadelExecutionStrategy([fooService, barService], fieldInfos, overallSchema, instrumentation, serviceExecutionHooks)
+        def executionData = createExecutionData(query, overallSchema)
+
+        when:
+        nadelExecutionStrategy.execute(executionData.executionContext, executionData.fieldSubSelection).join()
+
+        then:
+        1 * service1Execution.execute({ ServiceExecutionParameters sep ->
```
andimarek: please add also assertions regarding the result: we need to make this works as expected

PR #144, opened 2020-03-24
Blow up on redefinition

gnawf:
Currently we accept types that are redefined in Nadel. This becomes an issue when the multiple type definitions differ e.g.

```
service A {
    type GenericResponse {
        message: String
    }
}
service B {
    type GenericResponse {
        message: String
        someExtraField: String
    }
}
```

Depending on the order of parsing `someExtraField` may not be in the overall schema.

This PR raises an exception if there are types that are redefined in the NDSL.

### DIFF
--- a/src/main/java/graphql/nadel/DefinitionRegistry.java
+++ b/src/main/java/graphql/nadel/DefinitionRegistry.java
@@ -14,7 +14,6 @@ import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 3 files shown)


src/main/java/graphql/nadel/DefinitionRegistry.java
```diff
@@ -46,29 +45,39 @@ public SchemaDefinition getSchemaDefinition() {
...
+        return getOpsDefinitions(Operation.QUERY);
     }
 
     public List<ObjectTypeDefinition> getMutationType() {
-        return getOpsDefinitions(Operation.MUTATION.getName(), Operation.MUTATION.getDisplayName());
+        return getOpsDefinitions(Operation.MUTATION);
     }
 
-    private List<ObjectTypeDefinition> getOpsDefinitions(String typeName, String typeDisplay) {
+    private List<ObjectTypeDefinition> getOpsDefinitions(Operation operation) {
```
gnawf, line 59: This change refactors `getOpsDefinitions` into `getOpsDefinitions` and `getOperationTypeName`. I use `getOperationTypeName` in `OverallSchemaGenerator`.

PR #146, opened 2020-04-08
One Child Error when hydrating a field with extensive argument Field

Sajan-99:
The hydration for the assignee field fails when the argument passed is (accountIds: $source.issue.assignee.accountId). If passing, the second serviceExecution.execute should occur and the following response should be given.


andimarek:
I copied the test into https://github.com/graphql-java/nadel/pull/139

PR #159, opened 2020-07-03
fix concurrent modification occurence in ResultComplexityAggregator

faizan-siddiqui:
moved changing incrementServiceNodeCounts to outside for loop in HydrationInputResolver

### DIFF
--- a/src/main/java/graphql/nadel/result/ResultComplexityAggregator.java
+++ b/src/main/java/graphql/nadel/result/ResultComplexityAggregator.java
@@ -5,15 +5,16 @@ import graphql.Internal;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
 
 
 @Internal
 public class ResultComplexityAggregator {
-    private int totalNodeCount = 0;
+    private AtomicInteger totalNodeCount = new AtomicInteger(0);
     private Map<String, Integer> serviceNodeCounts = new ConcurrentHashMap<>();
 
     public int getTotalNodeCount() {
-        return totalNodeCount;
+        return totalNodeCount.get();
     }
 
     public Map<String, Integer> getServiceNodeCounts() {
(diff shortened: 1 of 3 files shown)


src/main/java/graphql/nadel/result/ResultComplexityAggregator.java
```diff
@@ -4,12 +4,13 @@
 
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 
 @Internal
 public class ResultComplexityAggregator {
     private int totalNodeCount = 0;
```
gnawf, line 12: Think this needs to be an `AtomicInteger` seeing as `serviceNodeCounts` has concurrency issues.

faizan-siddiqui: yea thats a good idea, done now

src/test/groovy/graphql/nadel/result/ResultComplexityAggregatorTest.groovy
```diff
@@ -63,7 +65,7 @@ class ResultComplexityAggregatorTest extends Specification {
         resultComplexityAggregator.incrementServiceNodeCount("service2", 5)
 
         then:
-        resultComplexityAggregator.toString() == "ResultComplexityAggregator{totalNodeCount=24, serviceNodeCountsMap={service1=14, service2=10}}"
+        resultComplexityAggregator.toString() == "ResultComplexityAggregator{totalNodeCount=24, serviceNodeCountsMap={service2=10, service1=14}}"
```
andimarek, line 68: why did this change?

andimarek: I think this comes from the fact that we are not using LinkedHashMap anymore and this also means this can change randomly.
I think we should use `LinkedHashMap` again but with a synchronized Wrapper (Collections.synchronizedMap). This should fix it.

faizan-siddiqui: yea I am just investigating this now, in any case I don't think the ordering of the elements matter, so i can change the toString test to not care about ordering

PR #160, opened 2020-08-13
WIP - Synthetic Field hydration

Sajan-99:


src/test/resources/hydration.json
```diff
@@ -61,7 +60,8 @@
                         "sourceType": "FIELD_ARGUMENT"
                       }
                     }
-                  ]
+                  ],
+                  "objectIdentifier": "id"
```
andimarek, line 64: why did we needed to change/add that?

src/main/java/graphql/nadel/engine/HydrationInputResolver.java
```diff
@@ -282,12 +327,17 @@ private ExecutionResultNode convertSingleHydrationResultIntoOverallResult(Execut
                                                                               ResultComplexityAggregator resultComplexityAggregator
     ) {
 
-
         Map<String, FieldTransformation> transformationByResultField = queryTransformationResult.getFieldIdToTransformation();
         Map<String, String> typeRenameMappings = queryTransformationResult.getTypeRenameMappings();
+
+        ExecutionResultNode root = rootResultNode.getChildren().get(0);
+        if (hydrationTransformation.getUnderlyingServiceHydration().getSyntheticField() != null && root.getChildren().size() != 0) {
```
andimarek, line 334: what happens if we have a synth field but root.getChildren.size() != 0?
We should be explicit and add an Assert here.

andimarek (changes requested):
great work!

We should add at least one test which deals with the case that the toplevel field inside the synth field returns null.
e.g.
```{projects{project(id:"1"){name}}}``` for the underlying service to hydrate but the result is:
```{projects:null}``` or ```{projects:{project:null}}```(I guess one version we test already, but not both.

File: src/main/java/graphql/nadel/engine/HydrationInputResolver.java (hunk 2 above)
Comment (Sajan-99) [reply src/main/java/graphql/nadel/engine/HydrationInputResolver.java:334]: Could you clarify what you mean since the if statement handles that case

File: src/test/resources/hydration.json (hunk 1 above)
Comment (Sajan-99) [reply src/test/resources/hydration.json:64]: I moved it down there in the previous grammar, but it yeah you're right its fine to go back

File: src/main/java/graphql/nadel/engine/HydrationInputResolver.java (hunk 2 above)
Comment (andimarek) [reply src/main/java/graphql/nadel/engine/HydrationInputResolver.java:331]: we should add an assertion there that root.getChildren.size() ==1  and change the condition to .size() > 0 (to make it more readable)

```
Assert.assertTrue(root.getChildren.size() == 1
```

File: src/main/java/graphql/nadel/engine/HydrationInputResolver.java [hunk 3]
```diff
@@ -282,12 +324,17 @@ private ExecutionResultNode convertSingleHydrationResultIntoOverallResult(Execut
                                                                               ResultComplexityAggregator resultComplexityAggregator
     ) {
 
-
         Map<String, FieldTransformation> transformationByResultField = queryTransformationResult.getFieldIdToTransformation();
         Map<String, String> typeRenameMappings = queryTransformationResult.getTypeRenameMappings();
+
+        ExecutionResultNode root = rootResultNode.getChildren().get(0);
```
andimarek, line 330: lets add an Assert here too to make sure getChildren.size() == 1

Sajan-99: yep, added

PR #166, opened 2020-11-06
Fix bug with renaming field when theres a hydration call

Sajan-99:
Basically, twofold:
 a) the path mapper is unable to find the correct environment when there is a list of hydration fields as it isn't updated to the current node, and
 b) if a field is renamed but it's child isn't immediately being hydrated this intermediate child's executionPath isn't updated resulting in a:
graphql.AssertException: Internal error: should never happen: could not find matching normalized field for parent node

### DIFF
--- a/src/test/groovy/graphql/nadel/e2e/NadelE2ERenameHydrationTest.groovy
+++ b/src/test/groovy/graphql/nadel/e2e/NadelE2ERenameHydrationTest.groovy
@@ -1,41 +1,16 @@
 package graphql.nadel.e2e
 
-import graphql.AssertException
-import graphql.ErrorType
-import graphql.GraphQLError
-import graphql.GraphqlErrorException
-import graphql.execution.AbortExecutionException
-import graphql.execution.ExecutionId
-import graphql.execution.ExecutionIdProvider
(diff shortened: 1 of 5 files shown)


src/test/groovy/graphql/nadel/e2e/NadelE2ERenameHydrationTest.groovy
```diff
@@ -0,0 +1,130 @@
...
+import graphql.nadel.hooks.ServiceExecutionHooks
+import graphql.nadel.instrumentation.NadelInstrumentation
+import graphql.nadel.instrumentation.parameters.NadelInstrumentRootExecutionResultParameters
+import graphql.nadel.instrumentation.parameters.NadelInstrumentationCreateStateParameters
+import graphql.nadel.result.ResultNodesUtil
+import graphql.nadel.result.RootExecutionResultNode
+import graphql.nadel.schema.SchemaTransformationHook
+import graphql.nadel.testutils.TestUtil
+import graphql.nadel.testutils.harnesses.IssuesCommentsUsersHarness
+import graphql.schema.*
```
andimarek, line 25: can you please update your formatter to never star imports?

andimarek (approved):

PR #168, opened 2020-11-16
Fix hydration involving field named service

gnawf:


src/main/antlr/StitchingDSL.g4
```diff
@@ -65,7 +65,7 @@ intValue: IntValue;
 
 serviceName: NAME;
 
-topLevelField: NAME;
+topLevelField: NAME | SERVICE;
```
gnawf, line 68: I get that it's descriptive having this as `topLevelField` but could it be something like

```
topLevelField: name;
```

Instead?

andimarek: @gnawf have a look how GraphQL Java handles this problem inside the grammar: 

https://github.com/graphql-java/graphql-java/blob/master/src/main/antlr/GraphqlCommon.g4#L36

`name` should be rule and not a token.

bbakerman (approved):

PR #171, opened 2020-11-30
add dummy field to empty selection sets

temaEmelyan:


src/main/java/graphql/nadel/engine/OverallQueryTransformer.java
```diff
@@ -444,7 +450,44 @@ public TraversalControl enter(TraverserContext<Node> context) {
         );
 
         //noinspection unchecked
-        return (T) newNode;
+        return (T) addUnderscoreTypeNameToEmptySelectionSets(nadelContext, overallTypeInformation, rootVars, treeTransformer, newNode);
```
temaEmelyan, line 453: Trying to traverse the tree again after we traversed it the first time. @andimarek as you correctly said, during the first traversal when we are inside of a parent node we don't know whether the children are going to be deleted or not so we can't make a correct decision on whether the typename alias should be added. During the second traversal we do have this information so we can add the dummy field conditionally. 

Not sure whether this approach is acceptable as we are going to traverse the whole tree again. I think we can optimize the second traversal a bit. During the first traversal when leaving a node we can check whether all children were deleted. We can collect all nodes like that into a set and during the second traversal instead of checking every time whether the alias needs to be added we can just look it up from the set?

I also considered another approach. I though about adding the typename alias when leaving the node during the first traversal by overriding the `leave` method [here](https://github.com/graphql-java/nadel/blob/master/src/main/java/graphql/nadel/engine/OverallQueryTransformer.java#L438) but looks like we can't modify the node on leave. Not sure whether we want to change that?

temaEmelyan: > I think we can optimize the second traversal a bit.

that probably not going to do anything though :(

temaEmelyan:
[This test](https://github.com/graphql-java/nadel/blob/master/src/test/groovy/graphql/nadel/RemovedFieldsTest.groovy#L131) is failing (it's the only one that fails). Instead of `result.data == [commentById: null] ` I get `result.data == [commentById: [:]] ` Trying to understand how to fix it. Should be something simple, like swapping an empty selection set with a null for a top level field when converting execution result to the response but haven't found yet where exactly it needs to be done

src/main/java/graphql/nadel/engine/OverallQueryTransformer.java
```diff
@@ -443,10 +449,50 @@ public TraversalControl enter(TraverserContext<Node> context) {
...
+    ) {
+        if (field == null) {
+            return null;
+        }
+        NodeVisitorStub addDummyFieldToEmptySubSelectsVisitor = new NodeVisitorStub() {
+
+            @Override
+            public TraversalControl visitField(Field node, TraverserContext<Node> context) {
+                OverallTypeInfo overallTypeInfo = overallTypeInformation.getOverallTypeInfo(getId(node));
+                if (overallTypeInfo == null) {
```
temaEmelyan, line 473: double check if this if statement can be removed

src/main/java/graphql/nadel/result/ResultNodesUtil.java
```diff
@@ -143,6 +143,10 @@ private static ExecutionResultData toDataImplObject(ExecutionResultNode root) {
         }
         errors.addAll(root.getErrors());
 
+        if (resultMap.isEmpty()) {
```
temaEmelyan, line 146: get rid of that

src/test/groovy/graphql/nadel/RemovedFieldsTest.groovy
```diff
@@ -42,6 +43,36 @@ class RemovedFieldsTest extends StrategyTestHelper {
         result.errors[0].message.contains("removed field")
     }
 
+    def "all hydrated fields in query are removed"() {
```
andimarek, line 46: I am not sure about these tests: I am missing assertions for the queries we send to the underlying service.
I think these tests should be more like  "underlying service returns null for non-nullable field" in NadelExecutionStrategyTest2.

temaEmelyan: Is that better?
`    def "underlying service returns null for non-nullable field"() {
        given:
        def overallSchema = TestUtil.schemaFromNdsl('''
        service Issues {
            type Query {
                issue: Issue
            }
            type Issue {
                id: ID!
            }
        }
        ''')
        def issueSchema = TestUtil.schema("""
            type Query {
                issue: Issue
            }
            type Issue {
                id: ID!
            }
        """)
        def query = "{issue {id}}"

        def expectedQuery1 = "query nadel_2_Issues {issue {empty_selection_set_typename__UUID:__typename}}"
        def response1 = [issue: [empty_selection_set_typename__UUID: "Query"]]

        Map overallResponse
        List<GraphQLError> errors
        when:
        (overallResponse, errors) = test1Service(
                overallSchema,
                "Issues",
                issueSchema,
                query,
                ["issue"],
                expectedQuery1,
                response1,
                createServiceExecutionHooksWithFieldRemoval(["id"])
        )
        then:
        overallResponse == [issue: [id: null]]
        errors.size() == 1
        errors[0].message == "removed field"
    }
`

src/main/java/graphql/nadel/engine/OverallQueryTransformer.java
temaEmelyan: tests are failing if I remove this if statement

src/main/java/graphql/nadel/engine/OverallQueryTransformer.java
```diff
@@ -443,10 +449,50 @@ public TraversalControl enter(TraverserContext<Node> context) {
...
+                    return TraversalControl.CONTINUE;
+                }
+                GraphQLFieldDefinition fieldDefinitionOverall = overallTypeInfo.getFieldDefinition();
+                GraphQLNamedOutputType fieldTypeOverall = (GraphQLNamedOutputType) GraphQLTypeUtil.unwrapAll(fieldDefinitionOverall.getType());
```
andimarek, line 477: I am wondering if we can make this a bit more clear: do we really need to call the general `maybeAddUnderscoreTypeName`. Can make it more explicit and make a `addEmptySelectionUnderscoeSetTypeName`?

src/main/java/graphql/nadel/engine/ArtificialFieldUtils.java
```diff
@@ -14,32 +14,47 @@
...
+ * with empty selection sets.
  */
 @Internal
 public class ArtificialFieldUtils {
 
     private static final String UNDERSCORE_TYPENAME = Introspection.TypeNameMetaFieldDef.getName();
+    private static final String TYPE_NAME_ALIAS_PREFIX_FOR_EMPTY_SELECTION_SETS = "empty_selection_set_";
+    public static final String TYPE_NAME_ALIAS_PREFIX_FOR_INTERFACES_AND_UNIONS = "type_hint_";
 
     public static Field maybeAddUnderscoreTypeName(NadelContext nadelContext, Field field, GraphQLOutputType fieldType) {
```
andimarek, line 28: similar question to the other: do we actually need to modify this method or can we add a second one which handles only the empty selection set part?

andimarek (approved):
See my comments.

src/main/java/graphql/nadel/engine/OverallQueryTransformer.java
temaEmelyan: Let me try to do that. I am not sure that this is the only place where the typename alias is added to empty selection sets.

temaEmelyan: sorry, might have understood wrong. Do you also want me to get rid of the code that deals with empty selection sets inside of `maybeAddUnderscoreTypeName`?

andimarek: yes if possible, because we don't need it in the general case. Or am I missing something?

temaEmelyan: I think we need it for the removed hydrated fields. When I remove code that deals with empty selection sets from `maybeAddUnderscoreTypeName` two tests are failing in `RemovedFieldsTest.groovy`: 

- all fields are removed from hydrated field
- nested hydrated field is removed

src/main/java/graphql/nadel/engine/ArtificialFieldUtils.java
temaEmelyan: replying to this in the other thread

src/main/java/graphql/nadel/engine/OverallQueryTransformer.java
temaEmelyan: for the removed hydrated fields we rely on the fact that `maybeAddUnderscoreTypeName` will add underscore typename to the empty selection sets here:
https://github.com/graphql-java/nadel/blob/master/src/main/java/graphql/nadel/engine/OverallQueryTransformer.java#L126

So we can get rid of the code that deals with empty selection sets inside of `maybeAddUnderscoreTypeName` if we do something like that:

```java
        Field transformedRootField = topLevelField.transform(builder -> builder.selectionSet(topLevelFieldSelectionSet));

        transformedRootField = ArtificialFieldUtils.maybeAddUnderscoreTypeName(nadelContext, transformedRootField, topLevelFieldTypeOverall);
        transformedRootField = ArtificialFieldUtils.maybeAddEmptySelectionSetUnderscoreTypeName(nadelContext, transformedRootField, topLevelFieldTypeOverall);
```

happy to do it if you are ok with it

temaEmelyan: so yeah, I have done it in the latest commit. https://github.com/graphql-java/nadel/pull/171/files#diff-be7fdedc11cd9c188e81f152e70f3e228c48101c61d1478c76a1ec3835d06ef2R133

andimarek (approved):

PR #176, opened 2020-12-08
GQLGW-664: fix hydration when grandparent field has been renamed

gnawf:


src/main/java/graphql/nadel/engine/transformation/FieldUtils.java
```diff
@@ -77,7 +77,7 @@ public static Field pathToFields(List<String> path,
         return curField;
     }
 
-    public static LeafExecutionResultNode geFirstLeafNode(ExecutionResultNode executionResultNode) {
+    public static LeafExecutionResultNode getFirstLeafNode(ExecutionResultNode executionResultNode) {
```
gnawf, line 80: Just fixing typo here

src/main/java/graphql/nadel/engine/PathMapper.java
```diff
@@ -19,21 +18,31 @@ public ExecutionPath mapPath(ExecutionPath executionPath, String resultKey, Unap
             //
             // /issue/reporter might lead to /userById and hence we need to collapse the top level hydrated field INTO the target field
             List<Object> tmp = environment.parentNode.getExecutionPath().toList();
-            tmp.add(fieldSegments.get(fieldSegments.size() - 1));
-            fieldSegments = tmp;
+            tmp.add(pathSegments.get(pathSegments.size() - 1));
+            pathSegments = tmp;
+        } else {
```
faizan-siddiqui, line 23: good catch

faizan-siddiqui (approved):

src/main/java/graphql/nadel/engine/PathMapper.java
andimarek: this look nearly identically to the hydration part now: if the current path only contains one more elements than the environment.parentPath they are identically. Can we try to make them the same?

gnawf: Done @andimarek 

I was originally worried there might be more than one child segment to add so I did the sublist, but this seems to work.

src/test/groovy/graphql/nadel/e2e/NadelE2ERenameHydrationTest.groovy
```diff
@@ -102,4 +107,103 @@ class NadelE2ERenameHydrationTest extends Specification {
         result.join().data == [updateSpecificIssue: [specificIssue: [name: [identity: "Luna"]]]]
     }
 
+    def 'hydration works when an ancestor field has been renamed'() {
```
andimarek, line 110: can you move this test into NadelExecutionStrategyTest2 and make suer you use the helper `test2Services` etc. ... thanks .

andimarek (changes requested):
Looks good, just move the test please, as we prefer to have less E2E tests.

src/test/groovy/graphql/nadel/e2e/NadelE2ERenameHydrationTest.groovy
gnawf: Done, I didn't end up using a `test2Services` etc. because I needed to call the same service twice.

I could create a helper but didn't seem like it was worth the effort?

bbakerman (approved):

PR #184, opened 2021-01-12
skip underlying service execution on forbidden top level field

temaEmelyan:


src/main/java/graphql/nadel/engine/NadelExecutionStrategy.java
```diff
@@ -157,6 +162,13 @@ public NadelExecutionStrategy(List<Service> services,
 
                 ExecutionContext newExecutionContext = buildServiceVariableOverrides(executionContext, queryTransform.getVariableValues());
 
+                String topLevelFieldId = esi.getFieldDefinition().getDefinition().getAdditionalData().get("id");
```
andimarek, line 165: we have static util method to get id I think

src/main/java/graphql/nadel/engine/NadelExecutionStrategy.java
```diff
@@ -211,6 +223,19 @@ public NadelExecutionStrategy(List<Service> services,
         return resultNodes;
     }
 
+    private RootExecutionResultNode getSkippedServiceCallResult(NadelContext nadelContext, ExecutionStepInfo esi, ExecutionContext newExecutionContext, TransformationMetadata.NormalizedFieldAndError topLevelFieldError) {
+        HashMap<String, Object> stringObjectHashMap = new HashMap<>();
```
andimarek, line 227: we use as default always `LinkedHashMap` because it preserves the order of insertion.

andimarek: also: lets find a better name for the variable :-)

src/main/java/graphql/nadel/engine/transformation/TransformationMetadata.java
```diff
@@ -50,6 +51,16 @@ public void add(List<NormalizedQueryField> fields, GraphQLError error) {
         return result;
     }
 
+    public Optional<NormalizedFieldAndError> getRemovedFieldById(String id) {
+        for (NormalizedFieldAndError fieldAndError : removedFields) {
+            String fieldId = fieldAndError.normalizedField.getFieldDefinition().getDefinition().getAdditionalData().get("id");
```
andimarek, line 56: same as above: we have util methods for getting id

src/main/java/graphql/nadel/engine/AsyncIsFieldForbidden.java
```diff
@@ -32,9 +32,32 @@ public AsyncIsFieldForbidden(ServiceExecutionHooks serviceExecutionHooks, NadelC
     }
 
     public CompletableFuture<Map<String, GraphQLError>> getForbiddenFields(Node<?> root) {
-        return visitChildren(root).thenApply(unused -> {
-            return fieldIdsToErrors;
-        });
+        return CompletableFuture.allOf(
```
andimarek, line 35: we should not visit the children if the top level field is remove. The current code does.

andimarek (changes requested):
great work ... see my comments.

src/main/java/graphql/nadel/engine/AsyncIsFieldForbidden.java
```diff
@@ -32,9 +32,32 @@ public AsyncIsFieldForbidden(ServiceExecutionHooks serviceExecutionHooks, NadelC
...
+                checkTopLevelField(root),
+                visitChildren(root)
+        ).thenApply(unused -> fieldIdsToErrors);
+    }
+
+    private CompletableFuture<Void> checkTopLevelField(Node root) {
+        if (!(root instanceof Field)) {
+            return CompletableFuture.completedFuture(null);
+        }
+        Field field = ((Field) root);
```
gnawf, line 45: Nit: remove redundant parentheses

src/main/java/graphql/nadel/engine/AsyncIsFieldForbidden.java
```diff
@@ -32,9 +32,32 @@ public AsyncIsFieldForbidden(ServiceExecutionHooks serviceExecutionHooks, NadelC
...
-        return visitChildren(root).thenApply(unused -> {
-            return fieldIdsToErrors;
-        });
+        return CompletableFuture.allOf(
+                checkTopLevelField(root),
+                visitChildren(root)
+        ).thenApply(unused -> fieldIdsToErrors);
+    }
+
+    private CompletableFuture<Void> checkTopLevelField(Node root) {
```
gnawf, line 41: I think you should compose this code instead of copy pasting the code. This function looks very similar to `visitField`.

e.g.

```
CompleteableFuture<void> visitField(Node<?> node) {
    if (!(node instanceof Field)) {
        return CompletableFuture.completedFuture(null);
    }
    Field field = (Field) node;

    …

    return serviceExecutionHooks.isFieldForbidden(field, normalizedFields, fieldDefinitionOverall, nadelContext.getUserSuppliedContext())
        .whenComplete((graphQLError, throwable) -> {
            if (graphQLError != null && graphQLError.isPresent()) {
                fieldIdsToErrors.put(fieldId, graphQLError.get());
            }
        });
}

CompletableFuture<Void> visitFieldAndChildren(Node<?> node) {
        return visitField(node).thenCompose((error) -> {
            if (error.isPresent()) {
                return CompletableFuture.completedFuture(null);
            }
            return visitChildren(node);
        });
}
```

I think it's possible, but let me know if you think otherwise.

src/main/java/graphql/nadel/engine/NadelExecutionStrategy.java
```diff
@@ -157,6 +162,13 @@ public NadelExecutionStrategy(List<Service> services,
 
                 ExecutionContext newExecutionContext = buildServiceVariableOverrides(executionContext, queryTransform.getVariableValues());
 
+                String topLevelFieldId = esi.getFieldDefinition().getDefinition().getAdditionalData().get("id");
+                Optional<TransformationMetadata.NormalizedFieldAndError> maybeTopLevelFieldError = queryTransform.getRemovedFieldMap().getRemovedFieldById(topLevelFieldId);
+                boolean topLevelFieldExecutionShouldBeSkipped = maybeTopLevelFieldError.isPresent();
+                if (topLevelFieldExecutionShouldBeSkipped) {
```
gnawf, line 168: Move the isPresent check into a util function inside `TransformationMetadata`

That way we can do

```java
var removedFields = queryTransform.getRemovedFieldMap();
if (removedFields.isFieldRemoved(topLevelFieldId)) {
    …
}
```

It's a small change but when your codebase is huge and complex the readability gains are worth it.

src/main/java/graphql/nadel/engine/NadelExecutionStrategy.java
gnawf: Name suggestion: `nulledServiceData`

src/main/java/graphql/nadel/engine/NadelExecutionStrategy.java
```diff
@@ -211,6 +223,19 @@ public NadelExecutionStrategy(List<Service> services,
         return resultNodes;
     }
 
+    private RootExecutionResultNode getSkippedServiceCallResult(NadelContext nadelContext, ExecutionStepInfo esi, ExecutionContext newExecutionContext, TransformationMetadata.NormalizedFieldAndError topLevelFieldError) {
+        HashMap<String, Object> stringObjectHashMap = new HashMap<>();
+        HashMap<String, Object> errorMap = new HashMap<>();
+        errorMap.put("message", topLevelFieldError.getError().getMessage());
+        stringObjectHashMap.put(esi.getFieldDefinition().getName(), null);
```
gnawf, line 230: Nit: move this line right below the definition, I think it reads better when you define the var and immediately use it. That way there's we don't have to juggle the two vars in our heads.

gnawf (changes requested):

src/main/java/graphql/nadel/engine/NadelExecutionStrategy.java
```diff
@@ -211,6 +223,19 @@ public NadelExecutionStrategy(List<Service> services,
         return resultNodes;
     }
 
+    private RootExecutionResultNode getSkippedServiceCallResult(NadelContext nadelContext, ExecutionStepInfo esi, ExecutionContext newExecutionContext, TransformationMetadata.NormalizedFieldAndError topLevelFieldError) {
```
gnawf, line 226: Change the type of `topLevelFieldError` to just `GraphQLError`. It's weird that the value type holds both a field AND error, when we only want the error. It just provides a type safe way to ensure the right value is passed in.

src/main/java/graphql/nadel/engine/NadelExecutionStrategy.java
temaEmelyan: we do have this method to get an id from a `Node` https://github.com/graphql-java/nadel/blob/master/src/main/java/graphql/nadel/dsl/NodeId.java which won't work here as I am getting the id from `FieldDefinition`. If I am to add something like 

```java    
    public static String getId(GraphQLFieldDefinition graphQLFieldDefinition) {
        return assertNotNull(
                graphQLFieldDefinition.getDefinition().getAdditionalData().get(ID),
                () -> String.format("expected field %s to have an id", graphQLFieldDefinition.getName()));
    }
```
some tests start breaking due to the `assertNotNull` as some mocks are not set to have an id on a definition I think. Investigating

src/main/java/graphql/nadel/engine/AsyncIsFieldForbidden.java
temaEmelyan: applied

src/main/java/graphql/nadel/engine/NadelExecutionStrategy.java
temaEmelyan: Not sure about this one. Are you saying the method `getRemovedFieldById(topLevelFieldId)` should stay or be gone? If it stays, does it still return an Optional?

src/main/java/graphql/nadel/dsl/NodeId.java
```diff
@@ -21,6 +22,10 @@ public static String getId(Node<?> node) {
         return assertNotNull(node.getAdditionalData().get(ID), () -> String.format("expected node %s to have an id", node));
     }
 
+    public static String getId(GraphQLFieldDefinition graphQLFieldDefinition) {
```
temaEmelyan, line 25: @andimarek in response to your comments that I should use existing util methods for getting field Ids.

I added this new util method, is that ok? I am not using `assertNotNull` yet as some tests are failing if I start using it. I am checking how to fix the tests. I think the problem is that some tests have mock data without field id in the additional data inside of `FieldDefinition`. I assume that 
`getDefinition().getAdditionalData().get(ID)` should always be not null, is that correct?

andimarek: yes it should always be non null

src/test/groovy/graphql/nadel/hooks/ServiceExecutionHooksTest.groovy
```diff
@@ -171,9 +173,11 @@ class ServiceExecutionHooksTest extends Specification {
...
         """)
 
-        def overallSchema = TestUtil.schema("""
-        type Query {
-            foo(id: String): String
+        def overallSchema = TestUtil.schemaFromNdsl("""
+        service service1 {
+            type Query {
+                foo(id: String): String
+            }
```
temaEmelyan, line 180: these changes to the tests are needed in order for the schema definitions to have field ids in the test data

src/main/java/graphql/nadel/engine/NadelExecutionStrategy.java
gnawf: I'm saying the checking logic here should be simplified.

There should be a method called `isFieldRemoved` that encapsulates the logic here

```java
public boolean isFieldRemoved(String fieldId) {
    return getRemovedFieldById(fieldId).isPresent();
}
```

So you call

```java
var removedFields = queryTransform.getRemovedFieldMap();
if (removedFields.isFieldRemoved(topLevelFieldId)) {
    …
}
```

It just reads a lot better than getting the removed field only to call is present on it

temaEmelyan: yeah but if we do what you are suggesting we will need to call `isFieldRemoved` and then we will need to get the removed field data (the error) anyway. This is why I am asking
> Are you saying the method getRemovedFieldById(topLevelFieldId) should stay or be gone?
I don't think it's providing much value to do what you are suggesting if we need to check `isFieldRemoved` first and then get the field.

gnawf: Ah shit I completely missed the part where you use the variable afterwards. Yea ok my bad. I thought we just got the error just to check if it was present.

gnawf (approved):

PR #189, opened 2021-01-27
change is field forbidden to take a normalized field

temaEmelyan:
This will allow us to make a decision on whether a field is forbidden per normalized field instead of an ast field.


temaEmelyan:
ignore the latest commit, I am adding more tests and noticed that some things got broken. I am working on a fix

andimarek (approved):

PR #193, opened 2021-02-01
Fix top level error

gnawf:
Before it would only put the message into the result instead of all the other information.


src/main/java/graphql/nadel/engine/NadelExecutionStrategy.java
```diff
@@ -227,8 +227,7 @@ public NadelExecutionStrategy(List<Service> services,
     }
 
     private RootExecutionResultNode getSkippedServiceCallResult(NadelContext nadelContext, ExecutionStepInfo esi, ExecutionContext newExecutionContext, GraphQLError error) {
-        HashMap<String, Object> errorMap = new LinkedHashMap<>();
-        errorMap.put("message", error.getMessage());
```
gnawf, line 231: This is the actual issue, rest of diff is just tests.

andimarek (approved):

temaEmelyan (approved):

PR #194, opened 2021-02-02
pass on hydration top level field arguments to isFieldForbidden

temaEmelyan:
not sure that's the way forward, hence the (SPIKE).

We need that to correctly calculate whether a field is allowed or not in the implementation of `isFIeldForbidden`.

### DIFF
--- a/api/src/main/java/graphql/nadel/hooks/HydrationArguments.java
+++ b/api/src/main/java/graphql/nadel/hooks/HydrationArguments.java
@@ -16,7 +16,11 @@ public class HydrationArguments {
     private final List<Argument> hydrationRootAstArguments;
 
     public HydrationArguments(List<GraphQLArgument> hydrationRootGqlArguments, List<Argument> hydrationRootAstArguments) {
-        if (hydrationRootGqlArguments == null || hydrationRootAstArguments == null) {
+        if (hydrationRootAstArguments == null ||
+                hydrationRootGqlArguments == null ||
+                hydrationRootAstArguments.isEmpty() ||
+                hydrationRootGqlArguments.isEmpty()
+        ) {
             this.hydrationRootGqlArguments = Collections.emptyList();
             this.hydrationRootAstArguments = Collections.emptyList();
         } else {
--- a/api/src/main/java/graphql/nadel/hooks/HydrationArguments.java
+++ b/api/src/main/java/graphql/nadel/hooks/HydrationArguments.java
@@ -1,0 +1,63 @@
+package graphql.nadel.hooks;
+
+import graphql.PublicApi;
+import graphql.language.Argument;
+import graphql.nadel.util.FpKit;
+import graphql.schema.GraphQLArgument;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+
+@PublicApi
+public class HydrationArguments {
+
+    private final List<GraphQLArgument> hydrationRootGqlArguments;
+    private final List<Argument> hydrationRootAstArguments;
+
+    public HydrationArguments(List<GraphQLArgument> hydrationRootGqlArguments, List<Argument> hydrationRootAstArguments) {
+        if (hydrationRootGqlArguments == null || hydrationRootAstArguments == null) {
+            this.hydrationRootGqlArguments = Collections.emptyList();
+            this.hydrationRootAstArguments = Collections.emptyList();
+        } else {
+            this.hydrationRootGqlArguments = hydrationRootGqlArguments;
+            this.hydrationRootAstArguments = hydrationRootAstArguments;
+        }
+    }
+
+    public static HydrationArguments empty() {
+        return new HydrationArguments(Collections.emptyList(), Collections.emptyList());
+    }
+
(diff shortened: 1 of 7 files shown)


src/main/java/graphql/nadel/hooks/HydrationArguments.java
```diff
@@ -0,0 +1,61 @@
...
+import graphql.language.Argument;
+import graphql.schema.GraphQLArgument;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+@PublicApi
+public class HydrationArguments {
```
andimarek, line 13: I could imagine we make this more convenient by providing get(name), getValue(name) like methods

temaEmelyan: added these methods

src/main/java/graphql/nadel/hooks/HydrationArguments.java
```diff
@@ -0,0 +1,92 @@
...
+        this.variables = Objects.requireNonNull(builder.variables);
+        this.hydrationRootGqlArguments = builder.hydrationRootGqlArguments == null ? Collections.emptyList() : builder.hydrationRootGqlArguments;
+        this.hydrationRootArguments = builder.hydrationRootArguments == null ? Collections.emptyList() : builder.hydrationRootArguments;
+    }
+
+    public List<GraphQLArgument> getHydrationRootGqlArguments() {
+        return hydrationRootGqlArguments;
+    }
+
+    public List<Argument> getHydrationRootArguments() {
```
andimarek, line 31: I would recommend to name this `AstArguments` to make it more clear.

src/test/groovy/graphql/nadel/RemovedFieldsTest.groovy
```diff
@@ -849,4 +850,373 @@ class RemovedFieldsTest extends StrategyTestHelper {
...
+        type User {
+            id: ID
+            restricted: String
+        }
+        """)
+        def query = 'query issueQ($productId: ID) {issue(id: "ISSUE-1") { id author(productId: $productId) { id restricted }} }'
+        given:
+        def hooks = new ServiceExecutionHooks() {
+            @Override
+            CompletableFuture<Optional<GraphQLError>> isFieldForbidden(NormalizedQueryField normalizedField, HydrationArguments executionParameters, Object userSuppliedContext) {
```
andimarek, line 1087: `HydrationArguments` is named `executionParameters`? that looks weird.

andimarek: also: I would not expect that you access the variables directly: I think you are only interested in argument values. So there should be a `getArgumentValue` method on `HydrationArguments`

andimarek (approved):
small things ... see comments. In general good

engine/src/main/java/graphql/nadel/engine/execution/AsyncIsFieldForbidden.java
```diff
@@ -22,13 +22,20 @@
 
     private final Map<NormalizedQueryField, GraphQLError> fieldsToErrors = new ConcurrentHashMap<>();
     private final ServiceExecutionHooks serviceExecutionHooks;
-    private final ExecutionContext executionContext;
     private final NadelContext nadelContext;
+    private final HydrationArguments hydrationArguments;
+    private final Map<String, Object> variables;
 
-    public AsyncIsFieldForbidden(ServiceExecutionHooks serviceExecutionHooks, ExecutionContext executionContext, NadelContext nadelContext) {
+    public AsyncIsFieldForbidden(
```
andimarek, line 29: lets use the new `nonNullable/nullable` annotations here to indicate what could be null or not...

api/src/main/java/graphql/nadel/hooks/HydrationArguments.java
```diff
@@ -0,0 +1,63 @@
...
+        } else {
+            this.hydrationRootGqlArguments = hydrationRootGqlArguments;
+            this.hydrationRootAstArguments = hydrationRootAstArguments;
+        }
+    }
+
+    public static HydrationArguments empty() {
+        return new HydrationArguments(Collections.emptyList(), Collections.emptyList());
+    }
+
```
andimarek, line 31: these are underlying schema `GraphQLArgument`, right?

temaEmelyan: no, it's overall, see this test https://github.com/atlassian-labs/nadel/pull/194/files#diff-4d0ae59bc268469697cf0a858dd7d447c88b139fd3dbf646ccb2592a82e64ecbR1185

engine/src/main/java/graphql/nadel/engine/execution/AsyncIsFieldForbidden.java
temaEmelyan: added `org.jetbrains.annotations.NotNull` to contructor args, fields, and the public method

andimarek (approved):

PR #195, opened 2021-02-02
Fix bug that redacts all top level fields

gnawf:
So the bug goes

You ask for two fields

```
{
  bar
  foo
}
```

Say the top level field `foo` is rejected, and executing `bar` returns data. The issues comes when we build the skipped/rejected response for `foo`, when that happens the old code actually looks through _all_ the and inserts what it can find, which is null. When it comes to merging the trees, the null field is taken.

This is all because the call to `resultToResultNode.resultToResultNode` uses the overall query, and not the specific top level field query.

### DIFF
--- a/src/main/java/graphql/nadel/engine/NadelExecutionStrategy.java
+++ b/src/main/java/graphql/nadel/engine/NadelExecutionStrategy.java
@@ -13,7 +13,6 @@ import graphql.nadel.FieldInfo;
 import graphql.nadel.FieldInfos;
 import graphql.nadel.Operation;
 import graphql.nadel.Service;
-import graphql.nadel.ServiceExecutionResult;
 import graphql.nadel.dsl.NodeId;
 import graphql.nadel.engine.transformation.FieldTransformation;
 import graphql.nadel.engine.transformation.TransformationMetadata.NormalizedFieldAndError;
@@ -21,7 +20,8 @@ import graphql.nadel.hooks.CreateServiceContextParams;
 import graphql.nadel.hooks.ResultRewriteParams;
 import graphql.nadel.hooks.ServiceExecutionHooks;
 import graphql.nadel.instrumentation.NadelInstrumentation;
-import graphql.nadel.result.ElapsedTime;
+import graphql.nadel.normalized.NormalizedQueryField;
+import graphql.nadel.normalized.NormalizedQueryFromAst;
 import graphql.nadel.result.ExecutionResultNode;
 import graphql.nadel.result.ResultComplexityAggregator;
 import graphql.nadel.result.RootExecutionResultNode;
@@ -31,8 +31,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -44,6 +42,7 @@ import static graphql.Assert.assertNotNull;
 import static graphql.nadel.result.RootExecutionResultNode.newRootExecutionResultNode;
 import static graphql.nadel.util.FpKit.map;
 import static java.lang.String.format;
+import static java.util.Collections.emptyMap;
 import static java.util.Collections.singletonList;
 
 @Internal
@@ -53,7 +52,6 @@ public class NadelExecutionStrategy {
     private final ServiceResultNodesToOverallResult serviceResultNodesToOverallResult = new ServiceResultNodesToOverallResult();
     private final OverallQueryTransformer queryTransformer = new OverallQueryTransformer();
     private final ServiceResultToResultNodes resultToResultNode = new ServiceResultToResultNodes();
-
 
     private final FieldInfos fieldInfos;
     private final GraphQLSchema overallSchema;
@@ -101,8 +99,6 @@ public class NadelExecutionStrategy {
                         log.debug("NadelExecutionStrategy time: {} ms, executionId: {}", elapsedTime, executionContext.getExecutionId());
                     });
         }).whenComplete(this::possiblyLogException);
-
-
     }
 
     private Map<Service, Object> serviceContextsByService(List<OneServiceExecution> oneServiceExecutions) {
@@ -132,7 +128,6 @@ public class NadelExecutionStrategy {
         return Async.each(result);
     }
 
-
     private List<CompletableFuture<RootExecutionResultNode>> executeTopLevelFields(
             ExecutionContext executionContext,
             NadelContext nadelContext,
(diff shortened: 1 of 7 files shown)


src/main/java/graphql/nadel/engine/NadelExecutionStrategy.java
```diff
@@ -219,28 +211,67 @@ public NadelExecutionStrategy(List<Service> services,
...
+     * A top level field error is present if the field should not be executed and an
+     * error should be put in lieu. We check this before calling out to the underlying
+     * service. This error is usually present when the field has been rejected by
+     * {@link ServiceExecutionHooks#isFieldForbidden(NormalizedQueryField, Object)}
+     *
+     * @param esi              the {@link ExecutionStepInfo} for the top level field
+     * @param transformedQuery the query for that specific top level field
+     * @return a {@link GraphQLError} if the field was rejected before, otherwise empty
+     */
+    private Optional<GraphQLError> getTopLevelFieldError(ExecutionStepInfo esi, QueryTransformationResult transformedQuery) {
```
andimarek, line 230: can we name that more specific to make clear that it is a forbidden field error.

src/main/java/graphql/nadel/engine/NadelExecutionStrategy.java
```diff
@@ -219,28 +211,67 @@ public NadelExecutionStrategy(List<Service> services,
...
+     * Util function to get a top level {@link NormalizedQueryField} out of the given query.
+     *
+     * @param query                  the query to look in
+     * @param topLevelFieldResultKey the result key of the top level field to find
+     * @return the top level field if found
+     */
+    private static NormalizedQueryField getTopLevelField(NormalizedQueryFromAst query, String topLevelFieldResultKey) {
+        NormalizedQueryField topLevelField = null;
+
+        for (NormalizedQueryField candidate : query.getTopLevelFields()) {
```
andimarek, line 265: this function got into could go into the NormalizedFieldQueryFromAst itself

src/main/java/graphql/nadel/engine/NadelExecutionStrategy.java
```diff
@@ -153,28 +149,24 @@ public NadelExecutionStrategy(List<Service> services,
...
-                        .getRemovedFieldById(topLevelFieldId)
-                        .map(NormalizedFieldAndError::getError);
-                boolean topLevelFieldExecutionShouldBeSkipped = maybeTopLevelFieldError.isPresent();
-                if (topLevelFieldExecutionShouldBeSkipped) {
+                Optional<GraphQLError> maybeTopLevelFieldError = getTopLevelFieldError(esi, transformedQuery);
+                boolean skipTopLevelFieldExecution = maybeTopLevelFieldError.isPresent();
+                if (skipTopLevelFieldExecution) {
                     GraphQLError topLevelFieldError = maybeTopLevelFieldError.get();
-                    return CompletableFuture.completedFuture(getSkippedServiceCallResult(nadelContext, esi, executionContext, topLevelFieldError));
+                    return CompletableFuture.completedFuture(getRejectedTopLevelFieldResult(nadelContext, esi, topLevelFieldError));
```
andimarek, line 165: the naming is off here:
They are either forbidden or rejected. Choose one and make it consistent :-)

src/main/java/graphql/nadel/engine/NadelExecutionStrategy.java
```diff
@@ -219,28 +211,67 @@ public NadelExecutionStrategy(List<Service> services,
...
+     * overall GraphQL response with an error for that specific top level field.
+     *
+     * @param nadelContext context for the execution
+     * @param esi          the {@link ExecutionStepInfo} for the top level field
+     * @param error        the {@link GraphQLError} to put in the overall response
+     * @return {@link RootExecutionResultNode} with the specified top level field nulled out and with the given GraphQL error
+     */
+    private RootExecutionResultNode getRejectedTopLevelFieldResult(NadelContext nadelContext, ExecutionStepInfo esi, GraphQLError error) {
+        String topLevelFieldResultKey = esi.getResultKey();
+        NormalizedQueryFromAst overallQuery = nadelContext.getNormalizedOverallQuery();
```
andimarek, line 250: use the `overallQuery` variable instead of calling `getNormalizedOverallQuery` twice ;)

src/main/java/graphql/nadel/engine/NadelExecutionStrategy.java
```diff
@@ -219,28 +211,67 @@ public NadelExecutionStrategy(List<Service> services,
...
+     * Creates the {@link RootExecutionResultNode} for a rejected field. Generally in
+     * case the underlying service was probably never called and we want to fill the
+     * overall GraphQL response with an error for that specific top level field.
+     *
+     * @param nadelContext context for the execution
+     * @param esi          the {@link ExecutionStepInfo} for the top level field
+     * @param error        the {@link GraphQLError} to put in the overall response
+     * @return {@link RootExecutionResultNode} with the specified top level field nulled out and with the given GraphQL error
+     */
+    private RootExecutionResultNode getRejectedTopLevelFieldResult(NadelContext nadelContext, ExecutionStepInfo esi, GraphQLError error) {
```
andimarek, line 248: consistent naming here: forbidden or rejected?

src/main/java/graphql/nadel/engine/NadelExecutionStrategy.java
```diff
@@ -153,28 +149,24 @@ public NadelExecutionStrategy(List<Service> services,
...
-
-                String topLevelFieldId = NodeId.getId(esi.getFieldDefinition());
-                Optional<GraphQLError> maybeTopLevelFieldError = queryTransform.getRemovedFieldMap()
-                        .getRemovedFieldById(topLevelFieldId)
-                        .map(NormalizedFieldAndError::getError);
-                boolean topLevelFieldExecutionShouldBeSkipped = maybeTopLevelFieldError.isPresent();
-                if (topLevelFieldExecutionShouldBeSkipped) {
+                Optional<GraphQLError> maybeTopLevelFieldError = getTopLevelFieldError(esi, transformedQuery);
+                boolean skipTopLevelFieldExecution = maybeTopLevelFieldError.isPresent();
+                if (skipTopLevelFieldExecution) {
```
andimarek, line 163: naming: what is skip? rejected or forbidden or skipped?

andimarek (changes requested):
see my comments. Especially the consistent naming we should at least aim for

src/main/java/graphql/nadel/engine/NadelExecutionStrategy.java
gnawf: Good point, thought it wasn't an API I could change

gnawf:
Fixed @andimarek

temaEmelyan (approved):

PR #199, opened 2021-02-17
skip unnecessary traversals during rename and hydration

Sajan-99:
This skips 2 unnecessary traversals when the conditions are met. There are obviously special cases in nadel where these wont meet e.g. nested objects for renames or hydrations but this is good enough for most renames to skip past traversing the entire subtree (which can be pretty large in some cases).

image


src/main/java/graphql/nadel/engine/ServiceResultNodesToOverallResult.java
```diff
@@ -324,6 +324,11 @@ private HandleResult unapplyTransformations(ExecutionId executionId,
                                                 Set<ResultPath> hydrationInputPaths) {
 
         if (isArtificialHydrationNode(node.getFieldIds(), new HashSet<String>(transformationToFieldId.values()))) {
+            if (getFieldIdsWithoutTransformationId(node, transformationMetadata).size() != 0) {
+                HandleResult handleResult = HandleResult.simple(nodesWithTransformationIds(node, null, transformationMetadata));
+                handleResult.traversalControl = TraversalControl.ABORT;
+                return handleResult;
+            }
```
Sajan-99, line 331: fixes a bug when an artificial field is also declared as a not-transformed field

PR #203, opened 2021-03-04
GQLGW-776: fix duplicated errors for N redacted nodes

gnawf:
The way this works it checks that the Node we're inserting the error for is the 0th node, not sure if this is fully correct so wanted some eyes on it.

The alternative is having mutable state inside `TransformationMetadata` about whether the error has been inserted in the tree, which felt icky. But tbh so is this solution because I can't be fully confident it's doing what we want.

### DIFF
--- a/src/test/groovy/graphql/nadel/RemovedFieldsTest.groovy
+++ b/src/test/groovy/graphql/nadel/RemovedFieldsTest.groovy
@@ -552,7 +552,6 @@ class RemovedFieldsTest extends StrategyTestHelper {
 
 
     ServiceExecutionHooks createServiceExecutionHooksWithFieldRemoval(List<String> fieldsToRemove) {
-
         return new ServiceExecutionHooks() {
             @Override
             CompletableFuture<Optional<GraphQLError>> isFieldForbidden(NormalizedQueryField normalizedField, Object userSuppliedContext) {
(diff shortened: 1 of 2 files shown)


src/test/groovy/graphql/nadel/RemovedFieldsTest.groovy
```diff
@@ -662,8 +661,69 @@ class RemovedFieldsTest extends StrategyTestHelper {
...
+            id: ID
+            relatedIssue: Issue                 
+            restricted: String
+        }
+        """)
+        def query = "{issues {id restricted}}"
+
+        def hooks = new ServiceExecutionHooks() {
+            @Override
+            CompletableFuture<Optional<GraphQLError>> isFieldForbidden(NormalizedQueryField normalizedField, Object userSuppliedContext) {
```
temaEmelyan, line 694: just a bit of a nit. You can do just 
`def hooks = createServiceExecutionHooksWithFieldRemoval(["restricted"])`

temaEmelyan (approved):

src/test/groovy/graphql/nadel/RemovedFieldsTest.groovy
gnawf: Ah nice, fixed that up

PR #222, opened 2021-04-12
Split Nadel into submodules

gnawf:


andimarek:
why do we still need the `nadel` module? Also: the `buildAndPublish` step seems to be not finished ... not sure why

PR #225, opened 2021-04-13
Try fix publishing

gnawf:


build.gradle
```diff
@@ -151,10 +151,10 @@ subprojects {
                 password = System.env.ARTIFACTORY_API_KEY
             }
             defaults {
-                publications(project.name)
+                publications("nadel")
```
gnawf, line 154: This needs to be the same as the defined publication earlier in the file i.e.

```
    publishing {
        publications {
            nadel(MavenPublication) {
```

Which is why I think the submodules aren't publishing.

andimarek (approved):

PR #231, opened 2021-04-21
Engine skeleton

gnawf:
Better to review the "Introduce skeleton code for engine" commit to get rid of the rename.

Comment any questions. The skeleton is mainly based off Nadel.kt code.

### DIFF
--- a/engine-nextgen/src/main/kotlin/graphql/nadel/KotlinEngine.kt
+++ b/engine-nextgen/src/main/kotlin/graphql/nadel/KotlinEngine.kt
@@ -5,12 +5,34 @@ import graphql.ExecutionResult
 import graphql.ExecutionResultImpl
 import graphql.execution.instrumentation.InstrumentationState
 import graphql.language.Document
+import graphql.language.NodeUtil
+import graphql.language.OperationDefinition
+import graphql.nadel.ServiceExecutionParameters.newServiceExecutionParameters
+import graphql.nadel.enginekt.normalized.NormalizedQueryToDocument
+import graphql.nadel.enginekt.plan.GraphQLQueryPlanner
+import graphql.nadel.enginekt.plan.GraphQLQueryPlan
+import graphql.nadel.enginekt.schema.GraphQLFieldInfos
+import graphql.nadel.enginekt.transform.query.GraphQLQueryTransformer
+import graphql.nadel.enginekt.util.singleOfType
+import graphql.nadel.util.ErrorUtil
+import graphql.normalized.NormalizedField
+import graphql.normalized.NormalizedQueryTreeFactory
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
+import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.future.asCompletableFuture
+import kotlinx.coroutines.future.asDeferred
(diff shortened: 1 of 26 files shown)


engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/normalized/NormalizedQueryToDocument.kt
```diff
@@ -0,0 +1,94 @@
...
+import graphql.language.TypeName
+import graphql.language.Value
+import graphql.nadel.enginekt.util.AnyList
+import graphql.nadel.enginekt.util.AnyMap
+import graphql.nadel.enginekt.util.AnyMapEntry
+import graphql.nadel.enginekt.util.JsonMapEntry
+import graphql.normalized.NormalizedField
+import graphql.normalized.NormalizedQueryTree
+
+class NormalizedQueryToDocument {
```
andimarek, line 27: this should normally come from graphql java if it is a general util thing.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/plan/GraphQLQueryPlan.kt
```diff
@@ -0,0 +1,15 @@
+package graphql.nadel.enginekt.plan
+
+import graphql.nadel.Service
+import graphql.nadel.enginekt.transform.result.GraphQLResultTransform
+import graphql.normalized.NormalizedField
+
+data class GraphQLQueryPlan(
```
andimarek, line 7: we need to rename this: it is not a query plan but a Result plan imho, if it is only concerned with results.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/schema/GraphQLFieldInfo.kt
```diff
@@ -0,0 +1,11 @@
+package graphql.nadel.enginekt.schema
+
+import graphql.nadel.OperationKind
+import graphql.nadel.Service
+import graphql.schema.GraphQLFieldDefinition
+
+data class GraphQLFieldInfo(
```
andimarek, line 7: I would expect that we have a type/kind on it which tells me if it is a renamed field, hydration etc. Also: if we have field definition we should have both definitions, but then again we should try to make it self containted wiithout schema referenced if we only need s smal subset.

engine-nextgen/src/main/kotlin/graphql/nadel/KotlinEngine.kt
```diff
@@ -5,12 +5,34 @@ import graphql.ExecutionResult
...
+import graphql.normalized.NormalizedQueryTreeFactory
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
+import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.future.asCompletableFuture
+import kotlinx.coroutines.future.asDeferred
 import java.util.concurrent.CompletableFuture
 
 class KotlinEngine(nadel: Nadel) : NadelExecutionEngine {
```
andimarek, line 28: Lets call it NextgenEngine and not Kotlin

andimarek (changes requested):

andimarek (approved):

PR #232, opened 2021-04-21
make beginExecute async

temaEmelyan:
exploring making `NadelInstrumentation.beginExecute` async (return a CompletableFuture).

I need it to be able to make an async operation inside of beginExecute and block the following execution depending on the result of the operation


api/src/main/java/graphql/nadel/instrumentation/ChainedNadelInstrumentation.java
```diff
@@ -85,13 +86,21 @@ public InstrumentationState createState(NadelInstrumentationCreateStateParameter
...
                 .map(instrumentation -> {
                     InstrumentationState state = getStateFor(instrumentation, parameters.getInstrumentationState());
                     return instrumentation.beginExecute(parameters.withNewState(state));
                 })
-                .collect(toList()));
+                .collect(toList());
+        return sequence(contextFutures).thenApply(ChainedInstrumentationContext::new);
+    }
+
+    private static <T> CompletableFuture<List<T>> sequence(List<CompletableFuture<T>> futures) {
```
temaEmelyan, line 99: for the spike just putting this method here to unwrap the list of futures into a list. Probably we can use something from `graphql.execution.Async`?

api/src/main/java/graphql/nadel/instrumentation/NadelInstrumentation.java
```diff
@@ -81,8 +81,8 @@ default InstrumentationState createState(NadelInstrumentationCreateStateParamete
      * @param parameters the parameters to this step
      * @return a non null {@link InstrumentationContext} object that will be called back when the step ends
      */
-    default InstrumentationContext<ExecutionResult> beginExecute(NadelInstrumentationExecuteOperationParameters parameters) {
-        return noOp();
+    default CompletableFuture<InstrumentationContext<ExecutionResult>> beginExecute(NadelInstrumentationExecuteOperationParameters parameters) {
```
temaEmelyan, line 84: the main change

andimarek (approved):

PR #238, opened 2021-04-30
Add context to beginExecute params

gnawf:
The tracer needs the user supplied context, so adding it back in here.


dondonz (approved):

bbakerman (commented):
Lets go

PR #250, opened 2021-05-31
Implement hydration on nextgen engine

gnawf:


engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/NadelExecutionContext.kt
```diff
@@ -0,0 +1,12 @@
+package graphql.nadel.enginekt
+
+import graphql.ExecutionInput
+
+data class NadelExecutionContext(
```
gnawf, line 5: We now pass this around instead of just `userContext`.

The `executionInput` is required to call the execution engine again.

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
```diff
@@ -97,30 +98,71 @@ class NextgenEngine(nadel: Nadel) : NadelExecutionEngine {
...
+        pathToSourceField: List<String>,
+        executionContext: NadelExecutionContext,
+    ): ServiceExecutionResult {
+        val sourceField = fold(initial = topLevelField, count = pathToSourceField.size - 1) {
+            it.children.single()
+        }
+
+        // Creates N plans for the children then merges them together into one big plan
+        val executionPlan = sourceField.children.map {
+            executionPlanner.create(executionContext, services, service, rootField = it)
```
gnawf, line 123: We need to think about the semantics here. We can't exactly pass in the source field and its parents into the execution planner because they're already underlying fields. We have an implicit contract that NFs going into the execution planner are overall fields.

The more and more I work with this code I think we need to revisit how we represent the schema in the engine. I'm talking about internal representation which does not affect how the schema files are passed in etc. I'm talking about transforming the schema once the `NextgenEngine` has it. In my rewrite this was much nicer because everything was expressed in terms of a "combined" schema, there was no issue passing anything around and the semantics are much easier to understand.

cc: @andimarek

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/NadelFieldInstruction.kt
```diff
@@ -10,14 +11,14 @@ sealed class NadelFieldInstruction {
 
 data class NadelHydrationFieldInstruction(
     override val location: FieldCoordinates,
-    val sourceService: String,
+    val sourceService: Service,
```
gnawf, line 14: Pass in service instead of just the name. Makes it much easier to operate on.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/plan/NadelExecutionPlanFactory.kt
```diff
@@ -19,19 +22,29 @@ internal class NadelExecutionPlanFactory(
...
+        services: Map<String, Service>,
         service: Service,
         rootField: NormalizedField,
     ): NadelExecutionPlan {
         val executionSteps = mutableListOf<AnyNadelExecutionPlanStep>()
         val relevantTypeRenames = mutableMapOf<String, NadelTypeRenameInstruction>()
 
         traverseQuery(rootField) { field ->
-            field.objectTypeNames.mapNotNull { executionBlueprint.typeInstructions[it] }.forEach { relevantTypeRenames[it.overallName] = it }
+            field.objectTypeNames
```
gnawf, line 35: Reformat across lines

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/plan/NadelExecutionPlanFactory.kt
```diff
@@ -62,13 +75,15 @@ internal class NadelExecutionPlanFactory(
...
             overallSchema: GraphQLSchema,
+            engine: NextgenEngine,
         ): NadelExecutionPlanFactory {
             return NadelExecutionPlanFactory(
                 executionBlueprint,
                 overallSchema,
                 transforms = listOfTransforms(
                     NadelDeepRenameTransform(),
                     NadelTypeRenameResultTransform(),
+                    NadelHydrationTransform(engine),
```
gnawf, line 86: Pass this in so hydration can occur

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelDeepRenameTransform.kt
```diff
@@ -305,31 +302,55 @@ internal class NadelDeepRenameTransform : NadelTransform<NadelDeepRenameTransfor
...
                     destinationPath = parentNode.path + field.resultKey,
                 ),
+            ) + makeRemoveFieldInstructions(state, parentNode, deepRenameInstruction)
+        }
+    }
+
+    /**
+     * Cleans up the [parentNode] by removing fields that were added by this transform.
+     */
+    private fun makeRemoveFieldInstructions(
```
gnawf, line 332: Extract these into a function so that we can return these for every node, so that we always clean up the artificial fields we add.

In the future these should be auto removed, but for now just did this.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelDeepRenameTransform.kt
```diff
@@ -377,19 +398,21 @@ internal class NadelDeepRenameTransform : NadelTransform<NadelDeepRenameTransfor
...
+    /**
+     * Note: this can be null if the type condition was not met
+     */
+    private fun getMatchingDeepRenameInstruction(
         parentMap: JsonMap,
         state: State,
-        executionPlan: NadelExecutionPlan
-    ): NadelDeepRenameFieldInstruction {
+        executionPlan: NadelExecutionPlan,
+    ): NadelDeepRenameFieldInstruction? {
```
gnawf, line 408: I realized that this _can_ be null. Say not all of the object types have the hydration, then we might not have an instruction, and in that case we shouldn't do anything e.g.

```graphql
type Query {
    pet: Pet
}

interface Pet {
    name: String
}

type Dog implements Pet {
    name: String @renamed(from: ["details", "name"])
}

type Cat implements Pet {
    name: String
}
```

If we receive a `Cat`, we might error out.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelTypeRenameResultTransform.kt
```diff
@@ -14,47 +15,58 @@ import graphql.nadel.enginekt.transform.result.json.JsonNodeExtractor
...
-        )
     }
 
-    override fun transformField(transformer: NadelQueryTransformer.Continuation,
-                                service: Service,
-                                overallSchema: GraphQLSchema,
-                                executionPlan: NadelExecutionPlan,
-                                field: NormalizedField,
-                                state: State): NadelTransformFieldResult {
+    override suspend fun transformField(
```
gnawf, line 40: Reformatted so args are not tied to the indentation of the function signature.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/NadelHydrationArgumentsBuilder.kt
```diff
@@ -0,0 +1,113 @@
...
+        return NormalizedInputValue(
+            GraphQLTypeUtil.simplePrint(argumentDef.type), // Type name
+            valueFactory().value,
+        )
+    }
+}
+
+private typealias AnyNormalizedInputValueValue = NormalizedInputValueValue<*>
+
+internal sealed class NormalizedInputValueValue<T> {
```
gnawf, line 98: Added this because I really didn't like how `NormalizedInputValue.value` was just an `Any` where you had to read the doc to find out what you needed.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/util/CollectionUtil.kt
```diff
@@ -19,15 +19,15 @@ inline fun <reified T> Collection<*>.singleOfType(predicate: (T) -> Boolean = {
     return singleOfTypeOrNull(predicate) as T
 }
 
-inline fun <K, E> Iterable<E>.toMap(crossinline keyExtractor: (E) -> K): Map<K, E> {
+inline fun <K, E> Iterable<E>.strictAssociateBy(crossinline keyExtractor: (E) -> K): Map<K, E> {
```
gnawf, line 22: I realize that Kotlin already has a `associateBy` extension function that does something similar. So I renamed this to better match that.

engine/src/test/groovy/graphql/nadel/e2e/nextgen/NadelE2ETest.groovy
```diff
@@ -458,7 +563,56 @@ class NadelE2ETest extends Specification {
         def response = nadel.execute(nadelExecutionInput)
 
         def executionResult = response.get()
-        assert calledService1
+
+        return [executionResult.getData(), executionResult.getErrors()]
+    }
+
+    Object[] testServices(
```
gnawf, line 570: Function to test N number of services.

engine/src/test/groovy/graphql/nadel/e2e/nextgen/NadelE2ETest.groovy
```diff
@@ -458,7 +563,56 @@ class NadelE2ETest extends Specification {
...
+        def makeServiceExecution = { String serviceName ->
+            println "On calling service '$serviceName'"
+            def calls = serviceCalls[serviceName]
+            def execution = Mock(ServiceExecution)
+            calls.each {
+                def expectedDocument = astSorter.sort(it.key)
+                println "Expecting query"
+                def result = new ServiceExecutionResult(it.value)
+                println printAstCompact(expectedDocument)
+                1 * execution.execute({ ServiceExecutionParameters sep ->
```
gnawf, line 587: I've noticed this `1 *` here doesn't actually enforce that it's been called once.

I think instead we should store these in a `Map<ServiceName, ServiceExecution>` and then run the `1 *` later in the `then:` to fix that. But not a big deal right now.

gnawf (commented):
Leaving some comments for reviewers

andimarek (approved):

PR #251, opened 2021-06-01
[READY] bugfix: ids dropped from response when there are hydrations and renames

temaEmelyan:


engine/src/main/java/graphql/nadel/engine/execution/ServiceResultNodesToOverallResult.java
```diff
@@ -595,7 +595,19 @@ public TraversalControl enter(TraverserContext<ExecutionResultNode> context) {
     }
 
     private List<String> getFieldIdsWithoutTransformationId(ExecutionResultNode node, TransformationMetadata transformationMetadata) {
-        return node.getFieldIds().stream().filter(fieldId -> FieldMetadataUtil.getTransformationIds(fieldId, transformationMetadata.getMetadataByFieldId()).size() == 0).collect(Collectors.toList());
```
temaEmelyan, line 598: when we have a top level field rename, child fields will have rename transformations for them. Since the field has a transformation for it, `getFieldIdsWithoutTransformationId` is empty and the `ExecutionResultNode treeWithout` in `splitTreeByTransformationDefinition` will be null.

engine/src/main/java/graphql/nadel/engine/execution/ServiceResultNodesToOverallResult.java
```diff
@@ -594,12 +590,46 @@ public TraversalControl enter(TraverserContext<ExecutionResultNode> context) {
...
-    private List<String> getFieldIdsWithoutTransformationId(ExecutionResultNode node, TransformationMetadata transformationMetadata) {
-        return node.getFieldIds().stream().filter(fieldId -> FieldMetadataUtil.getTransformationIds(fieldId, transformationMetadata.getMetadataByFieldId()).size() == 0).collect(Collectors.toList());
+    private ExecutionResultNode nodesWithoutTransformationIds(ExecutionResultNode executionResultNode, Set<String> transformationIds, TransformationMetadata transformationMetadata) {
+
+        return resultNodesTransformer.transform(executionResultNode, new TraverserVisitorStub<ExecutionResultNode>() {
+
+            @Override
+            public TraversalControl enter(TraverserContext<ExecutionResultNode> context) {
+                ExecutionResultNode node = context.thisNode();
+                List<String> fieldIdsWithId;
```
felipe-gdr, line 600: Does `getFieldIdsWithoutTransformationIds` return a list of fieldIds with or without ids? I'm just checking that the name of this var is not in conflict with what the method is actually returning. It seems to me that the method will return a list of fields **without** (transformation) ids?

engine/src/main/java/graphql/nadel/engine/execution/ServiceResultNodesToOverallResult.java
```diff
@@ -594,12 +590,46 @@ public TraversalControl enter(TraverserContext<ExecutionResultNode> context) {
         });
     }
 
-    private List<String> getFieldIdsWithoutTransformationId(ExecutionResultNode node, TransformationMetadata transformationMetadata) {
-        return node.getFieldIds().stream().filter(fieldId -> FieldMetadataUtil.getTransformationIds(fieldId, transformationMetadata.getMetadataByFieldId()).size() == 0).collect(Collectors.toList());
+    private ExecutionResultNode nodesWithoutTransformationIds(ExecutionResultNode executionResultNode, Set<String> transformationIds, TransformationMetadata transformationMetadata) {
```
felipe-gdr, line 593: Just to check that I understand what this method is doing: 
It will return a copy of the `executionResultNode` parameter, which I'm assuming is the root node, containing only fields that weren't transformed. And sub-nodes that ended up with no fields will be deleted?

engine/src/main/java/graphql/nadel/engine/execution/ServiceResultNodesToOverallResult.java
```diff
@@ -594,12 +590,46 @@ public TraversalControl enter(TraverserContext<ExecutionResultNode> context) {
...
+            }
+        });
     }
 
-    private List<String> getFieldIdsWithTransformationIds(ExecutionResultNode
-                                                                  node, Set<String> transformationIds, TransformationMetadata transformationMetadata) {
+    private List<String> getFieldIdsWithoutTransformationIds(ExecutionResultNode node, Set<String> notTransformationIds, TransformationMetadata transformationMetadata) {
+        return FpKit.filter(node.getFieldIds(), fieldId -> {
+            Map<String, List<FieldMetadata>> metadataByFieldId = transformationMetadata.getMetadataByFieldId();
+            List<String> transformationIds = FieldMetadataUtil.getTransformationIds(fieldId, metadataByFieldId);
```
felipe-gdr, line 617: Q: In which scenario would `getTransformationIds` return ids that are not present in `allRelevantTransformationIds` which is passed down as an argument to this method? Why just checking that the collection returned from `getTransformationIds` is empty (which was what the method was doing before) not enough to assert that the field doesn't have transformation ids associated with it?

felipe-gdr (dismissed):
I'm giving my approval but please take it with a grain of salt. I'm still getting familiar with the code base so I might have missed something important.
I've added a few comments, but they are mostly questions about the behaviour of the code.

engine/src/main/java/graphql/nadel/engine/execution/ServiceResultNodesToOverallResult.java
andimarek: yeah the naming is wrong here.

engine/src/main/java/graphql/nadel/engine/execution/ServiceResultNodesToOverallResult.java
```diff
@@ -594,12 +590,46 @@ public TraversalControl enter(TraverserContext<ExecutionResultNode> context) {
...
+                }
+                ExecutionResultNode changedNode = changeFieldIsInResultNode(node, fieldIdsWithId);
+                return TreeTransformerUtil.changeNode(context, changedNode);
+            }
+        });
     }
 
-    private List<String> getFieldIdsWithTransformationIds(ExecutionResultNode
-                                                                  node, Set<String> transformationIds, TransformationMetadata transformationMetadata) {
+    private List<String> getFieldIdsWithoutTransformationIds(ExecutionResultNode node, Set<String> notTransformationIds, TransformationMetadata transformationMetadata) {
```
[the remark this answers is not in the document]
andimarek: yes, we should improve the naming a bit.

andimarek (changes requested):
should clean up the naming a bit as mentioned by @felipe-gdr

engine/src/main/java/graphql/nadel/engine/execution/ServiceResultNodesToOverallResult.java
temaEmelyan: > containing only fields that weren't transformed
containing only fields that don't have any root transformations for them
if all fields have a root transformation for them, then yes, the field is deleted from that "copy"

temaEmelyan:
I tried addressing all of the comments

andimarek (approved):

PR #260, opened 2021-06-11
Dynamic service resolution error handling

felipe-gdr:
Introduces better error handling in the dynamic service resolution process.

Consumers should avoid throwing exceptions in the service resolution hook, and instead return GraphQL errors with helpful messages and metadata.

The Nadel engine will bypass normal execution for the root node which caused the error and instead craft a result node containing the error message and other details about the field.


felipe-gdr:
I should have created the branch from master but instead used the old branch as a based. It messed up the diff. Will raise another PR.

PR #266, opened 2021-06-18
hydration using namespaced field

temaEmelyan:
follow up on this comment from @felipe-gdr https://github.com/atlassian-labs/nadel/pull/255#discussion_r647943797 where Felipe asked for tests with hydrations.

I thought initially we will need to do more work for hydrations but as it turned out it's already supported. Actually `@namespaced` does not play any significance in this case because hydration syntax itself routes to the appropriate service: we use service name in the hydration statement
```
      issue: Issue => hydrated from Issues.issue.getIssue(id: $source.issueId)
      search: SearchResult => hydrated from IssueSearch.issue.search(arg: $source.arg)
```
so I am not even sure whether we need this test or not as it's basically the same as hydrating through a synthetic field which is covered with tests pretty well already. Let's see what people think :)

### DIFF
--- a/test/src/test/resources/fixtures/hydration through two services sharing a namespaced field.yml
+++ b/test/src/test/resources/fixtures/hydration through two services sharing a namespaced field.yml
@@ -1,0 +1,183 @@
+name: query with two services sharing a namespaced field
+enabled:
+  nextgen: false
+  current: true
+overallSchema:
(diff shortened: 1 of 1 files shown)


andimarek (approved):

andimarek:
Your call to either close it or merge it.

PR #268, opened 2021-06-24
Check for __typename when removing fields from query on namespaced field

felipe-gdr:


test/src/test/resources/fixtures/__typename-is-passed-on-queries-using-dynamic-resolved-services.yml
```diff
@@ -0,0 +1,95 @@
+name: __typename is passed on queries using dynamic resolved services
```
felipe-gdr, line 1: dynamic service resolution wasn't affected because it checks only top level fields

engine/src/main/java/graphql/nadel/engine/execution/NadelExecutionStrategy.java
```diff
@@ -228,7 +229,7 @@ public NadelExecutionStrategy(List<Service> services,
 
             Optional<MergedField> maybeNewMergedField = MergedFieldUtil.includeSubSelection(mergedField, namespacedObjectType, executionCtx,
                     field -> secondLevelFieldDefinitionsForService.stream()
-                            .anyMatch(graphQLFieldDefinition -> graphQLFieldDefinition.getName().equals(field.getName())));
+                            .anyMatch(graphQLFieldDefinition -> graphQLFieldDefinition.getName().equals(field.getName()) || field.getName().equals(Introspection.TypeNameMetaFieldDef.getName())));
```
felipe-gdr, line 232: This is the fix

andimarek (approved):

temaEmelyan (approved):

PR #271, opened 2021-06-29
Fix up tests

gnawf:


test/src/test/resources/fixtures/extending-types-via-hydration-with-variables-arguments.yml
```diff
@@ -60,8 +60,7 @@ query: |
   }
 variables:
   filter:
-    name:
-      - value
+    name: value
```
gnawf, line 63: If you check the definition the value is a `String` and NOT a `List`. The original test had this problem and the test dumper kept it.

test/src/test/resources/fixtures/one-hydration-call-with-fragments-defined.yml
```diff
@@ -61,14 +61,6 @@ query: |
       name
     }
   }
-
-  # TODO: why does this generate an error?
```
gnawf, line 65: Strategy tests were not end to end, unused fragments are against GraphQL spec.

temaEmelyan (approved):

PR #273, opened 2021-07-01
pass custom transforms to the nextgen engine

temaEmelyan:


api/src/main/java/graphql/nadel/Nadel.java
```diff
@@ -91,7 +93,8 @@ private Nadel(
             ServiceExecutionHooks serviceExecutionHooks,
             WiringFactory overallWiringFactory,
             WiringFactory underlyingWiringFactory,
-            SchemaTransformationHook schemaTransformationHook
+            SchemaTransformationHook schemaTransformationHook,
+            List<Object> nextGenTransforms
```
gnawf, line 97: This works… but honestly we should just feed it directly into the engine e.g.

```java
Nadel.newNadel()
  .engineFactory((nadel) -> {
    return new NextgenEngine(nadel, transforms)
  })
  .build()
```

We don't want to be passing around a `List<Object>`.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelTransformJavaCompat.kt
```diff
@@ -0,0 +1,112 @@
...
+        executionBlueprint: NadelOverallExecutionBlueprint,
+        service: Service,
+        overallField: ExecutableNormalizedField,
+        underlyingParentField: ExecutableNormalizedField?,
+        result: ServiceExecutionResult,
+        state: State,
+    ): CompletableFuture<List<NadelResultInstruction>>
+
+    companion object {
+        fun <State : Any> create(compat: NadelTransformJavaCompat<State>): NadelTransform<State> {
```
gnawf, line 54: We should probably add `@JvmStatic` to this function.

In Kotlin without that annotation there is no concept of real static, and for good reason. Effectively if you don't put that annotation on, in Java you have to call it like so:

```
NadelTransformJavaCompat.Companion.create(…)
```

vs if you have `@JvmStatic` on it

```
NadelTransformJavaCompat.create(…)
```

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/plan/NadelExecutionPlanFactory.kt
```diff
@@ -66,11 +67,15 @@ internal class NadelExecutionPlanFactory(
     companion object {
         fun create(
             executionBlueprint: NadelOverallExecutionBlueprint,
+            javaCompatTransforms: List<Any>,
             engine: NextgenEngine,
         ): NadelExecutionPlanFactory {
+            val customNadelTransforms = (javaCompatTransforms as List<NadelTransformJavaCompat<Any>>)
```
gnawf, line 73: We shouldn't assume that anything coming into the nextgen engine is a Java compat. What if one day we have Kotlin in the AGG codebase? Then if we feed in a pure `NadelTransform` instance here, this would crash. This is exactly why we should never use `Object` and unsafe casts if it's possible to use something more specific.

We need to have _type safety_ here. We should take in `List<NadelTransform>`. If anyone needs to feed in a `JavaCompat` that's their issue. In the engine we shouldn't deal with it.

In the AGG code we should have:

```
Nadel.newNadel()
  .engineFactory((nadel) -> {
    List<NadelTransform> transforms = FpKit.map(
      List.of(…),
      NadelTransformJavaCompat::create,
    );
    return new NextgenEngine(
      nadel,
      transforms,
    );
  })
  .build();
```

gnawf (changes requested):
We need to feed in the transformers directly to ensure type safety.

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
```diff
@@ -32,7 +33,9 @@ import kotlinx.coroutines.future.asCompletableFuture
 import kotlinx.coroutines.future.asDeferred
 import java.util.concurrent.CompletableFuture
 
-class NextgenEngine(nadel: Nadel) : NadelExecutionEngine {
+class NextgenEngine(nadel: Nadel, transforms: List<NadelTransform<Any>> = emptyList()) : NadelExecutionEngine {
+    constructor(nadel: Nadel) : this(nadel, emptyList())
```
temaEmelyan, line 37: additional constructor for Java compatibility. If I understand it correctly, Java does not understand Kotlin's default constructor arguments, so without this constructor java will need to do `new NextgenEngine(nadel, List.of())`

engine-nextgen/build.gradle.kts
```diff
@@ -10,3 +12,9 @@ dependencies {
     api("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.4.2")
     api("org.jetbrains.kotlinx:kotlinx-coroutines-jdk8:1.4.2")
 }
+
+tasks.withType<KotlinCompile> {
```
temaEmelyan, line 16: `JvmStatic` in interfaces does not work without this

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
gnawf: Ah you have to add `@JvmOverloads` and the Kotlin compiler will add the extra constructors with the default arguments.

You'll have to make the constructor keyword explicit

```
class Nadel @JvmOverloads constructor(nadel: Nadel ...
```

gnawf:
Looks good, just add the `@JvmOverloads` and it's good to go

gnawf (approved):

PR #274, opened 2021-07-01
GQLGW-1010 Upgrade Spock and Gradle

dondonz:
This PR
+ Upgrades Gradle from 6 to 7
+ Upgrades Spock from 1 to 2
+ Upgrades Groovy from 2.5 to 3.0
+ Removes references to Bintray


temaEmelyan:
looking at the build logs, the wrapper still appears to be gradle 6 (double check by running ./gradlew --version). I think you need to run 
`./gradlew wrapper --gradle-version=7.1` twice
check this doco https://gradle.org/install/ (paragraph called Upgrade with the Gradle Wrapper)

bbakerman (approved):

build.gradle
```diff
@@ -162,7 +162,7 @@ subprojects {
 }
 
 task myWrapper(type: Wrapper) {
```
temaEmelyan, line 164: I think nobody uses this anymore, I think the whole task should be removed

dondonz: Good idea - it's confusing to have this task here

temaEmelyan:
I think publishing might got broken. 
`./gradlew assemble && ./gradlew check --info && ./gradlew artifactoryPublish -x check --info` this is what we use for publishing to artifactory. If I run it on master on my machine I get 401 from artifactory but on this branch I get
```
FAILURE: Build completed with 4 failures.

1: Task failed with an exception.
-----------
* Where:
Build file '/Users/aemelyanenko/Documents/git-atlassian/agg/nadel-github/build.gradle' line: 96

* What went wrong:
Execution failed for task ':nadel:generatePomFileForNadelPublication'.
> Could not apply withXml() to generated POM
   > Cannot create a Publication named 'DELEGATE_FIRST' because this container does not support creating elements by name alone. Please specify which subtype of Publication to create. Known subtypes are: MavenPublication
...
...
```

same for publishing to local maven, for example `./gradlew publishToMavenLocal`

temaEmelyan (approved):

PR #277, opened 2021-07-07
support simple case for 2 transforms on a field

temaEmelyan:


engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/OauthTransform.kt
```diff
@@ -0,0 +1,87 @@
...
+import graphql.nadel.enginekt.transform.query.NadelQueryTransformer
+import graphql.nadel.enginekt.transform.query.QueryPath
+import graphql.nadel.enginekt.transform.result.NadelResultInstruction
+import graphql.nadel.enginekt.transform.result.json.JsonNodeExtractor
+import graphql.nadel.enginekt.util.queryPath
+import graphql.normalized.ExecutableNormalizedField
+import graphql.validation.ValidationError
+import graphql.validation.ValidationErrorType
+
+class OauthTransform : NadelTransform<GraphQLError> {
```
temaEmelyan, line 18: this class is here only for test purposes, I don't want to keep it, just something to help verify that a custom transformation works

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/query/NadelQueryTransformer.kt
```diff
@@ -137,6 +115,29 @@ class NadelQueryTransformer internal constructor(
...
 
+    private suspend fun applyTransformationSteps(
+        field: ExecutableNormalizedField,
+        transformationSteps: List<NadelExecutionPlan.Step<Any>>,
+        executionContext: NadelExecutionContext,
+        continuation: Continuation,
+        service: Service
+    ): NadelTransformFieldResult {
+        var fieldFromPreviousTransform: ExecutableNormalizedField = field
+        var transformFieldResult: NadelTransformFieldResult? = null
```
gnawf, line 126: We need to merge these results instead of just replacing them.

We use these results to keep track of artificial fields which then get automatically removed at the very end of all the transformations.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/plan/NadelExecutionPlanFactory.kt
```diff
@@ -73,11 +74,12 @@ internal class NadelExecutionPlanFactory(
                 executionBlueprint,
                 transforms = listOfTransforms(
                     *transforms.toTypedArray(),
```
gnawf, line 76: Unrelated to your changes, but I really hate how you can't just spread a collection. So dumb.

gnawf:
Good start 😄

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/OauthTransform.kt
```diff
@@ -0,0 +1,87 @@
...
+    ): GraphQLError? {
+        if (overallField.objectTypeNames
+                .map { executionBlueprint.schema.getType(it) }
+                .any { it == null }
+        ) {
+            return null
+        }
+
+        if (overallField.getOneFieldDefinition(executionBlueprint.schema)
+                .getDirective("scopes") != null
```
gnawf, line 35: Change this to some random directive name for now, or implement it only in tests.

gnawf (changes requested):
Looks good now, just change the transform to be out of the way or test-only and we're good to go.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/OauthTransform.kt
temaEmelyan: moved the transform to the test folder and I am feeding the transform to the engine in the `TestFixture.kt`. Let me know what you think

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/query/NadelQueryTransformer.kt
```diff
@@ -19,7 +19,7 @@ class NadelQueryTransformer internal constructor(
         suspend fun transform(fields: List<ExecutableNormalizedField>): List<ExecutableNormalizedField>
     }
 
-    internal data class TransformContext(
+    data class TransformContext(
```
gnawf, line 22: Any reason you removed `internal`? We want `internal` here because this is NOT public API

test/src/test/kotlin/graphql/nadel/tests/TestFixture.kt
```diff
@@ -36,9 +37,11 @@ data class TestFixture(
     }
 }
 
+val customTestTransforms = listOf(RemoveFieldTestTransform())
+
 data class EngineTypeFactories(
     override val current: NadelExecutionEngineFactory = NadelExecutionEngineFactory(::NadelEngine),
-    override val nextgen: NadelExecutionEngineFactory = NadelExecutionEngineFactory(::NextgenEngine),
+    override val nextgen: NadelExecutionEngineFactory = NadelExecutionEngineFactory { nadel -> NextgenEngine(nadel, customTestTransforms) },
```
gnawf, line 44: Can you add a TODO here

```
// TODO: transforms should be set via the engine test hook.
```

This isn't the ideal solution because we don't want this transform on all tests. But for now it'll work.

gnawf (changes requested):

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/query/NadelQueryTransformer.kt
temaEmelyan: thanks, have no idea why I did that

test/src/test/kotlin/graphql/nadel/tests/TestFixture.kt
temaEmelyan: done!

gnawf (approved):
LGTM

PR #278, opened 2021-07-07
GQLGW-1076: Add serviceContext to Service object

felipe-gdr:
This adds `serviceContext` to `Service`. I think it makes sense to have the service and the context packaged together, as consumers would often need the service + the context when writing their transformers.


engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
```diff
@@ -35,6 +37,7 @@ import java.util.concurrent.CompletableFuture
 
 class NextgenEngine @JvmOverloads constructor(nadel: Nadel, transforms: List<NadelTransform<out Any>> = emptyList()) : NadelExecutionEngine {
     private val services: Map<String, Service> = nadel.services.strictAssociateBy { it.name }
+    private val serviceExecutionHooks: ServiceExecutionHooks = nadel.serviceExecutionHooks
```
felipe-gdr, line 40: uses the existing `ServiceExecutionHooks` instance that is already being passed in `nadel`.

PR #279, opened 2021-07-08
Expose serviceContext to transforms and pass it down to serviceExecutions

felipe-gdr:
This adds serviceContext to Service. I think it makes sense to have the service and the context packaged together, as consumers would often need the service + the context when writing their transformers.


api/src/main/java/graphql/nadel/Service.java
```diff
@@ -45,4 +56,19 @@ public ServiceDefinition getServiceDefinition() {
     public DefinitionRegistry getDefinitionRegistry() {
         return definitionRegistry;
     }
+
+    public Object getServiceContext() {
+        return serviceContext;
+    }
+
+    public Service withContext(Object serviceContext) {
```
gnawf, line 64: I don't think this belongs here. A `Service` is something static for a given schema. I don't think we should change the lifecycle of it.

It also gets confusing because we pass in a `Map<String, Service>` i.e. all of the services to `NadelTransform.isApplicable`. It's weird having some `Service`s have a context and some that don't.

Also, in the future I'm looking to get rid of `Service` from the engine code once we have a combined runtime schema in place.

I would suggest moving this to `ExecutionContext.serviceContext` for now.

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
```diff
@@ -81,7 +85,11 @@ class NextgenEngine @JvmOverloads constructor(nadel: Nadel, transforms: List<Nad
             fieldToService.getServicesForTopLevelFields(query)
                 .map { (field, service) ->
                     async {
-                        executeTopLevelField(field, service, executionInput)
+                        val serviceContext = serviceExecutionHooks.createServiceContext(
+                            CreateServiceContextParams.newParameters().service(service).build()
+                        )
+
+                        executeTopLevelField(field, service.withContext(serviceContext.await()), executionInput)
```
gnawf, line 92: I think you need to create a new `ServiceContext` for the `executeHydration` function too.

I wish `executeHydration` didn't exist, but without a combined schema, we need it 😞.

gnawf (changes requested):

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
felipe-gdr: Hm, I think we won't need to do that if put the context into `NadelExecutionContext` (so many contexts 😄). 
`NadelExecutionContext` is passed around to hydrations too, so they can use the same instance of `ServiceContext`.

Also, having more than 1 `ServiceContext` instance per `Service` kind of defeats the purpose of having a `ServiceContext`.

api/src/main/java/graphql/nadel/Service.java
felipe-gdr: Great suggestion! 
`NadelExecutionContext` is also already being passed down to transformers, which is handy.

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
```diff
@@ -74,14 +78,20 @@ class NextgenEngine @JvmOverloads constructor(nadel: Nadel, transforms: List<Nad
             executionInput.variables,
         )
 
+        val serviceContexts = services.values.map {
```
gnawf, line 81: Also, this is the default way that you create a `Map` in Kotlin, but we don't really like this approach.

Here it's simple and fine, but if you have entries with duplicate keys, then they just get merged together. This ends up often times masking an underlying error.

Instead you can use:

```
mapFrom(
  service.values.map {
    it.name to …
  }
)
```

Which ends up throwing an error if the size of input entries != size of the output map. We could probably name it better, not sure.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/NadelExecutionContext.kt
```diff
@@ -4,9 +4,14 @@ import graphql.ExecutionInput
 
 data class NadelExecutionContext(
     val executionInput: ExecutionInput,
+    var serviceContexts: Map<String, Any>,
```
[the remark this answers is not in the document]
gnawf: We can probably pass in `serviceExecutionHooks` as a `private val` here and then add a `private serviceContexts: MutableMap` for the memoized values

gnawf (changes requested):

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
```diff
@@ -74,14 +78,16 @@ class NextgenEngine @JvmOverloads constructor(nadel: Nadel, transforms: List<Nad
             executionInput.variables,
         )
 
+        val executionContext = NadelExecutionContext(executionInput, serviceExecutionHooks)
```
felipe-gdr, line 81: Moves the instantiation of the `NadelExecutionContext` up, to ensure we have one instance per execution

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/NadelExecutionContext.kt
```diff
@@ -1,12 +1,33 @@
...
+
     val userContext: Any?
         get() {
             return executionInput.context
         }
+
+    /**
+     * Get the service context for a given service
+     */
+    fun getContextForService(service: Service): CompletableFuture<Any?> {
```
felipe-gdr, line 26: Need to use `CompletableFuture` because it's called by Java code.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/NadelExecutionContext.kt
```diff
@@ -1,12 +1,33 @@
...
+        return serviceContexts.getOrPut(service.name) {
+            hooks.createServiceContext(
+                CreateServiceContextParams.newParameters().service(service).build()
+            )
+        }
```
felipe-gdr, line 31: We could block the call with `.get()` here, but returning a future keeps things more async

gnawf (dismissed):
LGTM

gnawf:
Actually, would be nice to have some tests around this. You'd need to use the `EngineTestHook.makeNadel` and supply a custom transform. It'd be very similar to this test, except that you're checking for service context in the params https://github.com/atlassian-labs/nadel/blob/c94ba79aee0ecf3990e2f49b9b1bb7f5b332dec9/test/src/test/kotlin/graphql/nadel/tests/hooks/query-to-one-service-with-execution-input-passed-down.kt

test/src/test/kotlin/graphql/nadel/tests/hooks/service-context-is-being-set.kt
```diff
@@ -0,0 +1,41 @@
...
+import java.util.concurrent.CompletableFuture
+
+@KeepHook
+class `service-context-is-being-set` : EngineTestHook {
+    override fun makeNadel(engineType: NadelEngineType, builder: Nadel.Builder): Nadel.Builder {
+        val serviceExecutionFactory = builder.serviceExecutionFactory
+
+        return builder
+            .serviceExecutionHooks(object : ServiceExecutionHooks {
+                override fun createServiceContext(params: CreateServiceContextParams?): CompletableFuture<Any?>? {
```
gnawf, line 25: Yea this is the annoying thing about Java compat, it just assumes this is nullable because there's no nullability annotations 😢

test/src/test/kotlin/graphql/nadel/tests/hooks/service-context-is-being-set.kt
```diff
@@ -0,0 +1,41 @@
...
+                    return CompletableFuture.completedFuture("Context for ${params?.service?.name}")
+                }
+            })
+            .serviceExecutionFactory(object : ServiceExecutionFactory by serviceExecutionFactory {
+                override fun getServiceExecution(serviceName: String): ServiceExecution {
+                    val default = serviceExecutionFactory.getServiceExecution(serviceName)
+
+                    return ServiceExecution { params ->
+                        expectThat(params.getServiceContext() as String).isEqualTo("Context for MyService")
```
gnawf, line 34: Btw this could be

```
expectThat(params.getServiceContext()).isA<String>().isEqualTo(…)
```

But nbd. No need to change it.

gnawf (approved):
LGTM

gnawf (approved):

test/src/test/kotlin/graphql/nadel/tests/hooks/service-context-is-being-set.kt
felipe-gdr: Cool! I'll keep that in mind for the next times 👍

PR #280, opened 2021-07-09
aemelyanenko: bug fix hydration within hydration with variables

temaEmelyan:
wip for the bugfix


test/src/test/resources/fixtures/replicate-bug.yml
```diff
@@ -0,0 +1,212 @@
+name: replicate bug
+enabled:
+  current: true
+  nextgen: false
+overallSchema:
+  #language=graphQL
```
gnawf, line 6: Idk about you, but this turns out really bad for me. You always get syntax errors saying you redefined a field or something rather

test/src/test/resources/fixtures/hydration-within-hydration-with-variables.yml
```diff
@@ -0,0 +1,166 @@
...
+    board {
+      issue {
+        comments(first: $first) {
+          totalCount
+        }
+      }
+    }
+  }
+variables: {
+  "id": "some-id",
```
andimarek, line 87: remove that variable if not used/

andimarek (changes requested):
see comment

test/src/test/resources/fixtures/hydration-within-hydration-with-variables.yml
temaEmelyan: done

andimarek (approved):

PR #281, opened 2021-07-09
Nextgen Engine doesn't apply transforms to leaf nodes

felipe-gdr:


test/src/test/resources/fixtures/renamed-simple.yml
```diff
@@ -0,0 +1,89 @@
+name: simple renamed field
+enabled:
+  current: true
+  nextgen: true
```
felipe-gdr, line 4: Will fail in Nextgen

PR #286, opened 2021-07-14
Create single node extractor function

felipe-gdr:
This is function is useful in Transforms that need to manipulate a field in the result data to which they already have a complete json node path to.


engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/result/json/JsonNodeExtractor.kt
```diff
@@ -28,6 +30,34 @@ object JsonNodeExtractor {
...
+    fun getNodeAt(data: JsonMap, path: JsonNodePath): JsonNode? {
+        val rootNode = JsonNode(JsonNodePath.root, data)
+        return getNodeAt(rootNode, path)
+    }
+
+    /**
+     * Extract the node at the given json node path.
+     */
+    fun getNodeAt(rootNode: JsonNode, path: JsonNodePath): JsonNode? {
+        return path.segments.fold(rootNode as JsonNode?) { currentNode, segment ->
```
gnawf, line 45: I think we should short circuit this execution once we hit null. It shouldn't impact performance too much, but no point doing extra iterations if we're always going to return null.

We can add an extension function like the following to `CollectionUtil.kt`

```kotlin
fun <T, R> Iterable<T>.foldWhileNotNull(initial: R, operation: (acc: R, T) -> R): R? {
    var accumulator = initial
    for (element in this) {
        accumulator = operation(accumulator, element) ?: return null
    }
    return accumulator
}
```

Then call `foldWhileNotNull` instead of `fold`.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/result/json/JsonNodeExtractor.kt
```diff
@@ -28,6 +30,34 @@ object JsonNodeExtractor {
...
+            when (currentNode?.value) {
+                is Map<*, *> -> currentNode.value[segment.value]?.let {
```
gnawf, line 47: Can we change these to `AnyMap` which is a type alias for `Map<*, *>`. I've been using them elsewhere in the codebase, open to discussion if you think it's a bad pattern though. Personally I just think it looks nicer 🤷‍♂️.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/result/json/JsonNodeExtractor.kt
```diff
@@ -28,6 +30,34 @@ object JsonNodeExtractor {
...
+                    JsonNode(currentNode.resultPath + segment, it)
+                }
+                is List<*> -> when (segment.value) {
```
gnawf, line 50: Ditto with `AnyMap`.

gnawf (changes requested):

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/result/json/JsonNodeExtractor.kt
```diff
@@ -28,6 +30,34 @@ object JsonNodeExtractor {
...
+                    is Int -> currentNode.value.getOrNull(segment.value as Int)?.let {
```
gnawf, line 51: Not sure if we should strictly cast this to `Int`. Perhaps we should just return null under the assumption that we can't find a JSON node instead of erroring out?

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/result/json/JsonNodeExtractor.kt
```diff
@@ -28,6 +30,34 @@ object JsonNodeExtractor {
...
+                        JsonNode(currentNode.resultPath + segment, it )
```
gnawf, line 52: Nit: there's an extra space before the `)`

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/result/json/JsonNodeExtractor.kt
felipe-gdr: I initially didn't see the point in using the aliases, but I'm starting to get used to them. Happy to make the change

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/result/json/JsonNodeExtractor.kt
felipe-gdr: I've changed the pattern matching expression to use `segment` instead of `value`, that way I don't need to cast anything 👍

gnawf (dismissed):
LGTM

gnawf (approved):
LGTM

PR #288, opened 2021-07-21
Delete field instructions from underlying schema

gnawf:
This becomes an issue when a single underlying field is renamed multiple times. This happens with our testing schema, so just deleting this as a fix since we never look at this data anyway.


gnawf:
Am bad at this Github thing

temaEmelyan (approved):

PR #289, opened 2021-07-28
[READY] add support for beginExecute instrumentation step on next-gen engine

temaEmelyan:
adding support for `beginExecute` instrumentation step on the nextgen engine

### DIFF
--- a/api/src/main/java/graphql/nadel/instrumentation/parameters/NadelInstrumentationExecuteOperationParameters.java
+++ b/api/src/main/java/graphql/nadel/instrumentation/parameters/NadelInstrumentationExecuteOperationParameters.java
@@ -6,7 +6,9 @@ import graphql.execution.instrumentation.InstrumentationState;
 import graphql.language.Document;
 import graphql.language.OperationDefinition;
 import graphql.nadel.normalized.NormalizedQueryFromAst;
+import graphql.normalized.ExecutableNormalizedOperation;
 import graphql.schema.GraphQLSchema;
+import org.jetbrains.annotations.Nullable;
 
 import java.util.Map;
 
@@ -17,7 +19,10 @@ import java.util.Map;
 @PublicApi
 public class NadelInstrumentationExecuteOperationParameters {
     private final InstrumentationState instrumentationState;
+    @Nullable
     private final NormalizedQueryFromAst normalizedQueryFromAst;
+    @Nullable
+    private final ExecutableNormalizedOperation normalizedOperation;
     private final Document document;
     private final GraphQLSchema graphQLSchema;
     private final Map<String, Object> variables;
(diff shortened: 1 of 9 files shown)


api/src/main/java/graphql/nadel/instrumentation/parameters/NadelInstrumentationExecuteOperationParameters.java
```diff
@@ -49,13 +80,19 @@ public NadelInstrumentationExecuteOperationParameters(
...
+        return new NadelInstrumentationExecuteOperationParameters(normalizedQueryFromAst, normalizedOperation, document, graphQLSchema, variables, operationDefinition, instrumentationState, context);
     }
 
+    @Nullable
     public NormalizedQueryFromAst getNormalizedQueryFromAst() {
         return normalizedQueryFromAst;
     }
 
+    @Nullable
+    public ExecutableNormalizedOperation getNormalizedOperation() {
```
temaEmelyan, line 92: Consumers will get an `ExecutableNormalizedOperation` when a query is executed on the next-gen engine. Consumers code in the `beginExecute` callback will need to handle both `NormalizedQueryFromAst` and `ExecutableNormalizedOperation`

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
```diff
@@ -97,27 +100,63 @@ class NextgenEngine @JvmOverloads constructor(
         }
 
         val executionContext = NadelExecutionContext(executionInput, serviceExecutionHooks)
+        val instrumentationContext = getInstrumentationContext(query, queryDocument, executionInput, instrumentationState)
```
temaEmelyan, line 103: this is ported from here https://github.com/atlassian-labs/nadel/blob/master/engine/src/main/java/graphql/nadel/engine/execution/Execution.java#L105

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
```diff
@@ -97,27 +100,63 @@ class NextgenEngine @JvmOverloads constructor(
...
-            fieldToService.getServicesForTopLevelFields(query)
-                .map { (field, service) ->
-                    async {
-                        try {
-                            executeTopLevelField(field, service, executionContext)
-                        } catch (e: Throwable) {
-                            when (e) {
-                                is GraphQLError -> newExecutionResult(error = e)
-                                else -> throw e
+        val result: ExecutionResult
```
gnawf, line 105: In Kotlin `if`, `when`, `try` etc are all expressions rather than statements i.e. they have a output value.

So to be more concise you should write

```kotlin
val result: ExecutionResult = try {
    mergeResults(
        …
    )
} catch (e) {
    instrumentation.onComplete(null, e)
    throw e
}
instrumentation.onComplete(result, null)
```

gnawf (changes requested):
Looks good other than two style changes.

I see you marked the PR as WIP, what's left?

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
temaEmelyan: very nice!

temaEmelyan:
> I see you marked the PR as WIP, what's left?

Should be ready now. Just wanted to double-check compatibility on the AGG side, wasn't 100% sure everything works but apparently it does

gnawf (approved):
LGTM

PR #291, opened 2021-08-03
forward error extensions for hydration errors

temaEmelyan:
when hydration query response has errors, we are adding the errors to the overall response but we are not keeping `extensions` or `path` fields of the error. This PR keeps the `extensions` field. ~Will do the same for the `path` field in a separate PR as it a bit more involved (we need to recalculate the path in terms of the overall schema I believe)~
updated the PR to forward path as well


engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/util/GraphQLUtil.kt
```diff
@@ -49,9 +49,15 @@ fun newGraphQLError(
...
     raw: JsonMap,
 ): GraphQLError {
-    return newError()
+    val errorBuilder = newError()
         .message(raw["message"] as String?)
-        .build()
+    raw["extensions"]?.let { extensions ->
+        errorBuilder.extensions(extensions as JsonMap)
+    }
+    raw["path"]?.let { path ->
```
gnawf, line 57: Should we forward `location` too?

On that topic, this is a generic `JsonMap` -> `GraphQLError` so we _could_ change it to something like

```
fun toGraphQLError(
  map: JsonMap,
  retainPath: Boolean = true,
  retainLocation: Boolean = true,
): GraphQLError {
  if (retainPath) raw["path"]?.let { errorBuilder.path(path as AnyList) }
  return errorBuilder.build()
}
```

etc. and let the caller decide what to keep.

**But** we don't have to make this change for now, let's just merge it as is and fix it up later.

gnawf (approved):
Can ignore my comment, just a braindump. LGTM.

PR #292, opened 2021-08-03
Introspection on next gen

temaEmelyan:
Adding an internal `Service` that handles introspection queries. Overall feels a bit hacky but not sure if there is a better way to do it.

### DIFF
--- a/engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
+++ b/engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
@@ -139,15 +139,18 @@ class NextgenEngine @JvmOverloads constructor(
             executionPlan,
         )
         val transformedQuery = queryTransform.result.single()
-        val result = executeService(service, transformedQuery, executionContext)
-        val transformedResult = resultTransformer.transform(
-            executionContext = executionContext,
-            executionPlan = executionPlan,
-            artificialFields = queryTransform.artificialFields,
-            overallToUnderlyingFields = queryTransform.overallToUnderlyingFields,
-            service = service,
-            result = result,
-        )
+        val result: ServiceExecutionResult = executeService(service, transformedQuery, executionContext)
+        val transformedResult: ServiceExecutionResult = when {
+            topLevelField.name.startsWith("__") -> result
+            else -> resultTransformer.transform(
+                executionContext = executionContext,
+                executionPlan = executionPlan,
+                artificialFields = queryTransform.artificialFields,
+                overallToUnderlyingFields = queryTransform.overallToUnderlyingFields,
+                service = service,
+                result = result,
+            )
+        }
 
         @Suppress("UNCHECKED_CAST")
         return newExecutionResult()
--- a/engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
+++ b/engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
@@ -9,7 +9,9 @@ import graphql.execution.instrumentation.InstrumentationState
 import graphql.language.Document
 import graphql.nadel.ServiceExecutionParameters.newServiceExecutionParameters
 import graphql.nadel.enginekt.NadelExecutionContext
+import graphql.nadel.enginekt.blueprint.NadelDefaultIntrospectionRunner
 import graphql.nadel.enginekt.blueprint.NadelExecutionBlueprintFactory
+import graphql.nadel.enginekt.blueprint.NadelIntrospectionRunnerFactory
 import graphql.nadel.enginekt.plan.NadelExecutionPlan
 import graphql.nadel.enginekt.plan.NadelExecutionPlanFactory
 import graphql.nadel.enginekt.transform.NadelTransform
@@ -46,6 +48,7 @@ import java.util.concurrent.CompletableFuture
 class NextgenEngine @JvmOverloads constructor(
     nadel: Nadel,
     transforms: List<NadelTransform<out Any>> = emptyList(),
+    introspectionRunnerFactory: NadelIntrospectionRunnerFactory = NadelIntrospectionRunnerFactory(::NadelDefaultIntrospectionRunner),
 ) : NadelExecutionEngine {
     private val services: Map<String, Service> = nadel.services.strictAssociateBy { it.name }
     private val overallSchema = nadel.overallSchema
@@ -64,7 +67,7 @@ class NextgenEngine @JvmOverloads constructor(
     )
     private val resultTransformer = NadelResultTransformer(overallExecutionBlueprint)
     private val instrumentation = nadel.instrumentation
-    private val fieldToService = NadelFieldToService(overallExecutionBlueprint)
+    private val fieldToService = NadelFieldToService(overallExecutionBlueprint, introspectionRunnerFactory)
     private val executionIdProvider = nadel.executionIdProvider
 
     override fun execute(
--- a/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/IntrospectionService.kt
+++ b/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/IntrospectionService.kt
@@ -6,26 +6,32 @@ import graphql.GraphqlErrorHelper.toSpecification
 import graphql.language.AstPrinter
 import graphql.nadel.Service
 import graphql.nadel.ServiceExecution
+import graphql.nadel.ServiceExecutionParameters
 import graphql.nadel.ServiceExecutionResult
 import graphql.schema.GraphQLSchema
+import java.util.concurrent.CompletableFuture
 
-internal class IntrospectionService constructor(schema: GraphQLSchema) :
-    Service(name, schema, makeServiceExecution(schema), null, null) {
+internal class IntrospectionService constructor(
+    schema: GraphQLSchema,
+    introspectionRunnerFactory: NadelIntrospectionRunnerFactory
+) : Service(name, schema, introspectionRunnerFactory.make(schema), null, null) {
 
     companion object {
         const val name = "__introspection"
-
-        private fun makeServiceExecution(schema: GraphQLSchema): ServiceExecution {
-            val introspectionRunner = GraphQL.newGraphQL(schema)
-                .build()
-
-            return ServiceExecution { params ->
-                val executionInput = ExecutionInput.newExecutionInput()
-                    .query(AstPrinter.printAstCompact(params.query))
-
-                introspectionRunner.executeAsync(executionInput)
-                    .thenApply { ServiceExecutionResult(it.getData(), it.errors.map(::toSpecification)) }
-            }
-        }
     }
+}
+
+fun interface NadelIntrospectionRunnerFactory {
+    fun make(schema: GraphQLSchema): ServiceExecution
+}
+
+open class NadelDefaultIntrospectionRunner(schema: GraphQLSchema) : ServiceExecution {
+    private val graphQL = GraphQL.newGraphQL(schema).build()
+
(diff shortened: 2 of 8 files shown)


engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/query/NadelFieldToService.kt
```diff
@@ -43,7 +43,7 @@ internal class NadelFieldToService(
     }
 
     private fun getService(overallField: ExecutableNormalizedField): Service {
```
gnawf, line 45: I think a better solution might be to just hardcode the `Service` here. That way it doesn't affect the blueprint.

We do something like

```kotlin
fun getService(field: ExecutableNormalizedField): Service {
  // Guard
  if (field.name.startsWith("__")) {
    return introspectionService
  }

  // Normal code here
  return …
}
```

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/IntrospectionService.kt
```diff
@@ -0,0 +1,29 @@
...
+    Service(name, schema, makeServiceExecution(schema), null, null) {
+
+    companion object {
+        const val name = "__introspection"
+
+        private fun makeServiceExecution(schema: GraphQLSchema): ServiceExecution {
+            return ServiceExecution { params ->
+                val executionInput = ExecutionInput.newExecutionInput()
+                    .query(AstPrinter.printAstCompact(params.query))
+                GraphQL.newGraphQL(schema)
```
gnawf, line 22: Should we really be creating a `GraphQL` instance every time introspection is executed?

Not sure how expensive this object is, but I have a feeling it should be cached.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/IntrospectionService.kt
```diff
@@ -0,0 +1,29 @@
...
+import graphql.nadel.ServiceExecutionResult
+import graphql.schema.GraphQLSchema
+
+internal class IntrospectionService constructor(schema: GraphQLSchema) :
+    Service(name, schema, makeServiceExecution(schema), null, null) {
+
+    companion object {
+        const val name = "__introspection"
+
+        private fun makeServiceExecution(schema: GraphQLSchema): ServiceExecution {
```
gnawf, line 18: I think we need to provide a default impl and feed it in as an argument to `Nadel` or `NextgenEngine`.

We need to do this because the AGG needs to wrap around this code to time the introspection, for metrics sake and for internal latency.

e.g.

```kotlin
internal class IntrospctionService(
  schema: GraphQLSchema,
  introspectionRunnerFactory: IntrospectionRunnerFactory,
) : Service(name, schema, introspectionRunnerFactory.make(schema), null, null)

fun interface IntrospectionRunnerFactory {
  fun make(schema: GraphQLSchema): ServiceExecution
}

open class DefaultIntrospectionRunner(schema: GraphQLSchema): ServiceExecution {
  private val graphQL = GraphQL.newSchema(schema).build()

  override fun execute(params): CompletableFuture<ServiceExecutionResult> {
    return graphQL.executeAsync(…).thenApply { … }
  }
}
```

Then the AGG can be like

```kotlin
class GatewayIntrospectionRunner(schema: GraphQLSchema) : DefaultIntrospectionRunner(schema) {
  override fun execute(params): CompletableFuture {
    // start timer
    return super.execute(params).whenComplete((result, throwable) -> /* record internal latency etc. */)
  }
}
```

gnawf (changes requested):
Good work overall, left an option to improve the `Service` code by shoving it _only_ into `NadelFieldToService` instead of convoluting the blueprint factory code.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/IntrospectionService.kt
```diff
@@ -0,0 +1,29 @@
...
+            return ServiceExecution { params ->
+                val executionInput = ExecutionInput.newExecutionInput()
+                    .query(AstPrinter.printAstCompact(params.query))
```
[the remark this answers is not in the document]
andimarek: Future could look like this: we can extend the ServiceExecution to take actually a ExecutableNormalizedField which then could be used in the future to execute the request. I want that future, but this is not immediate as we we need an engine being able to execute a normalized fields instead of Document/String.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/util/CollectionUtil.kt
```diff
@@ -43,7 +43,7 @@ inline fun <K, E> Sequence<E>.strictAssociateBy(crossinline keyExtractor: (E) ->
  * into the [Map] is [List.map]ped from another [Collection].
  */
 @JvmName("mapFromPairs")
-fun <K, V> mapFrom(entries: Collection<Pair<K, V>>): Map<K, V> {
+fun <K, V> mapFrom(entries: Collection<Pair<K, V>>): MutableMap<K, V> {
```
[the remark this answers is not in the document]
temaEmelyan: opted out of this change

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/IntrospectionService.kt
temaEmelyan: interesting. does the current engine provide this functionality or is it a new feature you think about introducing?

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
```diff
@@ -139,15 +139,18 @@ class NextgenEngine @JvmOverloads constructor(
...
-            executionContext = executionContext,
-            executionPlan = executionPlan,
-            artificialFields = queryTransform.artificialFields,
-            overallToUnderlyingFields = queryTransform.overallToUnderlyingFields,
-            service = service,
-            result = result,
-        )
+        val result: ServiceExecutionResult = executeService(service, transformedQuery, executionContext)
+        val transformedResult: ServiceExecutionResult = when {
+            topLevelField.name.startsWith("__") -> result
```
felipe-gdr, line 144: Can you use something from `graphql-java` to check if this is an introspection field? Like `Introspection.isIntrospectionTypes`? Although that takes a `GraphQLNamedType`...

felipe-gdr (dismissed):

test/src/test/resources/fixtures/introspection/if-there-are-a-mix-of-system-fields-and-normal-fields-it-works-on-nextgen.yml
```diff
@@ -0,0 +1,85 @@
+name: if there are a mix of system fields and normal fields it works on nextgen
+enabled:
+  current: false
```
felipe-gdr, line 3: does it not work on current gen?

temaEmelyan: yeah, you cannot run a mixture of system and normal fiedls on the current gen https://github.com/atlassian-labs/nadel/blob/d790fcdc9cdd11b8cc6ff7446023a42c56aee218/api/src/main/java/graphql/nadel/introspection/DefaultIntrospectionRunner.java#L44

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/IntrospectionService.kt
gnawf: Current engine supports this. We use it to record a time to execute introspection metric.

See `GatewayIntrospectionRunner`. We should also be using the new class to exclude introspection times from internal latency. The way we do it in the AGG won't work going forward. We currently simply don't record internal latency if there's zero external calls. This works because currently we don't allow mixing of introspection and normal fields. So we NEED this functionality for nextgen.

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
gnawf: Tbh this kinda sucks. We should really skip the entire process of execution planning etc. for introspection queries as there's really nothing to do at all.

But it's fine for now, let's revisit this in the future.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/NadelExecutionBlueprintFactory.kt
```diff
@@ -342,12 +343,12 @@ private class Factory(
                     .filterIsInstance<AnyNamedNode>()
                     .filterNot {
                         it is ObjectTypeExtensionDefinition
-                            || it is InterfaceTypeExtensionDefinition
-                            || it is EnumTypeExtensionDefinition
-                            || it is ScalarTypeExtensionDefinition
-                            || it is InputObjectTypeExtensionDefinition
-                            || it is SchemaExtensionDefinition
-                            || it is UnionTypeExtensionDefinition
+                                || it is InterfaceTypeExtensionDefinition
```
gnawf, line 346: Have you run `./gradlew ktlintSetCodeStyle` in the project? Continued indentation should be 4 spaces not 8.

After running that, you may need to fix up the Java codestyle to not use `*` imports.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/query/NadelFieldToService.kt
```diff
@@ -43,8 +46,10 @@ internal class NadelFieldToService(
...
-        if (overallField.name == Introspection.TypeNameMetaFieldDef.name) {
+        if (overallField.name == Introspection.TypeNameMetaFieldDef.name && overallField.parent != null) {
             return getService(overallField.parent)
+        } else if (overallField.name.startsWith("__")) {
```
gnawf, line 51: What happens if someone asks for `__typename` as a top level field?

temaEmelyan: take a look at this test, let me know if it answers you question https://github.com/atlassian-labs/nadel/pull/292/files#diff-008d37ddb0e39a4a268702b2e99a0831bd625bd127cd01123b198d5a2043ea29R81

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/NadelExecutionBlueprintFactory.kt
temaEmelyan: is it `./gradlew ktlintFormat`?
it leaves this file as it is

temaEmelyan: if you want to enforce a certain style, we should make the build break if the code does not adhere to that style

temaEmelyan: `ktlintFormat` looks for the source files in the wrong place apparently. After fixing the path, it does apply style changes but it also removes some used imports which prevents code from compiling so I won't touch it for now

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/query/NadelFieldToService.kt
gnawf: Yup that looks good.

gnawf (approved):
Nice

PR #294, opened 2021-08-09
Prevent errors when merging null result nodes

felipe-gdr:
When merging result nodes, we want to merge `null` values with objects.

Like, if `serviceA` returns an error when serving a query for `field1`, but `serviceB` returns data for it (`field1: { data: "text"}`, we want to merged result to content the data from `serviceB` + the errors from `serviceA`

### DIFF
--- a/engine/src/main/java/graphql/nadel/engine/execution/StrategyUtil.java
+++ b/engine/src/main/java/graphql/nadel/engine/execution/StrategyUtil.java
@@ -159,18 +159,28 @@ public class StrategyUtil {
     private static ExecutionResultNode combineResultNodes(List<ExecutionResultNode> resultNodes) {
         List<ExecutionResultNode> children = new ArrayList<>();
         Map<String, Object> completedValueMap = new LinkedHashMap<>();
+        // TODO: feature flag check, perhaps?
+        List<ExecutionResultNode> nonNullNodes = resultNodes
+                .stream()
+                .filter(resultNode -> resultNode.getCompletedValue() != null)
+                .collect(Collectors.toList());
 
-        for (ExecutionResultNode resultNode : resultNodes) {
+        for (ExecutionResultNode resultNode : nonNullNodes) {
             Object completedValue = resultNode.getCompletedValue();
+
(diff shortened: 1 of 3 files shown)


engine/src/main/java/graphql/nadel/engine/execution/StrategyUtil.java
```diff
@@ -159,18 +159,28 @@ public static RootExecutionResultNode mergeTrees(List<RootExecutionResultNode> r
     private static ExecutionResultNode combineResultNodes(List<ExecutionResultNode> resultNodes) {
         List<ExecutionResultNode> children = new ArrayList<>();
         Map<String, Object> completedValueMap = new LinkedHashMap<>();
+        // TODO: feature flag check, perhaps?
+        List<ExecutionResultNode> nonNullNodes = resultNodes
```
temaEmelyan, line 163: not 100 percent on this but I think we are trying to not use `stream` in nadel and use lightweight functions from `nadel/util/FpKit` where possible. I think Andi and Brad did some benchmark and found that there is non negligible impact when using java's `Streams`. This `stream.filter.collect` can be substituted with `nadel/util/FpKit.filter`

temaEmelyan (dismissed):

engine/src/main/java/graphql/nadel/engine/execution/StrategyUtil.java
```diff
@@ -159,18 +159,28 @@ public static RootExecutionResultNode mergeTrees(List<RootExecutionResultNode> r
     private static ExecutionResultNode combineResultNodes(List<ExecutionResultNode> resultNodes) {
         List<ExecutionResultNode> children = new ArrayList<>();
         Map<String, Object> completedValueMap = new LinkedHashMap<>();
+        // TODO: feature flag check, perhaps?
```
temaEmelyan, line 162: I would get rid of this comment. If you do want to feature flag this, you can use `NadelExecutionHints` class. You would need to create a field there and a consumer (gateway) would need to supply the value of the hint with the `NadelExecutionInput` for every query. 
I wouldn't bother though, but up to you of course :)

temaEmelyan (approved):

PR #295, opened 2021-08-10
use NadelInstrumentation in nextgen instead of NadelEngineInstrumentation

temaEmelyan:
This is ready for review


api/src/main/java/graphql/nadel/instrumentation/ChainedNadelEngineInstrumentation.java
```diff
@@ -14,14 +11,13 @@ public ChainedNadelEngineInstrumentation(List<NadelInstrumentation> instrumentat
     }
 
     @Override
-    public RootExecutionResultNode instrumentRootExecutionResult(RootExecutionResultNode rootExecutionResultNode, NadelInstrumentRootExecutionResultParameters parameters) {
+    public void instrumentRootExecutionResult(Object rootExecutionResult, NadelInstrumentRootExecutionResultParameters parameters) {
```
temaEmelyan, line 14: `Object rootExecutionResult` is actually `Either<RootExecutionResultNode, ExecutionResult>` (left - current gen, right - nextgen)

this instrumentation method is used by gateway's tracer library.
to replicate the functionality that the tracer library has around `RootExecutionResultNode` argument, just supplying `ExecutionResult` is not going to be enough I believe. For example, tracer uses `RootExecutionResultNode`'s `ElapsedTime elapsedTime` field which does not exist in `ExecutionResult`. 
So we either need to extend `ExecutionResult` and add more stuff to it or wrap it around with timings and other stuff and pass that wrapped class to `instrumentRootExecutionResult`

api/src/main/java/graphql/nadel/instrumentation/parameters/NadelInstrumentRootExecutionResultParameters.java
```diff
@@ -13,12 +14,19 @@
 
     private final ExecutionContext executionContext;
     private final InstrumentationState instrumentationState;
-    private final NormalizedQueryFromAst normalizedQueryFromAst;
+    @Nullable
+    private final ExecutableNormalizedOperation query;
```
temaEmelyan, line 18: adding `ExecutableNormalizedOperation` so the tracer lib can resolve the types from the execution result (which is a Json map)

temaEmelyan: still need to add `__typename` to interfaces and unions

api/src/main/java/graphql/nadel/instrumentation/ChainedNadelEngineInstrumentation.java
gnawf: I don't think we need this, I think `beginExecute` already provides this value in the `onComplete`? From the `NadelInstrumentationExecuteOperationParameters` we even have the `ExecutableNormalizedOperation`.

Also just ranting below, not really necessary to read.

I'm always weary of having `Object` as the type. It's not a big deal to add another function with the new type, but when you change something to `Object` it's always a PITA because you or other devs never know what the value _could_ be unless you read documentation, see working with `ResultPath#getSegmentValue` or `NormalizedInputValue#getValue`. Little to no documentation about what the `Object` actually can be, makes it very error prone. There's next to 0 type safety around this.

In the ideal world we'd be using union types. The closest thing we have to that in Kotlin is sealed types to wrap the original objects e.g.

```kotlin
sealed class JsonNodePathSegment<T : Any> {
    abstract val value: T

    data class Int(override val value: kotlin.Int) : JsonNodePathSegment<kotlin.Int>()
    data class String(override val value: kotlin.String) : JsonNodePathSegment<kotlin.String>()
}
```

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
```diff
@@ -98,7 +103,14 @@ class NextgenEngine @JvmOverloads constructor(
         }
 
         val executionContext = NadelExecutionContext(executionInput, serviceExecutionHooks)
-        val instrumentationContext = getInstrumentationContext(query, queryDocument, executionInput, overallSchema, instrumentation, instrumentationState)
+        val instrumentationContext = getInstrumentationContext(
```
gnawf, line 106: A little unrelated to this PR, but this function should really be refactored as it's not clear at all what it's doing i.e. it's calling `beginExecute` but would you really know that from the name `getInstrumentationContext` or from the args?

We should change `getInstrumentationContext` to an extension function instead

```kotlin
internal suspend fun NadelInstrumentation.beginExecute(
    query: ExecutableNormalizedOperation,
    queryDocument: Document,
    executionInput: ExecutionInput,
    graphQLSchema: GraphQLSchema,
    instrumentationState: InstrumentationState?,
): InstrumentationContext<ExecutionResult> {
    val nadelInstrumentationExecuteOperationParameters = buildInstrumentationExecutionParameters(
        query,
        queryDocument,
        executionInput,
        graphQLSchema,
        instrumentationState,
    )

    return beginExecute(nadelInstrumentationExecuteOperationParameters)
        .asDeferred()
        .await()
}
```

That way we're just calling `instrumentation.beginExecute(query, queryDocument, executionInput, graphQLSchema, instrumentationState)`

gnawf (changes requested):
I think we already have all the info we need in `beginExecute` to run this.

The ONLY reason we used `instrumentRootExecutionResult` and didn't use `beginExecute` was because `RootExecutionResult` had type info in the old engine. But we don't have that at all now, so we should be solely be relying on `beginExecute`

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
temaEmelyan: very nice!

api/src/main/java/graphql/nadel/instrumentation/parameters/NadelInstrumentRootExecutionResultParameters.java
temaEmelyan: > still need to add __typename to interfaces and unions

ok, that's done

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
```diff
@@ -134,6 +145,7 @@ class NextgenEngine @JvmOverloads constructor(
         service: Service,
         executionContext: NadelExecutionContext,
     ): ExecutionResult {
+        addTypenameToInterfacesAndUnions(topLevelField, typenameHintAlias, overallSchema)
```
gnawf, line 148: Is there a reason we're not doing this via a `NadelTransform`?

I see we're using `recursivelyRemoveKey` which just traverses the entire JSON tree, which we actually don't want because we end up traversing the entire tree.

gnawf (commented):
I think for performance reasons we can't really approach it this way. We want to avoid traversing the entire tree.

Also Apollo tracing is generally async, if we remove the typenames right after invoking the instrumentation, then we lose all that information anyway.

PR #296, opened 2021-08-10
Improve test coverage

gnawf:
Adds code to analyse combinations of Nadel features via inspecting the blueprint.

Current coverage:

```
Found [type rename result]

Found [hydration]
Found [hydration] under [hydration]
Found [hydration] under [rename]
Found [hydration] under [hydration, hydration]

Found [batch hydration]
Found [batch hydration] under [rename]
Found [batch hydration] under [batch hydration]
Found [batch hydration] under [batch hydration, batch hydration]

Found [rename]
Found [rename] under [rename]
Found [rename] under [rename, rename]
Found [rename] under [deep rename]
Found [rename] under [batch hydration]
Found [rename] under [batch hydration, batch hydration]
Found [rename] under [batch hydration, batch hydration, batch hydration]
Found [rename] under [batch hydration, batch hydration, batch hydration, rename]
Found [rename] under [batch hydration, batch hydration, batch hydration, rename, rename]

Found [deep rename]
Found [deep rename] under [rename]
```

This PR aims to improve coverage of nested features. More analysis to be done.

### DIFF
--- a/test/src/test/kotlin/graphql/nadel/tests/transforms/RemoveFieldTestTransform.kt
+++ b/test/src/test/kotlin/graphql/nadel/tests/transforms/RemoveFieldTestTransform.kt
@@ -14,6 +14,7 @@ import graphql.nadel.enginekt.transform.result.NadelResultInstruction
 import graphql.nadel.enginekt.transform.result.json.JsonNodeExtractor
 import graphql.nadel.enginekt.util.queryPath
 import graphql.normalized.ExecutableNormalizedField
+import graphql.schema.GraphQLObjectType
 import graphql.validation.ValidationError
 import graphql.validation.ValidationErrorType
 
@@ -25,18 +26,18 @@ class RemoveFieldTestTransform : NadelTransform<GraphQLError> {
         service: Service,
         overallField: ExecutableNormalizedField,
     ): GraphQLError? {
-        if (overallField.objectTypeNames
-                .map { executionBlueprint.schema.getType(it) }
-                .any { it == null }
-        ) {
-            return null
-        }
+        val objectType = overallField.objectTypeNames.asSequence()
+            .map {
+                executionBlueprint.schema.getType(it) as GraphQLObjectType?
+            }
+            .filterNotNull()
+            .firstOrNull()
+            ?: return null
 
-        if (overallField.getOneFieldDefinition(executionBlueprint.schema)
-                .getDirective("toBeDeleted") != null
-        ) {
+        if (objectType.getField(overallField.name)?.getDirective("toBeDeleted") != null) {
             return ValidationError(ValidationErrorType.WrongType)
         }
+
         return null
     }
 
--- a/engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
+++ b/engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
@@ -12,7 +12,12 @@ import graphql.nadel.enginekt.NadelExecutionContext
 import graphql.nadel.enginekt.blueprint.NadelExecutionBlueprintFactory
 import graphql.nadel.enginekt.plan.NadelExecutionPlan
 import graphql.nadel.enginekt.plan.NadelExecutionPlanFactory
+import graphql.nadel.enginekt.transform.NadelDeepRenameTransform
+import graphql.nadel.enginekt.transform.NadelRenameTransform
 import graphql.nadel.enginekt.transform.NadelTransform
+import graphql.nadel.enginekt.transform.NadelTypeRenameResultTransform
+import graphql.nadel.enginekt.transform.hydration.NadelHydrationTransform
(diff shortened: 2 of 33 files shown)


engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelDeepRenameTransform.kt
```diff
@@ -152,24 +155,34 @@ internal class NadelDeepRenameTransform : NadelTransform<NadelDeepRenameTransfor
         field: ExecutableNormalizedField,
         state: State,
     ): NadelTransformFieldResult {
+        val objectTypesNoRenames = field.objectTypeNames.filterNot { it in state.instructionsByObjectTypeNames }
+
         return NadelTransformFieldResult(
-            newField = null,
-            artificialFields = state.instructions.map { (coordinates, instruction) ->
+            newField = objectTypesNoRenames
```
gnawf, line 161: So a common "bug" we have is that we never consider cases where the field being transformed is in an interface, and that other fields in the interface implementations don't have that transform e.g.

```graphql
interface World {
  hello: String
}
type Earth implements World {
  hello: String
}
type Mars implements World {
  hello: String @renamed(from: "core.hello")
}
```

In this case the underlying query needs to be

```graphql
query {
  world {
    ... on Earth {
      hello
    }
    ... on Mars {
      core {
        hello
      }
    }
  }
}
```

But without this change we only send down


```graphql
query {
  world {
    ... on Mars {
      core {
        hello
      }
    }
  }
}
```

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelDeepRenameTransform.kt
```diff
@@ -58,7 +61,7 @@ internal class NadelDeepRenameTransform : NadelTransform<NadelDeepRenameTransfor
          * }
          * ```
          */
-        val instructions: Map<FieldCoordinates, NadelDeepRenameFieldInstruction>,
+        val instructionsByObjectTypeNames: Map<GraphQLObjectTypeName, NadelDeepRenameFieldInstruction>,
```
gnawf, line 64: Changed the key to be a `String` aliased as `GraphQLObjectTypeName`. Having the key as `FieldCoordinates` never really made sense because the only thing different about any of these coordinates was the type. The field MUST have been the same.

You'll notice this weird behavior documented in some of the deleted code in this PR

```kotlin
    // NOTE: the given instructions must have the same field name, just differing type name
    // Otherwise this function doesn't make sense
    val fieldName = instructions.keys.first().fieldName

    instructions[makeFieldCoordinates(overallTypeName, fieldName)]
```

Whereas now the code is just

```kotlin
    instructions[overallTypeName]
```

test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
```diff
@@ -64,7 +64,11 @@ class EngineTests : FunSpec({
                     // Rename fixtures if they have the wrong name
                     val expectedName = fixture.name.toSlug()
                     if (file.nameWithoutExtension != expectedName) {
-                        file.renameTo(File(file.parentFile, "$expectedName.${file.extension}"))
+                        val expectedFile = File(file.parentFile, "$expectedName.${file.extension}")
```
felipe-gdr, line 67: is that why sometimes my test files were being deleted? 🤔

test/src/test/kotlin/graphql/nadel/tests/transforms/RemoveFieldTestTransform.kt
```diff
@@ -25,18 +26,18 @@ class RemoveFieldTestTransform : NadelTransform<GraphQLError> {
...
         overallField: ExecutableNormalizedField,
     ): GraphQLError? {
-        if (overallField.objectTypeNames
-                .map { executionBlueprint.schema.getType(it) }
-                .any { it == null }
-        ) {
-            return null
-        }
+        val objectType = overallField.objectTypeNames.asSequence()
+            .map {
```
felipe-gdr, line 30: could `mapNonNull` be used instead of `map -> filterNotNull`?

felipe-gdr (approved):

test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
gnawf: Ahh yes very sorry. It was supposed to be a cleanup thing so we could keep names consistent but I didn't think about what would happen when you copy pasted files.

I had a few tests disappear on me too :(

test/src/test/kotlin/graphql/nadel/tests/transforms/RemoveFieldTestTransform.kt
gnawf: Eeh yea but it's in test code so eeeeh let's just leave it as is. Too lazy to get it reapproved 🙃

PR #299, opened 2021-08-12
support field rename with arguments

temaEmelyan:
Seems like renamed fields with arguments were never supported in the nextgen. A bit scary that we never caught it with tests. I only noticed this because some tests in the gateway were failing

### DIFF
--- a/test/src/test/resources/fixtures/deep renames/deep-rename-with-argument-works.yml
+++ b/test/src/test/resources/fixtures/deep renames/deep-rename-with-argument-works.yml
@@ -1,0 +1,103 @@
+name: deep rename with argument works
+enabled:
+  current: false
+  nextgen: true
+overallSchema:
+  IssueService: |
+    service IssueService {
+      type Query {
+        issue: Issue
+      }
+      type Issue {
+        name(userId: ID!): String => renamed from detail.detailName
+      }
+    }
+underlyingSchema:
+  IssueService: |
+    type Issue {
+      detail: IssueDetails
+    }
+
+    type IssueDetails {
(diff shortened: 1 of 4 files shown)


test/src/test/resources/fixtures/deep renames/deep-rename-with-argument-works.yml
```diff
@@ -0,0 +1,103 @@
+name: deep rename with argument works
+enabled:
+  current: false
```
temaEmelyan, line 3: deep rename arguments do not work on the current gen

gnawf (approved):

test/src/test/resources/fixtures/deep renames/deep-rename-with-argument-works.yml
```diff
@@ -0,0 +1,103 @@
...
+    }
+query: |
+  query {
+    issue {
+      name(userId: "USER-01")
+    }
+  }
+variables: {}
+serviceCalls:
+  current:
```
felipe-gdr, line 36: do you need to declare `current` here if the test never runs for the current engine?

felipe-gdr (commented):
Approved

test/src/test/resources/fixtures/deep renames/deep-rename-with-argument-works.yml
temaEmelyan: you don't HAVE TO
I thought it's good to have it in case we want to fix it for the current gen. We already have the service call and the result so we don't need to tinker with the test, can jump straight to debugging

PR #300, opened 2021-08-13
__typename is sent only to service that owns namespace type

felipe-gdr:
Aligns the behaviour of the current gen with next gen.

-  `__typename` for namespace fields will only be added to queries sent to the service that declares the namespaced field.


gnawf (approved):
I think we're missing a test to determine behavior when there's a `__typename` AND multiple namespaced fields going to different services. But that doesn't affect this fix so nbd.

engine/src/main/java/graphql/nadel/engine/execution/NadelExecutionStrategy.java
```diff
@@ -227,9 +228,22 @@ public NadelExecutionStrategy(List<Service> services,
...
-                            .anyMatch(graphQLFieldDefinition -> graphQLFieldDefinition.getName().equals(field.getName()) || field.getName().equals(Introspection.TypeNameMetaFieldDef.getName())));
+            final boolean serviceExtendsNamespaceType = service.getDefinitionRegistry().getDefinitions(ObjectTypeExtensionDefinition.class)
+                    .stream()
+                    .anyMatch(objectTypeDef -> objectTypeDef.getName().equals(namespacedObjectType.getName()));
+
+            Optional<MergedField> maybeNewMergedField = MergedFieldUtil.includeSubSelection(
+                    mergedField,
+                    namespacedObjectType,
+                    executionCtx,
+                    field -> secondLevelFieldDefinitionsForService
```
temaEmelyan, line 239: stretching for a nit here but I think it would help to improve readability a bit if we had 2 helper functions:
`boolean isFieldIncluded(..)` and
`boolean serviceOwnsNamespacedType(..)`

temaEmelyan (dismissed):

temaEmelyan (approved):

gnawf (approved):

PR #302, opened 2021-08-17
Filter type names going to services based on ownership

gnawf:


engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelServiceTypeFilterTransform.kt
```diff
@@ -0,0 +1,144 @@
...
+    private fun getOwnedTypeNames(
+        executionBlueprint: NadelOverallExecutionBlueprint,
+        service: Service,
+    ): Set<String> {
+        return service.underlyingSchema
+            .typeMap
+            .values
+            .asSequence()
+            .map {
+                executionBlueprint.getOverallTypeName(service, it.name)
```
gnawf, line 140: I am slightly worried about accidentally exposing types that aren't meant to be.

However I convinced myself this can't happen because

- This set of types is only used to cross check that a type is valid
- The final object type name list is a subset of the original i.e. we can't be exposing types that weren't meant to be
- The original object type name list is based off the overall schema i.e. type was exposed

The only thing we need to do is add validation that for shared types that are used, the underlying schema has the exact underlying type.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/batch/NadelBatchHydrationObjectIdFieldBuilder.kt
```diff
@@ -0,0 +1,132 @@
...
+ *   }
+ * }
+ * ```
+ *
+ * We need an id to actually identify the objects to know where to insert them back
+ * into the original result.
+ *
+ * i.e. this code inserts the selection for the `id` field
+ */
+internal object NadelBatchHydrationObjectIdFieldBuilder {
```
gnawf, line 55: Moved this code to another class because it was bloating original class with random concerns WRT field buildin

gnawf: I had to edit this function because originally it sent the underlying type name, but we need to be setting the overall type name

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/result/NadelResultTransformer.kt
```diff
@@ -74,7 +74,8 @@ internal class NadelResultTransformer(private val executionBlueprint: NadelOvera
             it.run(context = transformContext)
         }
 
-        transformContext.emptyPaths.forEach(result.data::cleanup)
+        // TODO: investigate whether this code is actually needed. So far no tests fail after disabling it…
+        // transformContext.emptyPaths.forEach(result.data::cleanup)
```
gnawf, line 78: So this code nulls out a `Map` that has been edited in the transformation process _if_ it is empty. I guess nobody actually does that.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/plan/NadelExecutionPlanFactory.kt
```diff
@@ -74,6 +75,7 @@ internal class NadelExecutionPlanFactory(
             return NadelExecutionPlanFactory(
                 executionBlueprint,
                 transforms = listOfTransforms(
+                    NadelServiceTypeFilterTransform(),
```
felipe-gdr, line 78: does the order of transforms matter?

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelServiceTypeFilterTransform.kt
felipe-gdr, line 140: does `getOverallTypeName` handle renamed types?

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelServiceTypeFilterTransform.kt
```diff
@@ -0,0 +1,144 @@
...
+import graphql.nadel.enginekt.util.toBuilder
+import graphql.normalized.ExecutableNormalizedField
+import graphql.normalized.ExecutableNormalizedField.newNormalizedField
+
+class NadelServiceTypeFilterTransform : NadelTransform<State> {
+    data class State(
+        val aliasHelper: NadelAliasHelper,
+        val typeNamesOwnedByService: Set<String>,
+        val fieldObjectTypesOwnedByService: List<String>,
+        val overallField: ExecutableNormalizedField,
```
felipe-gdr, line 28: is `overallField` a different object than `field` passed in as an argument to `transformField`? Is this a mechanism to handle cases where other transforms can change the original field before this transform is executed?

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelServiceTypeFilterTransform.kt
```diff
@@ -0,0 +1,144 @@
...
+        service: Service,
+        overallField: ExecutableNormalizedField,
+        underlyingParentField: ExecutableNormalizedField?,
+        result: ServiceExecutionResult,
+        state: State,
+    ): List<NadelResultInstruction> {
+        return emptyList()
+    }
+
+    private fun getOwnedTypeNames(
```
felipe-gdr, line 131: will this function return **all** types owned by the service OR **all types declared by the field** that are owned by the service?

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelServiceTypeFilterTransform.kt
```diff
@@ -0,0 +1,144 @@
...
+    override suspend fun transformField(
+        executionContext: NadelExecutionContext,
+        transformer: NadelQueryTransformer.Continuation,
+        executionBlueprint: NadelOverallExecutionBlueprint,
+        service: Service,
+        field: ExecutableNormalizedField,
+        state: State,
+    ): NadelTransformFieldResult {
+        // Nothing to query if there are no fields, we need to add selection
+        if (state.fieldObjectTypesOwnedByService.isEmpty()) {
```
felipe-gdr, line 76: So, in case all types have been filtered out we add one artificial field per object type associated with the original field? 
And those artificial fields will later be discarded from the result data?
What's the purpose of this?

test/src/test/resources/fixtures/basic/service-types-are-filtered.yml
```diff
@@ -0,0 +1,174 @@
...
+    }
+  }
+variables: {}
+serviceCalls:
+  current:
+    - serviceName: test
+      request:
+        query: |
+          query nadel_2_service {
+            foo
```
felipe-gdr, line 120: where is `foo` in the schema?

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelServiceTypeFilterTransform.kt
gnawf: Yup exactly what it does. Gets overall type name from underlying type name.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelServiceTypeFilterTransform.kt
gnawf: Yup exactly.

test/src/test/resources/fixtures/basic/service-types-are-filtered.yml
gnawf: Oh yea good point, these tests are actually disabled for the current gen engine. So this section is actually obsolete. I'll delete it.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelServiceTypeFilterTransform.kt
gnawf: I'll rename the function. All types owned by a service.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelServiceTypeFilterTransform.kt
gnawf: Yea it's the same issue we have with removed fields. Basically take this query

```graphql
query {
    test {
        ... on Test {
            id
        }
        ... on QueryError {
            extensions {
                ... on LabQueryErrorExtension {
                    statusCode
                }
            }
        }
    }
}
```

There are two services `Test` and `Lab`. The `LabQueryErrorExtension` gets removed because it doesn't belong to the `Test` service which the query is going to.

In that case we end up with this query

```graphql
query {
    test {
        ... on Test {
            id
        }
        ... on QueryError {
            extensions
        }
    }
}
```

Which isn't legal because `extensions` needs subselections. So in this case we add `__typename` as a placeholder so that we still execute the `extensions` field.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/plan/NadelExecutionPlanFactory.kt
gnawf: Yes a bit. I figured it would be best to put it at the top so that transforms don't actually get the other types that won't end up getting executed. As for whether it's the correct decision, who knows.

I think we may need another layer of query transformation that effectively fixes up issues inside other transforms. Lest we push all that responsibility/rules onto our transform impls.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelServiceTypeFilterTransform.kt
felipe-gdr: I see. thanks for the explanation!

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelServiceTypeFilterTransform.kt
felipe-gdr: Right, so in that case, should we be checking that all items in `overallField.objectTypeNames` are present in `typesOwnedByService` instead of just comparing their sizes?

```
        if (typesOwnedByService.size == overallField.objectTypeNames.size) {
            return null
        }
```

Like, if we have this:

```
typesOwnedByService = [Issue, TodoIssue, PullRequest]
overallField.objectTypeNames = [Issue, TodoIssue]
```

we still want to return `null` here, right?

gnawf: Oh gosh that code is a mess. Sorry I didn't notice that after refactoring. There's actually two variables there `typeNamesOwnedByService` and `typesOwnedByService`. I've refactored it to the following now

```kotlin
        val typeNamesOwnedByService = getTypeNamesServiceOwns(executionBlueprint, service)
        val fieldObjectTypeNamesOwnedByService = overallField.objectTypeNames
            .filter { it in typeNamesOwnedByService }

        // All types are owned by service
        if (fieldObjectTypeNamesOwnedByService.size == overallField.objectTypeNames.size) {
            return null
        }
```

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/batch/NadelBatchHydrationObjectIdFieldBuilder.kt
```diff
@@ -0,0 +1,132 @@
...
+            underlyingType = underlyingParentTypeOfIdField,
+        ) ?: error("Could not find the overall output type for the actor field")
+
+        return resolveObjectTypes(
+            schema = executionBlueprint.schema,
+            type = overallType,
+            onNotObjectType = {
+                errorForUnsupportedObjectIdParentType()
+            },
+        ).map { it.name }.toList()
```
temaEmelyan, line 112: `.toList()`
do we need it? It's already a list I think

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelServiceTypeFilterTransform.kt
```diff
@@ -0,0 +1,193 @@
...
+import graphql.nadel.enginekt.transform.artificial.NadelAliasHelper
+import graphql.nadel.enginekt.transform.query.NadelQueryTransformer
+import graphql.nadel.enginekt.transform.result.NadelResultInstruction
+import graphql.nadel.enginekt.util.resolveObjectTypes
+import graphql.nadel.enginekt.util.toBuilder
+import graphql.normalized.ExecutableNormalizedField
+import graphql.normalized.ExecutableNormalizedField.newNormalizedField
+
+/**
+ * Nadel ends up building a complex schema from multiple different services. This class
```
felipe-gdr, line 19: awesome docs! 👏

felipe-gdr (approved):

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/batch/NadelBatchHydrationObjectIdFieldBuilder.kt
gnawf: Ah yes you're right idk why I did that…

gnawf: Fixed, thanks

temaEmelyan (approved):

temaEmelyan:
great stuff!

felipe-gdr (approved):

PR #303, opened 2021-08-17
Dynamic service resolution on nextgen

felipe-gdr:


engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
```diff
@@ -129,6 +133,22 @@ class NextgenEngine @JvmOverloads constructor(
         return result
     }
 
+    private suspend fun resolveServiceAndExecuteTopLevelField(
+        field: ExecutableNormalizedField,
+        executionContext: NadelExecutionContext,
+        service: Service
+    ): ExecutionResult = if (usesDynamicServiceResolution(topLevelField = field)) {
+        val serviceOrError = serviceExecutionHooks.resolveServiceForField(services.values, field)
```
temaEmelyan, line 141: can `serviceOrError` be `null`?

api/src/main/java/graphql/nadel/hooks/ServiceExecutionHooks.java
```diff
@@ -60,4 +62,8 @@ default NewVariableValue visitArgumentValueInQuery(HooksVisitArgumentValueEnviro
     default ServiceOrError resolveServiceForField(List<Service> services, ExecutionStepInfo executionStepInfo) {
         return null;
     }
+
+    default ServiceOrError resolveServiceForField(Collection<Service> services, ExecutableNormalizedField executableNormalizedField) {
```
temaEmelyan, line 66: add `@Nullable`

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
felipe-gdr: Yep it can. Good catch!

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
```diff
@@ -273,6 +293,24 @@ class NextgenEngine @JvmOverloads constructor(
...
+            }
+            .onEach {
+                Assert.assertTrue(GraphQLTypeUtil.unwrapNonNull(it.type) is GraphQLInterfaceType) {
+                    String.format(
+                        "field annotated with %s directive is expected to be of GraphQLInterfaceType",
+                        NadelDirectives.DYNAMIC_SERVICE_DIRECTIVE_DEFINITION.name
+                    )
+                }
+            }
+            .firstOrNull() != null
```
temaEmelyan, line 312: just FYI, don't have to change: you can use `.any()` to check if a sequence is empty or not

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
```diff
@@ -129,6 +133,22 @@ class NextgenEngine @JvmOverloads constructor(
...
 
+    private suspend fun resolveServiceAndExecuteTopLevelField(
+        field: ExecutableNormalizedField,
+        executionContext: NadelExecutionContext,
+        service: Service
+    ): ExecutionResult = if (usesDynamicServiceResolution(topLevelField = field)) {
+        val serviceOrError = serviceExecutionHooks.resolveServiceForField(services.values, field)
+
+        if (serviceOrError.error != null) {
+            newExecutionResult(error = serviceOrError.error, data = mapOf(field.resultKey to null))
```
temaEmelyan, line 144: I think it better be `mutableMapOf`

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
felipe-gdr: nice ;-)

felipe-gdr: interestingly, if I use `any` the code block in `onEach` doesn't execute and errors are not thrown when expected, strange.

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
felipe-gdr: Hm, the type of `data` in `ExecutionResult` is `Object`. Do consumer expect it to be a mutable data structure?

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
```diff
@@ -129,6 +134,29 @@ class NextgenEngine @JvmOverloads constructor(
         return result
     }
 
+    private suspend fun resolveServiceAndExecuteTopLevelField(
+        field: ExecutableNormalizedField,
+        executionContext: NadelExecutionContext,
+        service: Service
+    ): ExecutionResult {
+        return if (usesDynamicServiceResolution(topLevelField = field)) {
```
gnawf, line 142: Is it possible to move this code into `NadelFieldToService`. In my mind it belongs there. We have the ability to raise exceptions that implement `GraphQLError` to achieve the same effect.

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
```diff
@@ -273,6 +301,24 @@ class NextgenEngine @JvmOverloads constructor(
         )
     }
 
+    private fun usesDynamicServiceResolution(
```
gnawf, line 304: Particularly this code doesn't really belong here. It should be in a sort of `DynamicServiceResulotion` object/class.

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
```diff
@@ -273,6 +301,24 @@ class NextgenEngine @JvmOverloads constructor(
...
+        topLevelField: ExecutableNormalizedField
+    ): Boolean =
+        topLevelField.getFieldDefinitions(overallSchema)
+            .asSequence()
+            .filter {
+                it.getDirective(NadelDirectives.DYNAMIC_SERVICE_DIRECTIVE_DEFINITION.name) != null
+            }
+            .onEach {
+                Assert.assertTrue(GraphQLTypeUtil.unwrapNonNull(it.type) is GraphQLInterfaceType) {
```
gnawf, line 313: We have an extension function for unwrapping so you can do `it.type.unwrapNonNull()`

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
gnawf: Weird, maybe you need a `filterNotNull().any()`?

gnawf (changes requested):
Looks good, would like some code moved to different files, but other than that the code itself is good :)

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
temaEmelyan: currently the `result.data` is always recursively mutable. Not 100% sure whether it's by design or by coincidence, maybe @gnawf or @andimarek can clarify.

gnawf: At this point, I don't think it actually needs to be mutable, but we should set it to mutable anyway.

The reason the data needs to be mutable is because of `NadelResultTransformer` which transforms the JSON structure (Java `Map`s and `List`s) assuming that they're mutable. We don't create new `List`s or `Map`s.

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
felipe-gdr: Yep sounds good. I was worried that I was adding too much code in `NextgenEngine` thanks for the suggestion!

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
```diff
@@ -110,10 +120,15 @@ class NextgenEngine @JvmOverloads constructor(
...
-                                    executeTopLevelField(field, service, executionContext)
+                                    val resolvedService = fieldToService.resolveDynamicService(field, service)
+
+                                    executeTopLevelField(field, resolvedService, executionContext)
                                 } catch (e: Throwable) {
                                     when (e) {
-                                        is GraphQLError -> newExecutionResult(error = e)
+                                        is GraphQLError -> newExecutionResult(
+                                            error = e,
+                                            data = mutableMapOf(field.resultKey to null)
```
felipe-gdr, line 130: This changes general behaviour, not only dynamic resolution related stuff. But I think is for the better. It will ensure a `field: null` property on the data is added when there is a big error that prevented the actual execution from happening.

It didn't break any existing tests too.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/query/NadelFieldToService.kt
```diff
@@ -28,6 +29,21 @@ internal class NadelFieldToService(
         }
     }
 
+    /**
+     * Returns the dynamically resolved service for the field, if it is annotated with @dynamicServiceResolution,
+     * otherwise returns the originalService.
+     */
+    fun resolveDynamicService(
```
gnawf, line 36: Can't we just integrate this into `getServicesForTopLevelFields` as an abstracted detail or do you think that's a bad idea?

e.g.

```kotlin
    fun getServicesForTopLevelFields(query: ExecutableNormalizedOperation): List<NadelFieldAndService> {
        return query.topLevelFields.flatMap { topLevelField ->
            if (dynamicServiceResolution.needsDynamicServiceResolution(field)) { // NEW
                dynamicServiceResolution.resolveServiceForField(field) // NEW
            } else if (isNamespacedField(topLevelField)) {
                getServicePairsForNamespacedFields(topLevelField)
            } else {
                listOf(getServicePairFor(field = topLevelField))
            }
        }
    }
```

That way, we don't need to add code into `NextgenEngine` specifically to deal with that.

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
gnawf: Yup that sounds good.

gnawf (commented):
Just had a question about whether we should simplify `NadelFieldToService`, otherwise good to go.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/query/NadelFieldToService.kt
felipe-gdr: That would simplify the code yeah. 
The only problem is that if something goes wrong in the process of dynamically resolving the service for a field, the whole execution would blow up - which is what happens currently when a regular service association fails in `getServicesForTopLevelFields`.
With the current state of the code, only the field that uses dynamic service resolution would fail, so the execution could still be a partial success.

gnawf: Ahh I see, sorry didn't see this comment yesterday. That makes sense. Kinda annoying but yep, sounds good.

Let's just merge this as is then.

gnawf (approved):

PR #305, opened 2021-08-18
Enable tests for hydration on interface fields in nextgen

felipe-gdr:
As it turns out, no changes were required in Nextgen


gnawf (approved):
Nice

PR #311, opened 2021-09-03
Improve error handling

gnawf:
Surrounds all normal execution with try catch instead of just the query parsing (since we now have instrumentation that can "fail").

### DIFF
--- a/test/src/test/kotlin/graphql/nadel/tests/hooks/execution-is-aborted-when-begin-execute-completes-exceptionally-inside-cf.kt
+++ b/test/src/test/kotlin/graphql/nadel/tests/hooks/execution-is-aborted-when-begin-execute-completes-exceptionally-inside-cf.kt
@@ -28,7 +28,7 @@ import java.util.concurrent.CompletableFuture
 import java.util.concurrent.CompletableFuture.completedFuture
 
 @KeepHook
-class `execution-is-aborted-when-begin-execute-completes-exceptionally` : EngineTestHook {
+class `execution-is-aborted-when-begin-execute-completes-exceptionally-inside-cf` : EngineTestHook {
     var instrumentationParams: NadelInstrumentationExecuteOperationParameters? = null
     var resultBeforeFinalInstrumentation: ExecutionResult? = null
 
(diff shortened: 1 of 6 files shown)


test/src/test/kotlin/graphql/nadel/tests/hooks/execution-is-aborted-when-begin-execute-completes-exceptionally-inside-cf.kt
```diff
@@ -0,0 +1,94 @@
...
+    var instrumentationParams: NadelInstrumentationExecuteOperationParameters? = null
+    var resultBeforeFinalInstrumentation: ExecutionResult? = null
+
+    override fun makeNadel(engineType: NadelEngineType, builder: Nadel.Builder): Nadel.Builder {
+        return builder
+            .instrumentation(object : NadelInstrumentation {
+                override fun createState(parameters: NadelInstrumentationCreateStateParameters): InstrumentationState {
+                    return object : InstrumentationState {
+                        override fun toString(): String {
+                            return "so-annoying"
```
felipe-gdr, line 41: hm, why is this annoying?

test/src/test/resources/fixtures/instrumentation/execution-is-aborted-when-begin-execute-completes-exceptionally-inside-cf.yml
```diff
@@ -0,0 +1,66 @@
...
+        hello: World
+      }
+      type World {
+        id: ID
+        name: String
+      }
+      type Mutation {
+        hello: String
+      }
+      type Subscription {
```
felipe-gdr, line 18: this `Subscription` type doesn't seem to be relevant to test and can be a bit distracting.

felipe-gdr (dismissed):

test/src/test/kotlin/graphql/nadel/tests/hooks/execution-is-aborted-when-begin-execute-completes-exceptionally-inside-cf.kt
gnawf: Ah it's a copy paste from very similar test, I think I had issues with assertions when I was writing the test.

test/src/test/resources/fixtures/instrumentation/execution-is-aborted-when-begin-execute-completes-exceptionally-inside-cf.yml
gnawf: Ah yea, it's a copy paste from another test. I'll remove it.

felipe-gdr (approved):

PR #312, opened 2021-09-03
Fix potential collisions caused by artificial field names

gnawf:


engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/artificial/NadelAliasHelper.kt
```diff
@@ -76,7 +76,7 @@ class NadelAliasHelper private constructor(private val alias: String) {
     companion object {
         fun forField(tag: String, field: ExecutableNormalizedField): NadelAliasHelper {
             // TODO: detect when not in test environment and provide UUID or similar
-            return NadelAliasHelper("${tag}__${field.name}")
+            return NadelAliasHelper("${tag}__${field.resultKey}")
```
gnawf, line 79: This change kinda shows how little tests we have where we combine aliases with transforms. We should add some more. I'll create a ticket for that.

felipe-gdr (approved):

PR #313, opened 2021-09-07
Fix batch hydration by index

gnawf:
So we had an issue where index based hydration was not setting the hydration value for all nodes. This happened when the hydration returned null data (not an empty array, but NULL).

This adds more tests for some scenarios that were broken and ensures behavior is the same as the current engine's.

### DIFF
--- a/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/log/NadelLog.kt
+++ b/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/log/NadelLog.kt
@@ -1,5 +1,3 @@
-@file:JvmName("LogKt")
-
 package graphql.nadel.enginekt.log
 
 import graphql.nadel.util.LogKit
--- a/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/batch/NadelBatchHydrationByIndex.kt
+++ b/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/batch/NadelBatchHydrationByIndex.kt
@@ -1,0 +1,114 @@
+package graphql.nadel.enginekt.transform.hydration.batch
+
+import graphql.nadel.ServiceExecutionResult
+import graphql.nadel.enginekt.blueprint.NadelBatchHydrationFieldInstruction
+import graphql.nadel.enginekt.transform.hydration.NadelHydrationUtil
+import graphql.nadel.enginekt.transform.hydration.NadelHydrationUtil.getNumberOfInputNodes
+import graphql.nadel.enginekt.transform.hydration.batch.NadelBatchHydrationInputBuilder.getBatchInputDef
+import graphql.nadel.enginekt.transform.result.NadelResultInstruction
+import graphql.nadel.enginekt.transform.result.json.JsonNode
+import graphql.nadel.enginekt.util.AnyList
+import graphql.nadel.enginekt.util.isList
+import graphql.nadel.enginekt.util.listOfNulls
+import graphql.nadel.enginekt.util.subListOrNull
+import graphql.nadel.enginekt.util.unwrapNonNull
+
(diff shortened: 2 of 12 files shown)


engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/log/NadelLog.kt
```diff
@@ -0,0 +1,51 @@
+package graphql.nadel.enginekt.log
+
+import graphql.nadel.util.LogKit
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+internal inline fun <reified T> getLogger(): Logger {
```
gnawf, line 7: Sorry slightly unrelated, I worked off a branch I where I started adding debug logging to investigate the issue. Keeping these in because they're useful.

We need to improve logging too:

* We need a StructuredArgsLogger like in the AGG and RB, we aren't logging the request ID or execution ID along with every request and I had to guess which logs matched which requests based on anomalous data/times.
* Maybe others

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/batch/NadelBatchHydrationByIndex.kt
```diff
@@ -0,0 +1,114 @@
...
+import graphql.nadel.enginekt.transform.hydration.batch.NadelBatchHydrationInputBuilder.getBatchInputDef
+import graphql.nadel.enginekt.transform.result.NadelResultInstruction
+import graphql.nadel.enginekt.transform.result.json.JsonNode
+import graphql.nadel.enginekt.util.AnyList
+import graphql.nadel.enginekt.util.isList
+import graphql.nadel.enginekt.util.listOfNulls
+import graphql.nadel.enginekt.util.subListOrNull
+import graphql.nadel.enginekt.util.unwrapNonNull
+
+internal object NadelBatchHydrationByIndex {
```
gnawf, line 16: This is the main change, also moved this code out of the `NadelBatchHydrator` class as there were two batch hydration algorithms in there that were polluting the class.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/batch/NadelBatchHydrationByObjectId.kt
```diff
@@ -0,0 +1,144 @@
...
+import graphql.nadel.enginekt.util.AnyMap
+import graphql.nadel.enginekt.util.JsonMap
+import graphql.nadel.enginekt.util.asNullableJsonMap
+import graphql.nadel.enginekt.util.emptyOrSingle
+import graphql.nadel.enginekt.util.flatten
+import graphql.nadel.enginekt.util.isList
+import graphql.nadel.enginekt.util.queryPath
+import graphql.nadel.enginekt.util.unwrapNonNull
+
+internal object NadelBatchHydrationByObjectId {
```
gnawf, line 25: Nothing changed here, this code is stripped from `NadelBatchHydrator` exactly as it was.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/batch/NadelBatchHydrator.kt
```diff
@@ -76,171 +63,25 @@ internal class NadelBatchHydrator(
...
                 instruction = instruction,
                 parentNodes = parentNodes,
-                resultNodes = resultNodes,
+                batches = batches,
                 matchStrategy = matchStrategy,
             )
         } + getInstructionsToAddErrors(batches)
     }
 
-    private fun getHydrateInstructionsMatchingIndex(
```
gnawf, line 99: This is the code that was extracted into different classes.

temaEmelyan (approved):

PR #314, opened 2021-09-08
Fix case where query document has multiple operation definitions

gnawf:


engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/util/CollectionUtil.kt
```diff
@@ -16,7 +16,7 @@ inline fun <reified T> Collection<*>.singleOfTypeOrNull(predicate: (T) -> Boolea
  * Like [singleOrNull] but the single item must be of type [T].
  */
 inline fun <reified T> Collection<*>.singleOfType(predicate: (T) -> Boolean = { true }): T {
-    return singleOfTypeOrNull(predicate) as T
```
gnawf, line 19: Ok this is where it went "wrong", kinda. I thought casting to a not null type would also do a null check, but no. It only does a plain Java cast. Then the error went undetected.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/util/GraphQLUtil.kt
```diff
@@ -334,9 +349,10 @@ internal suspend fun NadelInstrumentation.beginExecute(
         queryDocument,
         graphQLSchema,
         executionInput.variables,
-        queryDocument.definitions.singleOfType(),
+        queryDocument.getOperationDefinitionOrNull(executionInput.operationName)
```
gnawf, line 352: Query document _can_ contain multiple operations. At this point in the execution, we have already validated that the user specified a valid `operationName`, so we just need to pick it out of the document.

test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
```diff
@@ -179,17 +178,18 @@ private suspend fun execute(
                 }
             })
             .let {
-                testHooks?.makeNadel(engineType, it) ?: it
+                testHooks.makeNadel(engineType, it)
```
gnawf, line 181: Random cleanup from a while back when I changed `testHooks` to be not-nullable.

test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
```diff
@@ -179,17 +178,18 @@ private suspend fun execute(
...
             }
             .build()
 
         val response = nadel.execute(
             newNadelExecutionInput()
                 .query(fixture.query)
                 .variables(fixture.variables)
+                .operationName(fixture.operationName)
```
gnawf, line 189: Tests can now specify operation name.

test/src/test/kotlin/graphql/nadel/tests/hooks/can-execute-document-with-multiple-operation-definitions.kt
```diff
@@ -0,0 +1,40 @@
...
+    override fun makeNadel(engineType: NadelEngineType, builder: Nadel.Builder): Nadel.Builder {
+        return builder
+            .instrumentation(object : NadelInstrumentation {
+                override fun beginExecute(
+                    parameters: NadelInstrumentationExecuteOperationParameters,
+                ): CompletableFuture<InstrumentationContext<ExecutionResult>> {
+                    hasCalledBeginExecute = true
+
+                    expectThat(parameters.operationDefinition)
+                        .get { name }
```
gnawf, line 29: Before, this would be null because we assumed there was only a single operation name.

temaEmelyan (approved):

PR #315, opened 2021-09-09
Add randomness to aliases

gnawf:


engine/src/test/groovy/graphql/nadel/e2e/nextgen/NadelE2ETest.groovy
```diff
@@ -1,1047 +0,0 @@
-package graphql.nadel.e2e.nextgen
```
gnawf, line 1: This is a really old file that predates the new test format we had.

buildSrc/src/main/kotlin/graphql/nadel/Deps.kt
```diff
@@ -0,0 +1,10 @@
+package graphql.nadel
+
+object Deps {
```
gnawf, line 3: Added common declarations for libs.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/artificial/AliasGenerator.kt
```diff
@@ -0,0 +1,41 @@
...
+import java.util.UUID
+
+object AliasGenerator {
+    /**
+     * Random substrings to attach to the field alias to make it unique enough such
+     * that nobody should ever guess the field result key and mess with the fields
+     * introduced by Nadel.
+     */
+    private val suffixes = Array(100) {
+        UUID.randomUUID().toString().substringAfterLast("-")
```
gnawf, line 13: Decided to cut it short because we don't need a full UUID in the request and response bloating the size.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/artificial/AliasGenerator.kt
```diff
@@ -0,0 +1,41 @@
...
+    fun getAlias(tag: String, field: ExecutableNormalizedField): String {
+        val staticAlias = getStaticAlias(tag, field)
+        if (isStatic) {
+            return staticAlias
+        }
+        /**
+         * Note that even without the suffix the alias here is unique among its siblings fields.
+         * See [suffixes] for more.
+         */
+        return "${staticAlias}__${suffixes.random()}"
```
gnawf, line 32: Picks a random UUID instead instead of generating one every time. Even if the same one is picked it doesn't matter as the prefix is already unique enough.

As stated in the doco, the suffix is just so outsiders/incoming queries don't mess with Nadel's artificial fields.

test/src/test/resources/fixtures/deep renames/different-deep-renames-on-same-normalized-field.yml
```diff
@@ -0,0 +1,111 @@
+name: different deep renames on same normalized field
```
gnawf, line 1: Pulled this test out of the deleted `NadelE2ETests`

test/src/test/resources/fixtures/hydration/hydration-with-deep-rename.yml
```diff
@@ -0,0 +1,124 @@
+name: hydration with deep rename
```
gnawf, line 1: Pulled this test out of the deleted `NadelE2ETests`

PR #317, opened 2021-09-09
Fix namespaced if first service call result returns null

gnawf:


test/src/test/resources/fixtures/namespaced/one-of-the-namespaced-services-returns-error-reversed.yml
```diff
@@ -0,0 +1,202 @@
...
+      "issue": {
+        "search": {
+          "count": 1
+        }
+      }
+    },
+    "errors": [
+      {
+        "message": "Error on IssueSearch",
+        "locations": [],
```
felipe-gdr, line 195: I wonder if `locations` here should point to `data/issue` or `data/issue/getIssue`. There's very little detail about this error in the response, with a large query this could be a bit confusing to consumers.
Just a thought. I don't think this is in the scope of this PR.

felipe-gdr (approved):

test/src/test/resources/fixtures/namespaced/one-of-the-namespaced-services-returns-error-reversed.yml
gnawf: I mean the error is just forwarded from the underlying service. We have a `locations` only because we parse it into `GraphQLError` then back into `Map` -> `String`.

If an underlying service returns N GraphQL errors, it'll be hard to figure out what belongs where.

PR #318, opened 2021-09-09
Make result transformation async

gnawf:
I think we forgot to do this in other places too.

Without this change hydrations etc wouldn't block the thread, BUT they would be sequential.


gnawf:
Replacing with #353

PR #319, opened 2021-09-09
Fix null batch hydration

gnawf:
So turns out there's a distinction between explicitly null and absent. Explicitly null sets the value to null, and absent sets an empty array.

Doesn't really make sense to me as absent is effectively null… but that's just how the current engine handles it…


engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/batch/NadelBatchHydrationByObjectId.kt
```diff
@@ -84,17 +84,14 @@ internal object NadelBatchHydrationByObjectId {
...
-                        when (val id = parentNodeIdentifierNode.value) {
-                            null -> emptySequence()
-                            is AnyList -> id.asSequence().flatten(recursively = true)
-                            else -> sequenceOf(id)
-                        }
-                    }
+                    .asSequence()
+                    .map { it.value }
+                    .flatten(recursively = true)
+                    .filterNotNull()
```
gnawf, line 94: Just a refactor, it's the same thing.

temaEmelyan (approved):

PR #320, opened 2021-09-13
Fix hydration input edge cases

gnawf:
Fixes behavior when plain (NOT batch) hydration input is null.

### DIFF
--- a/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/NadelHydrationInputBuilder.kt
+++ b/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/NadelHydrationInputBuilder.kt
@@ -1,14 +1,6 @@
 package graphql.nadel.enginekt.transform.hydration
 
-import graphql.language.ArrayValue
-import graphql.language.BooleanValue
-import graphql.language.FloatValue
-import graphql.language.IntValue
 import graphql.language.NullValue
-import graphql.language.ObjectField
-import graphql.language.ObjectValue
-import graphql.language.StringValue
-import graphql.language.Value
 import graphql.nadel.enginekt.blueprint.NadelHydrationFieldInstruction
 import graphql.nadel.enginekt.blueprint.hydration.NadelHydrationActorInputDef
 import graphql.nadel.enginekt.blueprint.hydration.NadelHydrationActorInputDef.ValueSource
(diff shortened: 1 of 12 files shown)


engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/NadelHydrationInputBuilder.kt
```diff
@@ -1,172 +1,159 @@
...
-        parentNode: JsonNode,
-    ): List<Map<String, NormalizedInputValue>> {
-        val inputDefsForAllCalls = instruction.actorInputValueDefs.asSequence()
-            .let { defs ->
-                when (val hydrationStrategy = instruction.hydrationStrategy) {
-                    is NadelHydrationStrategy.OneToOne -> defs
-                    is NadelHydrationStrategy.ManyToOne -> defs.filter { def -> def != hydrationStrategy.inputDefToSplit }
-                }
-            }
+    private fun build(): List<Map<String, NormalizedInputValue>> {
```
gnawf, line 37: Ok I ended up refactoring this code while I was at it because I was never happy with it. I think I am now.

I suggest reading the class by itself instead of looking at the diff. It's almost completely changed.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/NadelHydrationInputBuilder.kt
```diff
@@ -1,172 +1,159 @@
...
 
-internal typealias AnyAstValue = Value<*>
+internal class NadelHydrationInputBuilder private constructor(
+    private val instruction: NadelHydrationFieldInstruction,
+    private val aliasHelper: NadelAliasHelper,
+    private val fieldToHydrate: ExecutableNormalizedField,
+    private val parentNode: JsonNode,
+) {
+    companion object {
+        fun getInputValues(
```
gnawf, line 26: I think we should follow this pattern more.

One "static" function to take in all the arguments. Create a class that holds some global ish variables that would otherwise need to be passed around to all functions.

This REALLY clears up what a function really needs and what you need to pass around e.g.

```kotlin
    private fun makeInputValue(
        inputDef: NadelHydrationActorInputDef,
    ): NormalizedInputValue? {
```

Is much, much better than

```kotlin
    private fun makeInputValue(
        actorInputDef: NadelHydrationActorInputDef,
        argumentDef: GraphQLArgument,
        parentNode: JsonNode,
        hydrationField: ExecutableNormalizedField,
        aliasHelper: NadelAliasHelper,
    ): NormalizedInputValue? {
```

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/NadelHydrationInputBuilder.kt
```diff
@@ -1,172 +1,159 @@
...
-                        actorInputDef = actorInputDef,
-                        argumentDef = argumentDef,
-                        parentNode = parentNode,
-                        hydrationField = hydratedField,
-                        aliasHelper = aliasHelper,
-                    ) ?: return@mapNotNull null
-                    actorInputDef.name to inputValue
-                }
-                .toList()
+    private fun makeOneToOneArgs(): List<Map<String, NormalizedInputValue>> {
```
gnawf, line 44: Ok so if you're not aware, we support two types of plain hydration.

The usual

```graphql
type Query {
  userById(id: ID): User
}
type Issue {
  userId: ID
  user: User @hydrated(from: "userById" arguments: [
    {name: "id" valueFromField: "userId"}
  ])
}
```

i.e. one to one. But then we also have


```graphql
type Query {
  userById(id: ID): User
}
type Issue {
  userIds: [ID] # NOTE the array here
  user: User @hydrated(from: "userById" arguments: [
    {name: "id" valueFromField: "userIds"}
  ])
}
```

That is, for N `userIds` we need to send N queries to the underlying service i.e. many to one.

Not a great pattern because you should really be using batching, but we support it.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/NadelHydrationInputBuilder.kt
```diff
@@ -1,172 +1,159 @@
...
+            }
+            .toMapStrictly()
+    }
+
+    /**
+     * Valid in the sense that we should actually query for data.
+     *
+     * If all field values are null, then it doesn't makes sense to actually send the query.
+     */
+    private fun isInputMapValid(inputMap: Map<String, NormalizedInputValue>): Boolean {
```
gnawf, line 99: So this is the main change, we check if all the field values are null.

If all the values are null, then unless the underlying service is doing some weird thing, we won't be getting back any data at all.

We can always remove this, this is mainly to just appease the comparison algorithm by matching the current engine.

test/src/test/resources/fixtures/hydration/hydration-input-is-empty-list.yml
```diff
@@ -20,13 +20,19 @@ overallSchema:
       }
       type Issue {
         id: ID
-        authors: [User] => hydrated from UserService.usersByIds(ids: $source.authorIds) object identified by id, batch size 2
+        author: [User] @hydrated(
+          service: "UserService"
+          field: "userById"
+          arguments: [
+            {name: "id" value: "$source.authorIds"}
```
felipe-gdr, line 27: That's an interesting scenario - the source is a list of values, and the hydration field takes a single value. I'm guessing we would strongly be against consumers using this sort of config in a real-life case?
Also, what would happen if `$source.authorIds` returned more than 1 item?

test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
```diff
@@ -29,7 +29,7 @@ import java.util.concurrent.CompletableFuture
  * 2. Test name e.g. hydration inside a renamed field
  * 3. Copy paste output from selecting a test in the IntelliJ e.g. java:test://graphql.nadel.tests.EngineTests.current hydration inside a renamed field
  */
-private val singleTestToRun = ""
+private val singleTestToRun = (System.getenv("TEST_NAME") ?: "")
```
felipe-gdr, line 32: :-)

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/NadelHydrationInputBuilder.kt
felipe-gdr: Ah ok, this answers the question I've asked bellow (I usually review the tests first).

felipe-gdr (approved):

test/src/test/resources/fixtures/hydration/hydration-input-is-empty-list.yml
gnawf: Yea we would be strongly against it. But that's another story. We just ported this functionality over from Nadel. If we want, we can look into deprecating it in the future.

PR #322, opened 2021-09-14
Handle skip and include directives

gnawf:
Ready

### DIFF
--- a/api/src/main/java/graphql/nadel/Nadel.java
+++ b/api/src/main/java/graphql/nadel/Nadel.java
@@ -70,6 +70,7 @@ public class Nadel {
     final ServiceExecutionFactory serviceExecutionFactory;
     final List<Service> services;
     final GraphQLSchema overallSchema;
+    @NotNull
     final NadelInstrumentation instrumentation;
     final ServiceExecutionHooks serviceExecutionHooks;
     final PreparsedDocumentProvider preparsedDocumentProvider;
--- a/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/skipInclude/SkipIncludeTransform.kt
+++ b/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/skipInclude/SkipIncludeTransform.kt
@@ -19,6 +19,18 @@ import graphql.normalized.ExecutableNormalizedField
 import graphql.normalized.ExecutableNormalizedField.newNormalizedField
 import graphql.schema.GraphQLSchema
 
+/**
+ * So the way `@skip` and `@include` work is that in the [graphql.normalized.ExecutableNormalizedOperationFactory]
+ * is that they get automatically removed by the factory. Because of that, we never actually
+ * get the fields. This causes bad outcomes for Nadel because we don't execute the query here.
+ * We forward the query and we are _not_ allowed generate invalid queries with empty
+ * selection sets.
+ *
+ * So here, we add back in a `__typename` field to ensure we don't have an empty selection.
+ *
+ * This should probably be a more generic "if no subselections add an empty one for removed fields".
+ * But we'll deal with that separately.
+ */
 internal class SkipIncludeTransform : NadelTransform<State> {
     companion object {
         private const val skipFieldName = "__skip"
--- a/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/skipInclude/SkipIncludeTransform.kt
+++ b/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/skipInclude/SkipIncludeTransform.kt
@@ -1,0 +1,125 @@
+package graphql.nadel.enginekt.transform.skipInclude
+
+import graphql.execution.MergedField
+import graphql.introspection.Introspection
+import graphql.language.Field
+import graphql.nadel.Service
+import graphql.nadel.ServiceExecutionResult
+import graphql.nadel.enginekt.NadelExecutionContext
+import graphql.nadel.enginekt.blueprint.NadelOverallExecutionBlueprint
+import graphql.nadel.enginekt.transform.NadelTransform
+import graphql.nadel.enginekt.transform.NadelTransformFieldResult
+import graphql.nadel.enginekt.transform.artificial.NadelAliasHelper
+import graphql.nadel.enginekt.transform.query.NadelQueryTransformer
+import graphql.nadel.enginekt.transform.result.NadelResultInstruction
+import graphql.nadel.enginekt.transform.skipInclude.SkipIncludeTransform.State
+import graphql.nadel.enginekt.util.resolveObjectTypes
+import graphql.nadel.enginekt.util.toBuilder
+import graphql.normalized.ExecutableNormalizedField
+import graphql.normalized.ExecutableNormalizedField.newNormalizedField
+import graphql.schema.GraphQLSchema
+
+internal class SkipIncludeTransform : NadelTransform<State> {
+    companion object {
+        private const val skipFieldName = "__skip"
+    }
+
+    class State(
+        val aliasHelper: NadelAliasHelper,
+    )
+
+    override suspend fun isApplicable(
+        executionContext: NadelExecutionContext,
+        executionBlueprint: NadelOverallExecutionBlueprint,
+        services: Map<String, Service>,
+        service: Service,
+        overallField: ExecutableNormalizedField,
+    ): State? {
+        // This hacks together a child that will pass through here
+        if (overallField.children.isEmpty()) {
+            val mergedField = executionContext.query.getMergedField(overallField)
+            if (hasAnyChildren(mergedField)) {
+                // Adds a field so we can transform it
+                overallField.children.add(createSkipField(executionBlueprint.schema, overallField))
+            }
+        }
+
+        return if (overallField.name == skipFieldName) {
+            State(
+                aliasHelper = NadelAliasHelper.forField(
+                    tag = "skip_include",
+                    field = overallField,
+                ),
+            )
+        } else {
+            null
+        }
+    }
+
+    override suspend fun transformField(
+        executionContext: NadelExecutionContext,
+        transformer: NadelQueryTransformer.Continuation,
+        executionBlueprint: NadelOverallExecutionBlueprint,
+        service: Service,
+        field: ExecutableNormalizedField,
+        state: State,
+    ): NadelTransformFieldResult {
+        return NadelTransformFieldResult(
+            newField = null,
+            artificialFields = listOf(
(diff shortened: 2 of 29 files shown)


engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/skipInclude/SkipIncludeTransform.kt
```diff
@@ -0,0 +1,125 @@
...
+        services: Map<String, Service>,
+        service: Service,
+        overallField: ExecutableNormalizedField,
+    ): State? {
+        // This hacks together a child that will pass through here
+        if (overallField.children.isEmpty()) {
+            val mergedField = executionContext.query.getMergedField(overallField)
+            if (hasAnyChildren(mergedField)) {
+                // Adds a field so we can transform it
+                overallField.children.add(createSkipField(executionBlueprint.schema, overallField))
```
gnawf, line 43: Ok so Nadel transforms right now aren't designed to transform the children to add artificial fields.

It's not impossible but it's a bit of a mess right now with hydrations etc.

So what this does here is it hacks back in the skipped field that was removed by the normalised query factory. The query traversal will then immediately pick is up and pass it back into `isApplicable`

The new field is added with a specific name that we check, then we transform that field to automatically remove it.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/skipInclude/SkipIncludeTransform.kt
```diff
@@ -0,0 +1,125 @@
...
+        executionContext: NadelExecutionContext,
+        executionBlueprint: NadelOverallExecutionBlueprint,
+        services: Map<String, Service>,
+        service: Service,
+        overallField: ExecutableNormalizedField,
+    ): State? {
+        // This hacks together a child that will pass through here
+        if (overallField.children.isEmpty()) {
+            val mergedField = executionContext.query.getMergedField(overallField)
+            if (hasAnyChildren(mergedField)) {
```
gnawf, line 41: So the normalised field builder automatically removes fields that resolve to `@skip(if: true)`. So we check if the field is currently empty, then check if the original merged field _had_ children

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/skipInclude/SkipIncludeTransform.kt
```diff
@@ -0,0 +1,125 @@
...
+        executionContext: NadelExecutionContext,
+        transformer: NadelQueryTransformer.Continuation,
+        executionBlueprint: NadelOverallExecutionBlueprint,
+        service: Service,
+        field: ExecutableNormalizedField,
+        state: State,
+    ): NadelTransformFieldResult {
+        return NadelTransformFieldResult(
+            newField = null,
+            artificialFields = listOf(
```
gnawf, line 69: Marks field for deletion and changes name accordingly to ask for `__typename`

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/skipInclude/SkipIncludeTransform.kt
```diff
@@ -0,0 +1,125 @@
...
+import graphql.nadel.enginekt.transform.query.NadelQueryTransformer
+import graphql.nadel.enginekt.transform.result.NadelResultInstruction
+import graphql.nadel.enginekt.transform.skipInclude.SkipIncludeTransform.State
+import graphql.nadel.enginekt.util.resolveObjectTypes
+import graphql.nadel.enginekt.util.toBuilder
+import graphql.normalized.ExecutableNormalizedField
+import graphql.normalized.ExecutableNormalizedField.newNormalizedField
+import graphql.schema.GraphQLSchema
+
+internal class SkipIncludeTransform : NadelTransform<State> {
```
gnawf, line 22: So this transform handles the edge case where all fields are deleted. I'll update the class with docs to mention this.

Fields are automatically removed by the executable normalised factory _if_ the `@skip` or `@include` directives are true or false respectively. Because the field is entirely removed, we can end up sending an empty selection set _if_ all the fields are removed.

We obviously can't send an empty selection set, so we need to add a `__typename` selection.

test/src/test/resources/fixtures/skip-include-fields/hydration/handles-skip-directive-on-hydrated-field.yml
```diff
@@ -0,0 +1,71 @@
+name: handles skip directive on hydrated field
```
felipe-gdr, line 1: this test is not testing `@skip`

test/src/test/resources/fixtures/skip-include-fields/hydration/handles-skip-directive-on-field-with-batch-hydrated-parent.yml
```diff
@@ -0,0 +1,125 @@
+name: handles skip directive on field with batch hydrated parent
```
felipe-gdr, line 1: this is also not testing `@skip`

test/src/test/resources/fixtures/skip-include-fields/hydration/handles-skip-directive-on-batch-hydrated-field.yml
```diff
@@ -0,0 +1,83 @@
+name: handles skip directive on batch hydrated field
```
felipe-gdr, line 1: not testing `@skip`

test/src/test/resources/fixtures/skip-include-fields/hydration/handles-skip-directive-field-with-hydrated-parent.yml
```diff
@@ -0,0 +1,101 @@
+name: handles skip directive field with hydrated parent
```
felipe-gdr, line 1: skip

test/src/test/resources/fixtures/skip-include-fields/handles-skip-directive-on-single-top-level-field.yml
```diff
@@ -0,0 +1,31 @@
+name: handles skip directive on single top level field
+enabled:
+  current: false # Not fixing
+  nextgen: true
+overallSchema:
+  service: |
+    service service {
```
felipe-gdr, line 7: nit: do we need this `service service {` ?

test/src/test/resources/fixtures/skip-include-fields/handles-include-directive-on-single-field-whose-parent-returns-null.yml
```diff
@@ -0,0 +1,60 @@
...
+  service: |
+    type Query {
+      foo: Foo
+    }
+    type Foo {
+      id: String
+    }
+query: |
+  query ($test: Boolean!) {
+    foo {
```
felipe-gdr, line 23: do we have a test for when a parent field `@skip` resolves to `true`, and a child field `@skip` resolves to `false`? This should result in the parent being removed

api/src/main/java/graphql/nadel/Nadel.java
```diff
@@ -70,6 +70,7 @@
     final ServiceExecutionFactory serviceExecutionFactory;
     final List<Service> services;
     final GraphQLSchema overallSchema;
+    @NotNull
```
felipe-gdr, line 73: Interesting. Does placing the annotation here has the same effect of placing it in the constructor?

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/skipInclude/SkipIncludeTransform.kt
felipe-gdr: So we're more or less adding the field as a child of itself so it will be picked up by a following step of the traverser?
Couldn't we just return a non-null `State` at this point? 🤔

felipe-gdr (changes requested):
Requesting changes because of the test descriptions that don't match the directive being tested. Other than that it looks good!
Good test coverage!!!

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/skipInclude/SkipIncludeTransform.kt
gnawf: Nah unfortunately not. The issue is is with hydrations. If we return a non-null `State` here, then we're targetting the _parent_ field e.g.

```
{
  root {
    thing {
      child @skip(if: true)
    }
  }
}
```

The issue is hydration. Say `thing` is the hydration field, then it kinda just deletes its portion of the tree and no transforms get applied. When it deletes the tree, it executes it again as the children of the hydration and does the entire process from the top down i.e. goes back to `NextgenEngine`.

BUT since this transformation runs _before_ hydration, we will insert the `__typename` when looking at `thing`, which hydration then deletes and moves to a completely new execution. The new execution doesn't have context that the `__typename` is an artificial field.

Anyway sorry if that's a horrible explanation. Kinda complex issue with the way transforms work.

test/src/test/resources/fixtures/skip-include-fields/handles-skip-directive-on-single-top-level-field.yml
gnawf: Nope, removed for all skip/include tests.

api/src/main/java/graphql/nadel/Nadel.java
gnawf: Uh, tbh idk I only put it there because I used it from Kotlin code. But then I deleted that code. So I'm just leaving this in.

test/src/test/resources/fixtures/skip-include-fields/handles-include-directive-on-single-field-whose-parent-returns-null.yml
gnawf: Ok done in https://github.com/atlassian-labs/nadel/commit/eac63b3e7fcba68e30c97daba5035d21f11019f3

Also added test for with default values for the variables.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/skipInclude/SkipIncludeTransform.kt
felipe-gdr: Thanks for explaining it Franklin

felipe-gdr (approved):

temaEmelyan (approved):

PR #323, opened 2021-09-14
Change logic for defining owner of namespaced field for __typename querying

felipe-gdr:


test/src/test/resources/fixtures/namespaced/typename-is-sent-to-owning-service-even-when-no-fields-are-queried.yml
```diff
@@ -0,0 +1,199 @@
+name: typename is sent to owning service even when no fields are queried
```
felipe-gdr, line 1: Added a test where the query doesn't contain any fields from the service that owns the namespaced field

temaEmelyan: I think would be good to have a test with the same scenario that surfaced the bug. The namespaced type is extended in the same service where it's defined. What do you think?

gnawf: Yep having both this test and the one @temaEmelyan suggested would be great.

temaEmelyan (approved):

engine/src/main/java/graphql/nadel/engine/execution/NadelExecutionStrategy.java
```diff
@@ -258,11 +258,11 @@ private static boolean fieldMatchesDefinition(GraphQLFieldDefinition graphQLFiel
     }
 
     private static boolean serviceOwnsNamespacedField(GraphQLObjectType namespacedObjectType, Service service) {
-        final boolean serviceExtendsNamespaceType = service.getDefinitionRegistry().getDefinitions(ObjectTypeExtensionDefinition.class)
+        return service.getDefinitionRegistry().getDefinitions(ObjectTypeDefinition.class)
                 .stream()
+                // the type can't be an extension in the owning service
+                .filter(objectTypeDef -> !(objectTypeDef instanceof ObjectTypeExtensionDefinition))
```
gnawf, line 264: Ah, I guess that's why you did the no object ext instead.

gnawf (approved):

engine/src/main/java/graphql/nadel/engine/execution/NadelExecutionStrategy.java
felipe-gdr: Yep

PR #324, opened 2021-09-14
Handle namespaced __typename internally

gnawf:
I think this is ready, but just marking it as a draft because I don't really want to merge this as of yet. It may screw around with the replay mechanism we have for request validation.

### DIFF
--- a/test/src/test/resources/fixtures/namespaced/typename-is-resolved-even-when-namespaced-type-is-extended-in-declaring-service.yml
+++ b/test/src/test/resources/fixtures/namespaced/typename-is-resolved-even-when-namespaced-type-is-extended-in-declaring-service.yml
@@ -1,0 +1,122 @@
+name: "typename is resolved even when namespaced type is extended in declaring service"
+enabled: true
+overallSchema:
+  Issues: |
+    directive @namespaced on FIELD_DEFINITION
+
+    type Query {
+      issue: IssueQuery @namespaced
+    }
+
+    type IssueQuery {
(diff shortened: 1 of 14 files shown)


test/src/test/resources/fixtures/namespaced/typename-is-resolved-even-when-namespaced-type-is-extended-in-declaring-service.yml
```diff
@@ -0,0 +1,122 @@
+name: "typename is resolved even when namespaced type is extended in declaring service"
```
gnawf, line 1: These tests are duplicates of existing tests with the execution hint turned on.

The only difference is that the `__typename` has been removed from the expected service query and result.

test/src/test/resources/fixtures/namespaced/typename-is-resolved-even-when-namespaced-type-is-extended-in-declaring-service.yml
```diff
@@ -0,0 +1,122 @@
...
+          }
+        },
+        "extensions": {}
+      }
+# language=JSON
+response: |-
+  {
+    "data": {
+      "issue": {
+        "__typename": "IssueQuery",
```
gnawf, line 115: Final result still sees `__typename` but the service call never has it in the query nor response.

temaEmelyan (approved):

felipe-gdr (approved):

PR #325, opened 2021-09-15
Fix batch hydration when input nodes are null

gnawf:
We had an issue when hydration input nodes were null. We don't ask the underlying service for null, so we asked for less, but we expected a value for the null. In expecting more, we accidentally throw an exception when we shouldn't have.

See the new test case for more.


engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/batch/NadelBatchHydrationByIndex.kt
```diff
@@ -2,46 +2,77 @@ package graphql.nadel.enginekt.transform.hydration.batch
...
         return parentNodes
             .map { parentNode ->
+                val inputValues = getInputValues(parentNode)
+
                 NadelResultInstruction.Set(
-                    subjectPath = parentNode.resultPath + state.hydratedField.resultKey,
+                    subjectPath = parentNode.resultPath + fieldToHydrate.resultKey,
                     newValue = if (isManyInputNodesToParentNodes) {
-                        chunker.take(
-                            n = getNumberOfInputNodes(
```
gnawf, line 32: So with this change, we still don't ask the underlying service for null nodes. But instead of blindly taking N nodes that the parent node needs, we pass in the input nodes to the `Chunker` which returns extra nulls if it needs to.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/batch/NadelBatchHydrationByIndex.kt
```diff
@@ -75,21 +117,22 @@ internal object NadelBatchHydrationByIndex {
         private val values: List<Any?> = getValues(batches)
         private var cursorIndex = 0
 
-        fun take(): Any? {
+        fun take(inputValues: List<Any?>): List<Any?> {
+            return inputValues.map(::takeOne)
+        }
+
+        fun takeOne(inputValue: Any?): Any? {
+            if (inputValue == null) {
```
gnawf, line 125: Return null if the input node was null (i.e. null means not sent to underlying service). Also does NOT advance the cursor for the list hydration results.

felipe-gdr (approved):

PR #326, opened 2021-09-15
Align how __typename fields are processed for namespaced

gnawf:
BEFORE this change

To determine who gets the __typename
* Nextgen looks at who defines the namespaced __field__ in the overall schema
* Current gen looks at who defines (not extends) the namespaced __type__

AFTER this change both engines look at who defines (not extends) the namespaced __type__.

### DIFF
--- a/test/src/test/resources/fixtures/namespaced/typename-is-sent-owning-service-even-when-namespaced-field-and-type-are-defined-in-different-services.yml
+++ b/test/src/test/resources/fixtures/namespaced/typename-is-sent-owning-service-even-when-namespaced-field-and-type-are-defined-in-different-services.yml
@@ -1,0 +1,160 @@
+name: typename is sent owning service even when namespaced field and type are defined in different services
+enabled:
+  nextgen: true
+  current: true
+overallSchema:
+  Issues: |
+    directive @namespaced on FIELD_DEFINITION
+
+    type IssueQuery {
+      getIssue: Issue
+    }
+
+    extend type IssueQuery {
+      getIssues: [Issue]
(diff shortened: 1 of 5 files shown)


test/src/test/resources/fixtures/namespaced/typename-is-sent-owning-service-even-when-namespaced-field-and-type-are-defined-in-different-services.yml
```diff
@@ -0,0 +1,160 @@
+name: typename is sent owning service even when namespaced field and type are defined in different services
+enabled:
+  nextgen: true
+  current: true
+overallSchema:
+  Issues: |
+    directive @namespaced on FIELD_DEFINITION
+
+    type IssueQuery {
```
gnawf, line 9: Namespaced type is defined here.

test/src/test/resources/fixtures/namespaced/typename-is-sent-owning-service-even-when-namespaced-field-and-type-are-defined-in-different-services.yml
```diff
@@ -0,0 +1,160 @@
...
+      getIssues: [Issue]
+    }
+
+    type Issue {
+      id: ID
+      text: String
+    }
+  IssueSearch: |
+    type Query {
+      issue: IssueQuery @namespaced
```
gnawf, line 23: Namespaced field is defined here in another service.

felipe-gdr (approved):

PR #327, opened 2021-09-17
Attempt to fix shared types

gnawf:


engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/NadelExecutionBlueprintFactory.kt
```diff
@@ -57,16 +66,20 @@ private class Factory(
 ) {
     private val definitionNamesToService: Map<String, Service> = makeDefinitionNamesToService()
     private val coordinatesToService: Map<FieldCoordinates, Service> = makeCoordinatesToService()
+    private val typeRenameInstructions = makeTypeRenameInstructions()
+    private val typesToServices: Map<String, Set<Service>> =
+        TypeOwnership(overallSchema, services, typeRenameInstructions)
```
gnawf, line 71: Ok so what we're doing here is actually figuring out which services define a given type.

Before this change, we never considered an overall type that was shared across multiple services i.e. `definitionNamesToService: Map<String, Service>`

But now we have `typesToService: Map<String, Set<Service>>`, please fix the naming for me, or do something.

The issue with the `Map<String, Service>` was that in the shared service we have, we never define underlying types. So in this factory, we could never find the underlying type if it was shared. So here we store all possible services and look at _one_ given underlying type.

---

Somewhere else, we should be validating that ALL underlying schemas are valid WRT to their overall type i.e. if the SHARED overall type is

```graphql
type Test {
  id: ID
 name: String
}
```

We don't have two underlying definitions that are:

```graphql
# in service A
type Test {
  id: ID
  name: String
}
# in service B
type Test {
  id: ID
}
```

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/NadelExecutionBlueprintFactory.kt
```diff
@@ -171,9 +184,11 @@ private class Factory(
                     return@mapNotNull null
                 }
 
-                val underlyingParentType = getUnderlyingType(hydratedFieldParentType)
-                    ?: error("No underlying type for: ${hydratedFieldParentType.name}")
-                val fieldDefs = underlyingParentType.getFieldsAlong(inputValueDef.valueSource.queryPathToField.segments)
+                val fieldDefs = mapFirstNotNullFromUnderlyingType(hydratedFieldParentType) { underlyingParentType ->
```
gnawf, line 187: Look at `mapFirstNotNullFromUnderlyingType`'s documentation.

So basically we're finding the first underlying type that has `getFieldsAlong(inputValueDef.valueSource.queryPathToField.segments)` as NOT null, and getting that value.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/NadelExecutionBlueprintFactory.kt
```diff
@@ -567,3 +599,138 @@ private class SharedTypesAnalysis(
...
+    fun getTypesThatServiceDefines(): Map<Service, Set<String>> {
+        return services
+            .map { service ->
+                service to getTypeThatServiceDefined(service)
+            }
+            .toMapStrictly()
+    }
+
+    private fun getTypeThatServiceDefined(service: Service): Set<String> {
+        val typesDefined = service.definitionRegistry.definitions
```
gnawf, line 641: Looks at all the types the service contributes to the overall schema.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/NadelExecutionBlueprintFactory.kt
```diff
@@ -567,3 +599,138 @@ private class SharedTypesAnalysis(
...
+            .asSequence()
+            .filterIsInstance<AnyNamedNode>()
+            .filterNot { it.isExtensionDef }
+            .map { it.name }
+            .toHashSet()
+
+        val rawTypesUsedAsOutputTypes = service.definitionRegistry.definitions
```
gnawf, line 648: Looks at all the types the service uses as output types for fields.

e.g. consider `PageInfo` which is defined in the shared service. Every service uses it, but it will never show up it their `Service.definitionRegistry` because it has already been defined once in the overall schema. We need to include that in the types the service defines.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/NadelExecutionBlueprintFactory.kt
```diff
@@ -567,3 +599,138 @@ private class SharedTypesAnalysis(
...
+                it.value // Get the List<Pair<…>> out
+                    .asSequence()
+                    .map {
+                        it.second // Get the Service out of the Pair<String, Service>
+                    }
+                    .toSet()
+            }
+    }
+
+    fun getTypesThatServiceDefines(): Map<Service, Set<String>> {
```
gnawf, line 632: Perhaps a better name is types that a service exposes. Because of shared types, we can't rely on `service.definitionRegistry`, see commentary below for more.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/NadelExecutionBlueprintFactory.kt
```diff
@@ -567,3 +599,138 @@ private class SharedTypesAnalysis(
...
+            .filter {
+                // If the definition is an extension, only consider it if it is extending a type the service owns
+                it.isNotExtensionDef || it.name in typesDefined
+            }
+            .filterIsInstance<AnyImplementingTypeDefinition>()
+            .flatMap {
+                it.fieldDefinitions
+            }
+            .filterNot {
+                isHydrationField(it)
```
gnawf, line 660: Output types of hydrated fields must NOT be considered. A service can reference ANY type for a hydration and never define it in their service.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/NadelExecutionBlueprintFactory.kt
```diff
@@ -567,3 +599,138 @@ private class SharedTypesAnalysis(
...
+            }
+            .map {
+                it.type.unwrapAll().name
+            }
+            .toHashSet()
+
+        // Resolve the types to reachable types i.e. interfaces to their implementations, unions etc.
+        val outputTypesResolved = rawTypesUsedAsOutputTypes
```
gnawf, line 668: Takes the output types and resolves abstract types like interfaces and unions to their possible types.

test/src/test/kotlin/graphql/nadel/tests/CentralSchemaTesting.kt
```diff
@@ -0,0 +1,199 @@
...
+import java.io.File
+import java.util.concurrent.CompletableFuture
+import kotlin.system.measureTimeMillis
+
+/**
+ * Run this and replace the path to central schema to load it up in Nadel.
+ */
+suspend fun main() {
+    val schema = File(
+        "/Users/fwang/Documents/GraphQL/graphql-central-schema/schema/",
```
gnawf, line 26: If you want to feed central schema straight into Nadel, use this class and change this path to your repo.

test/src/test/kotlin/graphql/nadel/tests/CentralSchemaTesting.kt
```diff
@@ -0,0 +1,199 @@
...
+                }
+            }
+
+            override fun getUnderlyingTypeDefinitions(serviceName: String): TypeDefinitionRegistry {
+                return SchemaParser().parse(underlyingSchemas[serviceName] ?: return TypeDefinitionRegistry())
+            }
+        })
+        .build()
+
+    var totalTime = 0L
```
gnawf, line 88: You can delete everything below here if it starts breaking or something. It's to mock service calls and execute a query.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/NadelExecutionBlueprintFactory.kt
```diff
@@ -280,8 +296,9 @@ private class Factory(
                     val pathToField = remoteArgDef.remoteArgumentSource.path
                     NadelHydrationActorInputDef.ValueSource.FieldResultValue(
                         queryPathToField = NadelQueryPath(pathToField),
-                        fieldDefinition = getUnderlyingType(hydratedFieldParentType)
-                            ?.getFieldAt(pathToField)
+                        fieldDefinition = mapFirstNotNullFromUnderlyingType(hydratedFieldParentType) { underlyingParentType ->
```
gnawf, line 299: We're getting the first underlying type that has `getFieldAt(pathToField)` as NOT null and fetching that value.

gnawf:
There's no tests on this. So to finish it up, add tests and clean up the code (if you so desire, I was in a rush to finish this).

gnawf:
Ahh I notice there's an issue with schema transformation deleting fields. You'll have to fix the code to ignore that.

See https://github.com/atlassian-labs/nadel/blob/a66bd82942ffe27dfd2d71076afc88c42044a07b/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/NadelExecutionBlueprintFactory.kt#L537-L539 and the `can-delete-fields-and-types.yml` test and its TEST HOOK.

gnawf:
Fixing in #404 instead.

PR #329, opened 2021-09-27
Nextgen coerces like current gen

felipe-gdr:
Current gen currently coerces the value of scalars using the type defined in the overall schema.

This PR reproduces this behaviour in Nextgen

### DIFF
--- a/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelCoerceTransform.kt
+++ b/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelCoerceTransform.kt
@@ -5,18 +5,61 @@ import graphql.nadel.ServiceExecutionResult
 import graphql.nadel.enginekt.NadelExecutionContext
 import graphql.nadel.enginekt.blueprint.NadelOverallExecutionBlueprint
 import graphql.nadel.enginekt.transform.NadelCoerceTransform.State
-import graphql.nadel.enginekt.transform.query.NadelQueryPath.Companion.root
 import graphql.nadel.enginekt.transform.query.NadelQueryTransformer
 import graphql.nadel.enginekt.transform.result.NadelResultInstruction
-import graphql.nadel.enginekt.transform.result.json.JsonNode
-import graphql.nadel.enginekt.transform.result.json.JsonNodeExtractor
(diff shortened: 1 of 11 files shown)


engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelCoerceTransform.kt
```diff
@@ -0,0 +1,98 @@
...
+                        it,
+                        overallField.fieldName
+                    )
+                )
+            }
+            .map { it.type }
+            .toList()
+
+        if (types.distinct().size != 1) {
+            // field type on all definitions should be the same
```
felipe-gdr, line 48: maybe I should put an "assert" here to throw an error instead?

temaEmelyan: don't think we can throw an error (at least not as it is currently), for example for a field `Foo.__typename` `disticntTypes` is 0

temaEmelyan: also I think `distinctTypes` will be equal to 2 for some crazy edge cases like this:

```graphql
type Query {
  user: User
}
interface User {
  id: ID
  name: String
  user: User
}

type AUser implements User {
  id: ID
  name: String
  user: AUser
}

type BUser implements User {
  id: ID
  name: String
  user: BUser
}

```

temaEmelyan: I think we need to remove the comment or update it

felipe-gdr: Good call @temaEmelyan.  I'll add more tests for the `__typename` and the interface scenarios and update the comment 👍

felipe-gdr: @temaEmelyan I've pushed a commit with a better comment and more documentation as well.

temaEmelyan: very nice!

PR #334, opened 2021-10-17
Upgrade test syntax

gnawf:
Just refactored tests, so not replying to PR template.

FWIW, this was all done using Regex, so there should be no mistakes. Plus tests still pass so surely that's something.


gnawf:
There are a few tests still using the old syntax e.g.

- repeated-hydrated-fields-on-the-same-level.yml
- repeated-hydrated-fields-on-the-same-level-overlapping-fields-in-the-query.yml
- repeated-hydrated-fields-on-the-same-level-when-using-batch-hydration.yml

That's because they're actually bugged. Bugs don't exist in nextgen so not bothering with a fix for now.

PR #339, opened 2021-11-02
Fix rename transform chaining

gnawf:
This transform was made a long time ago before we had transform chaining in mind. What it does is it just takes the overall field and renames it. But it doesn't take into consideration the new field from transformations before it.

We will actually have a similar issue with hydrations. Will fix that in another PR.

- [X] Add tests that use __typename in queries
- [X] Does this change work with all nadel transformations (rename, type rename, hydration, etc)? Add tests for this.
- [X] Is it worth using hints for this change in order to be able to enable a percentage rollout?
No
Yes, will add
- [X] Do we need a pollinator check for this?
No

### DIFF
--- a/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelDeepRenameTransform.kt
+++ b/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelDeepRenameTransform.kt
@@ -69,7 +69,7 @@ internal class NadelDeepRenameTransform : NadelTransform<NadelDeepRenameTransfor
         /**
          * Stored for easy access in other functions.
          */
-        val field: ExecutableNormalizedField,
+        val overallField: ExecutableNormalizedField,
     )
 
     /**
--- a/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelRenameTransform.kt
+++ b/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelRenameTransform.kt
@@ -73,7 +73,7 @@ internal class NadelRenameTransform : NadelTransform<State> {
             } else {
                 null
             },
-            artificialFields = makeRenamedFields(state, transformer, executionBlueprint).let {
+            artificialFields = makeRenamedFields(state, transformer, field, executionBlueprint).let {
                 when (val typeNameField = makeTypeNameField(state)) {
                     null -> it
                     else -> it + typeNameField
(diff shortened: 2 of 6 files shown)


engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelDeepRenameTransform.kt
```diff
@@ -69,7 +69,7 @@ internal class NadelDeepRenameTransform : NadelTransform<NadelDeepRenameTransfor
         /**
          * Stored for easy access in other functions.
          */
-        val field: ExecutableNormalizedField,
+        val overallField: ExecutableNormalizedField,
```
gnawf, line 72: No bug here, but renamed the field to be clear.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelRenameTransform.kt
```diff
@@ -116,7 +117,7 @@ internal class NadelRenameTransform : NadelTransform<State> {
                 transformer,
                 executionBlueprint,
                 state.service,
-                state.overallField,
```
gnawf, line 119: This is the fix.

test/src/test/kotlin/graphql/nadel/tests/hooks/chain-rename-transform.kt
```diff
@@ -0,0 +1,86 @@
...
+                    })
+                    .build(),
+            )
+        }
+
+        if (field.normalizedArguments["continent"] != null) {
+            return NadelTransformFieldResult(
+                newField = field.toBuilder()
+                    .normalizedArguments(field.normalizedArguments.let {
+                        it + ("continent" to NormalizedInputValue("Continent", EnumValue("Asia")))
```
gnawf, line 52: These transform the argument values of renamed fields.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelRenameTransform.kt
```diff
@@ -93,34 +99,46 @@ internal class NadelRenameTransform : NadelTransform<State> {
...
-                rename = instruction,
-            )
-        }
+        val setOfFieldObjectTypeNames = field.objectTypeNames.toSet()
+        return state.instructionsByObjectTypeNames
+            .asSequence()
+            .asFlow() // For coroutines
+            .filter { (typeName) ->
+                // Don't insert type renames for fields that were never asked for
+                typeName in setOfFieldObjectTypeNames
```
gnawf, line 128: This is the other fix.

temaEmelyan (approved):

PR #344, opened 2021-11-10
Fix duplicate transforms

gnawf:

- [X] Add tests that use __typename in queries
Not applicable. This is a general change, there are existing `__typename` tests that pass.
- [X] Does this change work with all nadel transformations (rename, type rename, hydration, etc)? Add tests for this.
This is basically a refactor. So yes. Will try to add more tests.
- [X] Is it worth using hints for this change in order to be able to enable a percentage rollout?
No
- [X] Do we need to add integration tests for this change in the graphql gateway?
Yes 
- [X] Do we need a pollinator check for this?
No


engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
```diff
@@ -58,6 +64,7 @@ class NextgenEngine @JvmOverloads constructor(
     private val logNotSafe = getNotPrivacySafeLogger<NextgenEngine>()
     private val log = getLogger<NextgenEngine>()
 
+    private val coroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
```
gnawf, line 67: Ok I'm sneaking in this change. Coroutines need a "scope" or an effectively an owner to be launched from. We were previously using `GlobalScope` which basically means nobody is managing the coroutines.

This isn't as necessary for us, since once a Nadel instance is "closed" by the AGG we don't send requests to it anymore.

But this can be useful for launching long standing tasks e.g. periodically doing something, but that something is tied to the Nadel instance.

This change is really just housekeeping.

engine-nextgen/src/main/kotlin/graphql/nadel/NextgenEngine.kt
```diff
@@ -138,10 +150,7 @@ class NextgenEngine @JvmOverloads constructor(
                                         executeTopLevelField(field, resolvedService, executionContext)
                                     } catch (e: Throwable) {
                                         when (e) {
-                                            is GraphQLError -> newExecutionResult(
-                                                data = mutableMapOf(field.resultKey to null),
-                                                error = e,
-                                            )
+                                            is GraphQLError -> newExecutionErrorResult(field, error = e)
```
gnawf, line 153: Refactor

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/NadelCoerceTransform.kt
```diff
@@ -70,49 +71,42 @@ internal class NadelCoerceTransform : NadelTransform<State> {
...
+                    ),
                 )
             }
             .map { it.type }
-            .map { GraphQLTypeUtil.unwrapAll(it) }
-            .toList()
-
-        val distinctTypes = unwrappedTypes.distinct().count()
+            .map(GraphQLType::unwrapAll)
+            .toSet()
```
gnawf, line 90: Sorry this is unrelated refactor.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/query/NadelQueryTransformer.kt
```diff
@@ -4,24 +4,51 @@ import graphql.nadel.Service
 import graphql.nadel.enginekt.NadelExecutionContext
 import graphql.nadel.enginekt.blueprint.NadelOverallExecutionBlueprint
 import graphql.nadel.enginekt.plan.NadelExecutionPlan
+import graphql.nadel.enginekt.transform.NadelTransform
 import graphql.nadel.enginekt.transform.NadelTransformFieldResult
 import graphql.nadel.enginekt.util.toBuilder
 import graphql.normalized.ExecutableNormalizedField
 
-class NadelQueryTransformer internal constructor(
+class NadelQueryTransformer private constructor(
```
gnawf, line 12: Ok so a lot of the change in this class is a refactor. The big issue with `NadelQueryTransformer` is that it used to be that it was PER Nadel instance. Now created it's per query.

This means that for a query, we store the `Service` and `ExecutionContext` etc. associated with that query as global vars. This greatly cleans things up and means instead of passing a `Continuation` object to the transforms, we can just expose this class as API.

e.g.

```kotlin
private suspend fun transformField(
    transformContext: TransformContext,
    executionPlan: NadelExecutionPlan,
    service: Service,
): List<ExecutableNormalizedField>
```

is now

```kotlin
suspend fun transform(
    fields: List<ExecutableNormalizedField>,
): List<ExecutableNormalizedField>
```

because all those params are now stored as global props on the class.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/query/NadelQueryTransformer.kt
```diff
@@ -36,145 +63,113 @@ class NadelQueryTransformer internal constructor(
...
+                .objectTypeNames(getUnderlyingTypeNames(it.objectTypeNames))
+                .build()
+        }
+
+        val newField = listOfNotNull(
+            transformResult.newField?.let {
+                it.toBuilder()
+                    .clearObjectTypesNames()
+                    .objectTypeNames(getUnderlyingTypeNames(it.objectTypeNames))
+                    .children(transform(it.children))
```
gnawf, line 110: Ok so this is the actual fix. Previously we would make the transforms themselves do this.

Issue is, say you have this

```
class AriArgumentTransform {
  fun transformField() {
    return Result(
      field = field.toBuilder()
        .arguments(…) // Interpret ARIs
        .children(transformer.transform(field.children)) // You NEEDED this
        .build()
    )
  }
}
```

We MUST call `transformer.transform` on children because the children can have renames, hydrations etc.

BUT what happens if the `AriArgumentTransform` is NOT the only transform on the field. What if the field with an ARI argument is renamed. Then we have BOTH the `AriArgumentTransform` AND `RenameTransform` transforming the children. Which is not good.

gnawf (commented):
Will provide more context later.

temaEmelyan (approved):

temaEmelyan (approved):

PR #351, opened 2021-11-30
pass alias helper to getHydrationInstruction hook

temaEmelyan:
currently we are passing a parent node to the hook (which is a json map) which is something like this:
```json
{
  "__typename": "Foo"
  "__typename__batch_hydration__data": "Foo"
  "batch_hydration__data__dataId": "ari:cloud:something/PET-0"
  "id": "FOO-0"
}
```

In the hook we look at every hydration instruction and we need to understand which one is the appropriate one based on the value of the hydration argument (value of "batch_hydration__data__dataId" field in this example). Just by looking at the `parentNode` json map alone, we don't know that `batch_hydration__data__dataId` is the special field that we need to be examining. So I propose we pass the alias helper as well and then inside the hook we will know that `batch_hydration__data__dataId` is the special field that we need to look at. 

I tested it with locally deployed nadel here: https://bitbucket.org/atlassian/graphql-gateway/pull-requests/891#chg_internal/src/main/java/io/atlassian/graphql/gateway/hooks/GatewayServiceExecutionHooks.java_newline307

Alternatively what we could do is to pass these artificial fields to the hook https://github.com/atlassian-labs/nadel/blob/master/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/batch/NadelBatchHydrationTransform.kt#L82. This is a list of `ExecutableNormalizedField` and it has the field name and the alias. So in the hook we know that we need to find the value of the field `dataId` (based on the hydration instruction), we check what the alias is for this artificial field (it's `batch_hydration__data__dataId`) and then we know what field we need to look in the json map. So it's similar but different.

### DIFF
--- a/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/NadelEngineExecutionHooks.kt
+++ b/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/NadelEngineExecutionHooks.kt
@@ -1,12 +1,14 @@
 package graphql.nadel.enginekt
 
 import graphql.nadel.enginekt.blueprint.NadelGenericHydrationInstruction
+import graphql.nadel.enginekt.transform.artificial.NadelAliasHelper
 import graphql.nadel.enginekt.transform.result.json.JsonNode
 import graphql.nadel.hooks.ServiceExecutionHooks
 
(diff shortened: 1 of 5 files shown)


gnawf (approved):
Seems reasonable, let's get this in place if polymorphic hydration calls for it

PR #356, opened 2022-01-06
Parallelize hydration more

gnawf:
Was digging into logs and noticed there's still some sequential-ness to the hydrations.

### DIFF
--- a/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/NadelHydrationTransform.kt
+++ b/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/NadelHydrationTransform.kt
@@ -30,6 +30,7 @@ import graphql.nadel.enginekt.util.toBuilder
 import graphql.nadel.hooks.ServiceExecutionHooks
 import graphql.normalized.ExecutableNormalizedField
 import graphql.schema.FieldCoordinates
+import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.coroutineScope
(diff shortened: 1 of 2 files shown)


engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/NadelHydrationTransform.kt
```diff
@@ -146,15 +147,21 @@ internal class NadelHydrationTransform(
...
 
-        return parentNodes.flatMap {
-            hydrate(
-                parentNode = it,
-                state = state,
-                executionBlueprint = executionBlueprint,
-                fieldToHydrate = overallField,
-                executionContext = executionContext,
-            )
+        val jobs: List<Deferred<List<NadelResultInstruction>>> = coroutineScope {
```
gnawf, line 150: So this is the main one, it does hydrations one by one here.

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/transform/hydration/batch/NadelBatchHydrator.kt
```diff
@@ -35,31 +36,40 @@ internal class NadelBatchHydrator(
...
                 pairs.value.map { pair -> pair.first }
             }
 
-        return parentNodesByInstruction.flatMap { (instruction, parentNodes) ->
-            if (instruction == null) parentNodes.map {
-                NadelResultInstruction.Set(
-                    it.resultPath + state.hydratedField.fieldName,
-                    null
-                )
+        val jobs: List<Deferred<List<NadelResultInstruction>>> = coroutineScope {
```
gnawf, line 56: This one is still needed, but doesn't have as much of an impact. We already parallelize the batches later down, this parallelizes it if there are different instructions per field e.g.

```
interface Test {
  something: Thing
}
type TestA implements Test {
  something: Thing @hydrated
}
type TestB implements Test {
  something: Thing @hydrated
}
```

Then this would speed that up.

temaEmelyan (approved):

PR #363, opened 2022-01-21
send json scalar arguments as variables to underlying services

temaEmelyan:

### DIFF
--- a/api/src/main/java/graphql/nadel/normalized/ExecutableNormalizedOperationToAstCompiler.java
+++ b/api/src/main/java/graphql/nadel/normalized/ExecutableNormalizedOperationToAstCompiler.java
@@ -1,0 +1,189 @@
+package graphql.nadel.normalized;
+
+import graphql.Internal;
+import graphql.com.google.common.collect.ImmutableList;
+import graphql.com.google.common.collect.ImmutableMap;
+import graphql.language.Argument;
+import graphql.language.ArrayValue;
+import graphql.language.Document;
+import graphql.language.Field;
+import graphql.language.InlineFragment;
+import graphql.language.NullValue;
+import graphql.language.ObjectField;
+import graphql.language.ObjectValue;
+import graphql.language.OperationDefinition;
+import graphql.language.Selection;
+import graphql.language.SelectionSet;
+import graphql.language.TypeName;
+import graphql.language.Value;
+import graphql.nadel.normalized.ValueToVariableValueCompiler.VariableValueWithDefinition;
+import graphql.normalized.ExecutableNormalizedField;
+import graphql.normalized.NormalizedInputValue;
+import graphql.schema.GraphQLSchema;
+import kotlin.Pair;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import static graphql.collect.ImmutableKit.map;
+import static graphql.language.Argument.newArgument;
+import static graphql.language.Field.newField;
+import static graphql.language.InlineFragment.newInlineFragment;
+import static graphql.language.SelectionSet.newSelectionSet;
+import static graphql.language.TypeName.newTypeName;
+import static graphql.nadel.normalized.ValueToVariableValueCompiler.normalizedInputValueToVariable;
+
+@Internal
+public class ExecutableNormalizedOperationToAstCompiler {
+
+    private static final String JSON_SCALAR_TYPENAME = "JSON";
+
+    public static Pair<Document, Map<String, Object>> compileToDocument(
+            GraphQLSchema schema,
+            OperationDefinition.Operation operationKind,
+            String operationName,
+            List<ExecutableNormalizedField> topLevelFields
+    ) {
+        List<VariableValueWithDefinition> variables = new ArrayList<>();
+        List<Selection<?>> selections = selectionsForNormalizedFields(schema, topLevelFields, variables);
+        SelectionSet selectionSet = new SelectionSet(selections);
+
+        OperationDefinition.Builder definitionBuilder = OperationDefinition.newOperationDefinition()
+                .name(operationName)
+                .operation(operationKind)
+                .selectionSet(selectionSet);
+
+        definitionBuilder.variableDefinitions(map(variables, (variableWithDefinition -> variableWithDefinition.definition)));
+
+        Map<String, Object> variableValuesByNames = variables.stream()
+                .collect(Collectors.toMap(
+                        variableWithDefinition -> variableWithDefinition.definition.getName(),
+                        variableWithDefinition -> variableWithDefinition.value
+                ));
+
+        return new Pair<>(
+                Document.newDocument()
+                        .definition(definitionBuilder.build())
+                        .build(),
+                variableValuesByNames
+        );
+    }
+
+    private static List<Selection<?>> selectionsForNormalizedFields(
+            GraphQLSchema schema,
+            List<ExecutableNormalizedField> executableNormalizedFields,
+            List<VariableValueWithDefinition> variables
+    ) {
+        ImmutableList.Builder<Selection<?>> selections = ImmutableList.builder();
+
+        // All conditional fields go here instead of directly to selections so they can be grouped together
+        // in the same inline fragement in the output
+        Map<String, List<Field>> conditionalFieldsByObjectTypeName = new LinkedHashMap<>();
+
+        for (ExecutableNormalizedField nf : executableNormalizedFields) {
+            Map<String, List<Field>> groupFieldsForChild = selectionForNormalizedField(schema, nf, variables);
+            if (nf.isConditional(schema)) {
+                groupFieldsForChild.forEach((objectTypeName, fields) -> {
+                    List<Field> fieldList = conditionalFieldsByObjectTypeName.computeIfAbsent(objectTypeName, ignored -> new ArrayList<>());
+                    fieldList.addAll(fields);
+                });
+            } else {
+                List<Field> fields = groupFieldsForChild.values().iterator().next();
+                selections.addAll(fields);
+            }
+        }
+
+        conditionalFieldsByObjectTypeName.forEach((objectTypeName, fields) -> {
+            TypeName typeName = newTypeName(objectTypeName).build();
+            InlineFragment inlineFragment = newInlineFragment().
+                    typeCondition(typeName)
+                    .selectionSet(selectionSet(fields))
+                    .build();
+            selections.add(inlineFragment);
+        });
+
+        return selections.build();
+    }
+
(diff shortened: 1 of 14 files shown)


api/src/main/java/graphql/nadel/normalized/ValueToVariableValueCompiler.java
```diff
@@ -0,0 +1,107 @@
...
+        if (value instanceof FloatValue) {
+            return ((FloatValue) value).getValue();
+        }
+        if (value instanceof IntValue) {
+            return ((IntValue) value).getValue();
+        }
+        if (value instanceof BooleanValue) {
+            return ((BooleanValue) value).isValue();
+        }
+        return null; //todo
```
temaEmelyan, line 88: need to double check all possible types here

api/src/main/java/graphql/nadel/normalized/ExecutableNormalizedOperationToAstCompiler.java
```diff
@@ -0,0 +1,189 @@
...
+        }
+        if (value == null) {
+            return NullValue.newNullValue().build();
+        }
+        return (Value<?>) value;
+    }
+
+    @NotNull
+    private static Value<?> argValue(NormalizedInputValue normalizedInputValue, List<VariableValueWithDefinition> variableDefinitionAccumulator) {
+        if (JSON_SCALAR_TYPENAME.equals(normalizedInputValue.getUnwrappedTypeName())) {
```
gnawf, line 181: Nit: flip this around?

```
normalizedInputValue.getUnwrappedTypeName().equals(JSON_SCALAR_TYPENAME)
```

api/src/main/java/graphql/nadel/normalized/ExecutableNormalizedOperationToAstCompiler.java
```diff
@@ -0,0 +1,189 @@
...
+import static graphql.language.SelectionSet.newSelectionSet;
+import static graphql.language.TypeName.newTypeName;
+import static graphql.nadel.normalized.ValueToVariableValueCompiler.normalizedInputValueToVariable;
+
+@Internal
+public class ExecutableNormalizedOperationToAstCompiler {
+
+    private static final String JSON_SCALAR_TYPENAME = "JSON";
+
+    public static Pair<Document, Map<String, Object>> compileToDocument(
```
gnawf, line 45: I'm going to assume the only thing you changed in this class after copying was make the return `Pair` and update `argValue` to generate variables? (+ new variables List)

api/src/main/java/graphql/nadel/normalized/ValueToVariableValueCompiler.java
```diff
@@ -0,0 +1,107 @@
...
+    static VariableValueWithDefinition normalizedInputValueToVariable(NormalizedInputValue normalizedInputValue) {
+        Object variableValue = null;
+        Object normalizedInputValueValue = normalizedInputValue.getValue();
+        if (normalizedInputValueValue instanceof ObjectValue) {
+            variableValue = toVariableValue((ObjectValue) normalizedInputValueValue);
+        }
+        if (normalizedInputValueValue instanceof ArrayValue) {
+            variableValue = toVariableValues(((ArrayValue) normalizedInputValueValue).getValues());
+        }
+        if (normalizedInputValueValue instanceof List) {
```
gnawf, line 34: Man I hate how `normalizedInputValue.getValue` gives you an `Object` or and not a `Value`.

Do you need to handle `Map`s here?

api/src/main/java/graphql/nadel/normalized/ValueToVariableValueCompiler.java
gnawf: Here's a list of `Value` implementations according to IntelliJ

ArrayValue
BooleanValue
EnumValue
FloatValue
IntValue
NullValue
ObjectValue
ScalarValue
StringValue
VariableReference

Once you're done with that, we should throw an exception at the bottom instead of returning null. We wouldn't want to silently send a malformed query.

api/src/main/java/graphql/nadel/normalized/ValueToVariableValueCompiler.java
```diff
@@ -0,0 +1,107 @@
...
+            return ((IntValue) value).getValue();
+        }
+        if (value instanceof BooleanValue) {
+            return ((BooleanValue) value).isValue();
+        }
+        return null; //todo
+    }
+
+    private static String getVarName() {
+        return "var_" + UUID.randomUUID().toString().replace("-", "_");
```
gnawf, line 92: Hmmm, would it be ok to just use the current size of `variableDefinitionAccumulator` to define this? e.g. `var0`, `var1` etc.

api/src/test/groovy/graphql/nadel/normalized/ExecutableNormalizedOperationToAstCompilerTest.groovy
```diff
@@ -0,0 +1,696 @@
...
+        when:
+        def document = compileToDocument(schema, MUTATION, null, fields).first
+        then:
+        AstPrinter.printAst(document) == '''mutation {
+  foo1(arg: {arg1 : "Mutation"})
+}
+'''
+    }
+
+    def "test JSON when input is a variable"() {
```
gnawf, line 321: Can you move these to the bottom of the file and have like a separator comment? so the merge is easier in the future when we eventually move this code back to GraphQL Java.

e.g.

```
def "GraphQL Java tests above"() {
}

// --------------------------------------------------------------------------------------
// Custom JSON handling

def "test JSON when input is a variable"() {
}

…
```

gnawf (commented):
Nice, looks great. Happy to approve once you're done.

api/src/main/java/graphql/nadel/normalized/ValueToVariableValueCompiler.java
```diff
@@ -0,0 +1,107 @@
...
+    private static List<Object> toVariableValues(List<Value> values) {
+        ArrayList<Object> objects = new ArrayList<>();
+        for (Value value : values) {
+            objects.add(toVariableValue(value));
+        }
+        return objects;
+    }
+
+    private static Map<String, Object> toVariableValue(ObjectValue objectValue) {
+        HashMap<String, Object> map = new HashMap<>();
```
andimarek, line 57: I recommend to always use `LinkedHashMap` as default implementation to keep the insertion order which helps with debugging and testing.

api/src/main/java/graphql/nadel/normalized/ValueToVariableValueCompiler.java
andimarek, line 92: this looks like overkill as we only need to be unique for this query. UUID.randomUUID() has also a non zero cpu cost.
Can we just use a counter?

api/src/main/java/graphql/nadel/normalized/ValueToVariableValueCompiler.java
andimarek: +1 for throwing an exception at the end.
Some scalars are not relevant as we rely here on a specific JSON implementation: there will never be a VariableReference for example.

api/src/main/java/graphql/nadel/normalized/ExecutableNormalizedOperationToAstCompiler.java
andimarek: actually I recommend this style as it prevents any NPE from `getUnwrappedTypeName` returning `null`

api/src/main/java/graphql/nadel/normalized/ValueToVariableValueCompiler.java
temaEmelyan: so I believe `EnumValue`, `ScalarValue`, `VariableReference` are the only ones that are not relevant

api/src/main/java/graphql/nadel/normalized/ValueToVariableValueCompiler.java
temaEmelyan: so I believe EnumValue, ScalarValue, VariableReference are the only ones that are not relevant

api/src/main/java/graphql/nadel/normalized/ExecutableNormalizedOperationToAstCompiler.java
temaEmelyan: that's correct

api/src/main/java/graphql/nadel/normalized/ValueToVariableValueCompiler.java
temaEmelyan: done

api/src/test/groovy/graphql/nadel/normalized/ExecutableNormalizedOperationToAstCompilerTest.groovy
temaEmelyan: done

api/src/main/java/graphql/nadel/normalized/ValueToVariableValueCompiler.java
temaEmelyan: > Do you need to handle Maps here?
for args like this for example
```
type Mutation {
    foo1(arg1: JSON!, arg2: [JSON!]): String
}
```

`arg1` is going to be an `ObjectValue` and `arg2` is a `List`

similarly for input types
```
input InputWithJson {
    id: ID
    json: JSON
    jsons: [JSON]
}
```
`json` is an `ObjectValue` and jsons is a `List`

api/src/main/java/graphql/nadel/normalized/ValueToVariableValueCompiler.java
temaEmelyan: these types are never returned from https://github.com/graphql-java/graphql-java-extended-scalars/blob/master/src/main/java/graphql/scalars/object/ObjectScalar.java#L107

test/src/test/resources/fixtures/hydration/query-with-hydrated-interfaces-work-as-expected-on-nextgen.yml
```diff
@@ -145,101 +145,6 @@ query: |
 variables:
   isLoyal: true
 serviceCalls:
-  current:
```
temaEmelyan, line 148: removed as the test only runs on the nextgen

test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
```diff
@@ -142,6 +142,8 @@ private suspend fun execute(
                 engineFactory.make(nadel, testHooks)
             }
             .dsl(fixture.overallSchema)
+            .overallWiringFactory(testHooks.wiringFactory)
+            .underlyingWiringFactory(testHooks.wiringFactory)
```
gnawf, line 146: Nice

gnawf (approved):
LGTM

PR #364, opened 2022-02-01
pass hydration info to NadelTransform isApplicable

temaEmelyan:

### DIFF
--- a/test/src/test/kotlin/graphql/nadel/tests/hooks/transformer-on-hydration-fields.kt
+++ b/test/src/test/kotlin/graphql/nadel/tests/hooks/transformer-on-hydration-fields.kt
@@ -2,6 +2,7 @@ package graphql.nadel.tests.hooks
 
 import graphql.language.StringValue
 import graphql.nadel.Service
+import graphql.nadel.ServiceExecutionHydrationDetails
 import graphql.nadel.ServiceExecutionResult
 import graphql.nadel.enginekt.NadelExecutionContext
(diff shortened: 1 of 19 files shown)


test/src/test/kotlin/graphql/nadel/tests/hooks/transformer-on-hydration-fields.kt
```diff
@@ -33,8 +43,18 @@ class `transformer-on-hydration-fields` : EngineTestHook {
...
                     service: Service,
                     overallField: ExecutableNormalizedField,
+                    hydrationDetails: ServiceExecutionHydrationDetails?,
                 ): Any? {
-                    return overallField.name.takeIf { it == "barById" }
+                    return if (overallField.name == "barById") {
+                        assert(hydrationDetails != null)
+                        overallField
+                    } else if (hasParentWithName(overallField, "barById")) {
+                        assert(hydrationDetails != null)
```
gnawf, line 52: Ohh, is this an indirect test for whether it's set correctly? Took me a bit to figure that out lol.

gnawf (approved):

test/src/test/kotlin/graphql/nadel/tests/hooks/transformer-on-hydration-fields.kt
temaEmelyan: yes it is!

PR #365, opened 2022-02-02
handle primitive JSON args

temaEmelyan:

- [n/a] Add tests that use __typename in queries
- [n/a] Does this change work with all nadel transformations (rename, type rename, hydration, etc)? Add tests for this.
- [n/a] Is it worth using hints for this change in order to be able to enable a percentage rollout?
- [y] Do we need to add integration tests for this change in the graphql gateway?
- [no] Do we need a pollinator check for this?


bbakerman (dismissed):
Cool

bbakerman (approved):

gnawf (approved):

PR #369, opened 2022-02-07
READY nadel hydration actor field required in overall schema

Nikola-3:


engine-nextgen/src/test/kotlin/graphql/nadel/validation/NadelHydrationValidationTest.kt
```diff
@@ -15,16 +15,19 @@ import io.kotest.core.spec.style.DescribeSpec
 private const val source = "$" + "source"
 private const val argument = "$" + "argument"
 
+//1. are we getting rid of the @hydrated argument 'service' now or in the separate PR?
```
temaEmelyan, line 18: remove this comment

engine-nextgen/src/main/kotlin/graphql/nadel/validation/NadelHydrationValidation.kt
```diff
@@ -57,24 +58,29 @@ internal class NadelHydrationValidation(
                 errors.add(MissingHydrationActorService(parent, overallField, hydration))
                 continue
             }
-
-            val actorServiceQueryType = actorService.underlyingSchema.queryType
-            val actorField = actorServiceQueryType.getFieldAt(hydration.pathToActorField)
-            if (actorField == null) {
-                errors.add(MissingHydrationActorField(parent, overallField, hydration, actorServiceQueryType))
-                continue
+            if (newHydrationValidation) {
```
temaEmelyan, line 61: you can simplify to something like
```kotlin
val schema = when {
  newHydrationValidation -> overallSchema
  else -> actorService.underlyingSchema
}

hydrationActorFieldExistsInSchema(hydration, errors, parent, overallField, actorService, isPolymorphicHydration, schema.queryType)
```

engine-nextgen/src/main/kotlin/graphql/nadel/validation/NadelSchemaValidationError.kt
```diff
@@ -195,7 +195,7 @@ sealed interface NadelSchemaValidationError {
             val of = makeFieldCoordinates(parentType.overall.name, overallField.name)
             val s = hydration.serviceName
             val af = "${actorServiceQueryType.name}.${hydration.pathToActorField.joinToString(separator = ".")}"
-            "Field $of tried to hydrate from non-existent field $af on service $s"
+            "Field $of tried to hydrate from non-existent field in overall schema $af on service $s"
```
temaEmelyan, line 198: that shouldn't be the default message since we are running new logic behind the feature flag

Nikola-3: Should I leave a TODO then? It feels like it can be confusing to see this message if you are used to fields only being required in underlying schema

engine-nextgen/src/main/kotlin/graphql/nadel/validation/NadelHydrationValidation.kt
Nikola-3: Do you find this clearer over deciding the query type in the when?

engine-nextgen/src/main/kotlin/graphql/nadel/validation/NadelSchemaValidationError.kt
temaEmelyan: we can either make the message conditional (if running with the FF on - overall, not - leave as is) or not worry about it at all/leave TODO

engine-nextgen/src/test/kotlin/graphql/nadel/validation/NadelHydrationValidationTest.kt
```diff
@@ -124,6 +127,58 @@ class NadelHydrationValidationTest : DescribeSpec({
...
+                            id: ID!
+                            name: String!
+                        }
+                    """.trimIndent(),
+                ),
+            )
+
+            val errors = validate(fixture)
+
+            assert(errors.map { it.message }.isNotEmpty())
```
felipe-gdr: Can we have a more specific assert here? Checking for more specific error messages makes the test more accurate and easier to understand what it is actually testing.

temaEmelyan: wait I swear I just pushed a commit with that

temaEmelyan: @felipe-gdr I think you are looking at the old diff, take a look here https://github.com/atlassian-labs/nadel/pull/369/files#diff-5737c48bda1f8571497df6122df9345befa787904d7fd2f66d3900d507d4a4f7R181

engine-nextgen/src/main/kotlin/graphql/nadel/validation/NadelHydrationValidation.kt
```diff
@@ -65,6 +66,23 @@ internal class NadelHydrationValidation(
                 continue
             }
 
+            val serviceRequiresActorFieldsToBeDeclaredInOverallSchema =
+                nadelValidationHints?.requireHydrationActorFieldInOverallSchemaHint?.getHintValue(actorService) ?: false
+            if (serviceRequiresActorFieldsToBeDeclaredInOverallSchema) {
+                val actorFieldFromOverall = overallSchema.queryType.getFieldAt(hydration.pathToActorField)
```
felipe-gdr, line 72: Just double checking: 
- I thought hydration instructions used the underlying schema as a reference (as in, the hydration query is sent directly to the underlying service, so it should be valid against that schema), so it didn't really matter if the actor field exists or not in the overall, is that still the case (or maybe it never was the cause and I'm dreaming 😬 ).

temaEmelyan: definitely has been the case so far but we are changing it now
we want actor fields in the overall schema so we can get `@CloudId` and `@ARI` from arguments for hydration queries. also we believe that after some additional refactoring we will be able to simplify the code when we don't need to be constantly switch between overall and underlying schemas

felipe-gdr: Awesome! Thanks for confirming ;-)

felipe-gdr (approved):

PR #372, opened 2022-02-09
READY handle NPE in GraphQLUtil.kt

Nikola-3:
There were NPE's found in splunk on this line, since type is never null I assume field was the cause. I will raise a follow up ticket as not all calling contexts handle the null value properly which could lead to more NPE's in the future


gnawf (dismissed):
Nice

felipe-gdr:
A quick test reproducing the NPE before the fix would be nice.

felipe-gdr (approved):

gnawf (approved):

PR #374, opened 2022-02-10
add validation for missing hydration arg

faizan-siddiqui:


engine-nextgen/src/main/kotlin/graphql/nadel/validation/NadelHydrationValidation.kt
```diff
@@ -157,7 +157,23 @@ internal class NadelHydrationValidation(
             }
         }
 
-        return duplicatedArgumentsErrors + remoteArgErrors
+        val missingActorArgErrors = actorField.arguments.mapNotNull { actorArg ->
+            val hydrationArg = hydration.arguments.find { it.name.equals(actorArg.name) }
```
gnawf, line 161: Nit: use `==` in Kotlin instead of `.equals`

In Kotlin `==` invokes `.equals` and `===` does the identity check like `==` does in Java.

engine-nextgen/src/main/kotlin/graphql/nadel/validation/NadelHydrationValidation.kt
```diff
@@ -157,7 +157,23 @@ internal class NadelHydrationValidation(
...
+            if (hydrationArg == null) {
+                //this is allowed if the actor field arg is nullable so check for that
```
gnawf, line 163: If you're looking at a functional way to do it, you should add a filter on `arguments` and filter only the arguments with non-nullable types

gnawf (commented):
Looks good

engine-nextgen/src/test/kotlin/graphql/nadel/validation/NadelHydrationValidationTest.kt
```diff
@@ -185,6 +185,63 @@ class NadelHydrationValidationTest : DescribeSpec({
...
+            val errors = validate(fixture)
+
+            assert(errors.size == 1)
+            val error = errors.assertSingleOfType<MissingHydrationActorFieldInOverallSchema>()
+            assert(error.service.name == "issues")
+            assert(error.hydration.serviceName == "users")
+            assert(error.overallField.name == "creator")
+            assert(error.parentType.overall.name == "Issue")
+        }
+
```
gnawf, line 244: Where'd all the tests go? You've duplicated the test above here?

gnawf (changes requested):
Can you add the tests back? Looks like they were deleted in 7ff20e9ea7be8ba7269d1ef6f2a137596e02f54f ?

engine-nextgen/src/test/kotlin/graphql/nadel/validation/NadelHydrationValidationTest.kt
faizan-siddiqui: yikes yea not sure what happened there, adding them back

engine-nextgen/src/test/kotlin/graphql/nadel/validation/NadelHydrationValidationTest.kt
```diff
@@ -724,6 +724,125 @@ class NadelHydrationValidationTest : DescribeSpec({
...
+                            name: String!
+                        }
+                    """.trimIndent(),
+                    ),
+            )
+
+            val errors = validate(fixture)
+            assert(errors.map { it.message }.isNotEmpty())
+
+            val error = errors.assertSingleOfType<MissingHydrationActorFieldArgument>()
```
gnawf, line 782: Just occurred to me, you need to create a new class for this.

I think we rename MissingHydrationActorFieldArgument → NonExistentHydrationActorFieldArgument
And create a new MissingHydrationActorFieldArgument

Because the current impl will generate the error message

```
Hydration on field $of references non-existent argument $argument on hydration actor $s.$af
```

engine-nextgen/src/main/kotlin/graphql/nadel/validation/NadelSchemaValidationError.kt
```diff
@@ -317,6 +317,25 @@ sealed interface NadelSchemaValidationError {
...
+            val actorServiceQueryType: GraphQLObjectType,
+            val argument: String,
+    ) : NadelSchemaValidationError {
+        val service: Service get() = parentType.service
+
+        override val message = run {
+            val of = makeFieldCoordinates(parentType.overall.name, overallField.name)
+            val s = hydration.serviceName
+            val af = "${actorServiceQueryType.name}.${hydration.pathToActorField.joinToString(separator = ".")}"
+            "Hydration on field $of is missing argument $argument on hydration actor $s.$af"
```
gnawf, line 333: This is the other way around? The class names should be flipped.

Also:

```suggestion
            "Hydration on field $of is missing the required argument $argument on hydration actor $s.$af"
```

gnawf (changes requested):

gnawf (approved):

PR #381, opened 2022-02-21
add overall hydration actor field to NadelBatchHydrationFieldInstruction

temaEmelyan:


engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/NadelFieldInstruction.kt
```diff
@@ -103,6 +103,7 @@ data class NadelBatchHydrationFieldInstruction(
     override val sourceFields: List<NadelQueryPath>,
     val batchSize: Int,
     val batchHydrationMatchStrategy: NadelBatchHydrationMatchStrategy,
+    val overallActorFieldDef: GraphQLFieldDefinition?,
```
gnawf, line 106: Might be worth deprecating `actorFieldDef` in case we don't get around to this work soon.

Also, add a note that this should be not-null in the future?

gnawf (approved):

gnawf (approved):

PR #383, opened 2022-02-25
add ENF to service execution parameters

temaEmelyan:


engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/NadelFieldInstruction.kt
```diff
@@ -76,6 +77,12 @@ interface NadelGenericHydrationInstruction {
      * or [NadelBatchHydrationMatchStrategy.MatchObjectIdentifier.sourceId].
      */
     val sourceFields: List<NadelQueryPath>
+
+    /**
+     * The field definition in the overall schema referenced by [queryPathToActorField].
+     * should be non-null in the future
+     */
+    val overallActorFieldDef: GraphQLFieldDefinition?
```
temaEmelyan, line 85: previously I added this to the batch instruction only, now adding it to `NadelGenericHydrationInstruction`

bbakerman (approved):

PR #384, opened 2022-02-27
Improve isConditional for ENFs

gnawf:

- [x] Is it worth using hints for this change in order to be able to enable a percentage rollout? No
- [x] Do we need to add integration tests for this change in the graphql gateway? No
- [x] Do we need a pollinator check for this? No

Waiting on https://github.com/graphql-java/graphql-java/pull/2638

### DIFF
--- a/test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
+++ b/test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
@@ -44,6 +44,7 @@ private val singleTestToRun = (System.getenv("TEST_NAME") ?: "")
  * NadelExecutionInput.Builder#transformExecutionHints method in your test hook class.
  */
 private val defaultHints = NadelExecutionHints.newHints()
+    .newDocumentCompiler { true }
     .build()
 
 private val sep = "-".repeat(50)
--- a/test/src/test/resources/fixtures/hydration/hydration-with-interfaces-asking-typename.yml
+++ b/test/src/test/resources/fixtures/hydration/hydration-with-interfaces-asking-typename.yml
@@ -1,0 +1,118 @@
+name: hydration with interfaces asking typename
+enabled:
+  current: false
+  nextgen: true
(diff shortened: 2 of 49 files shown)


gnawf:
This should speed up the Polaris queries by a decent bit. Only metrics will tell though.

test/src/test/resources/fixtures/hydration/hydration-with-interfaces-asking-typename.yml
```diff
@@ -0,0 +1,118 @@
...
+      nodes: [Node]
+    }
+    type JiraIssue implements Node @renamed(from: "Issue") {
+      id: ID! 
+    }
+    interface Node {
+      id: ID! 
+    }
+    type User implements Node {
+      id: ID! @hydrated(service: "Issues" field: "ariById" arguments: [{name: "id" value: "$source.id"}])
```
gnawf, line 17: I added a bunch of tests that ask for fields from interfaces where some impls of the interface have synthetic definitions i.e. hydrations, renames, deep renames.

This is in line with what Townsquare does with their schema i.e. `id: ID @renamed(from: "ari")` in some places where `id` is specified by `Node`.

bbakerman (approved):

test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
```diff
@@ -44,6 +44,7 @@ private val singleTestToRun = (System.getenv("TEST_NAME") ?: "")
  * NadelExecutionInput.Builder#transformExecutionHints method in your test hook class.
  */
 private val defaultHints = NadelExecutionHints.newHints()
+    .newDocumentCompiler { true }
```
gnawf, line 47: Tests have already been updated to new compiler, so I'm defaulting this to true. One thing to note, when we eventually remove the variables FF, we'll need to update all our tests.

bbakerman (approved):

PR #385, opened 2022-03-01
Improve is conditional 2

gnawf:
Part 2 for changes in https://github.com/graphql-java/graphql-java/pull/2738

### DIFF
--- a/api/src/main/java/graphql/nadel/compat/ExecutableNormalizedOperationToAstCompiler.java
+++ b/api/src/main/java/graphql/nadel/compat/ExecutableNormalizedOperationToAstCompiler.java
@@ -22,6 +22,7 @@ import graphql.normalized.NormalizedInputValue;
 import graphql.normalized.VariableAccumulator;
 import graphql.normalized.VariablePredicate;
 import graphql.normalized.VariableValueWithDefinition;
+import graphql.schema.GraphQLObjectType;
 import graphql.schema.GraphQLSchema;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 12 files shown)


test/src/test/resources/fixtures/hydration/polymorphic hydrations/solitary-polymorphic-hydration-when-hook-returns-null.yml
```diff
@@ -134,7 +134,6 @@ serviceCalls:
         query: |
           query {
             humanById(id: "HUMAN-0") {
-              __typename__type_filter____typename: __typename
```
gnawf, line 137: Just note here, this is deleted because the ENF now merges the `__typename` field correctly. So the field is no longer completely removed and thus triggering the typename filter transform.

Can explain more in call, too lazy to type out full details.

test/src/test/resources/fixtures/polymorphism/query-with-pass-through-interfaces-and-unions-that-dont-have-typename-in-them-work-as-expected.yml
```diff
@@ -1,5 +1,4 @@
-name: query with pass through interfaces and unions that DONT have __typename in them
-  work as expected
+name: query with pass through interfaces and unions that DONT have __typename in them work as expected
```
gnawf, line 1: Couldn't find these tests by project search because the test name was split across lines.

api/src/main/java/graphql/nadel/compat/ExecutableNormalizedOperationToAstCompiler.java
```diff
@@ -180,4 +182,15 @@ private static SelectionSet selectionSet(List<Field> fields) {
             return argValue(executableNormalizedField, argName, normalizedInputValue.getValue(), variableAccumulator);
         }
     }
+
+    private static boolean isConditionalCompat(ExecutableNormalizedField nf, GraphQLSchema schema) {
```
gnawf, line 186: New GraphQL Java version removed the old `isConditional` behavior. So we're adding it back here for the FF.

bbakerman (approved):

PR #388, opened 2022-03-07
Delete more code

gnawf:


api/src/main/java/graphql/nadel/hooks/CreateServiceContextParams.java
```diff
@@ -1,79 +1,32 @@
...
-import graphql.nadel.engine.NadelContext;
-import graphql.schema.GraphQLSchema;
 
 @PublicApi
 public class CreateServiceContextParams {
     private final Service service;
-    private final ExecutionStepInfo executionStepInfo;
-    private final ExecutionId executionId;
-    private final GraphQLSchema schema;
-    private final NadelContext context;
```
gnawf, line 18: Not used, the new engine only supplies service. I've also checked with AGG and it seems fine.

api/src/main/java/graphql/nadel/Nadel.java
```diff
@@ -77,7 +75,6 @@
     final ServiceExecutionHooks serviceExecutionHooks;
     final PreparsedDocumentProvider preparsedDocumentProvider;
     final ExecutionIdProvider executionIdProvider;
-    final IntrospectionRunner introspectionRunner;
```
gnawf, line 80: New engine gets supplied one directly to `NextgenEngine` constructor. At some point we'll create a tighter coupling between the two classes/modules again so that avoid the need to directly construct `NextgenEngine`.

api/src/main/java/graphql/nadel/instrumentation/parameters/NadelInstrumentationExecuteOperationParameters.java
```diff
@@ -30,42 +25,16 @@
     private final Object context;
 
     public NadelInstrumentationExecuteOperationParameters(
-            NormalizedQueryFromAst normalizedQueryFromAst,
```
gnawf, line 33: Removes reference to old normalized query

api/src/main/java/graphql/nadel/schema/UnderlyingWiringFactory.java
```diff
@@ -39,38 +32,17 @@ public boolean providesTypeResolver(InterfaceWiringEnvironment environment) {
 
     @Override
     public TypeResolver getTypeResolver(InterfaceWiringEnvironment environment) {
-        return underScoreTypeNameResolver();
+        return delegateWiringFactory.getTypeResolver(environment);
```
gnawf, line 35: Tbh idk what this class really does. I think we can just delete it? After removing this code it seems to just delegate everything to `delegateWiringFactory`…

api/src/test/groovy/graphql/nadel/testutils/TestUtil.groovy
```diff
@@ -307,11 +300,6 @@ class TestUtil {
         new NadelGraphQLParser().parseDocument(query)
     }
 
-    static NormalizedQueryFromAst createNormalizedQuery(GraphQLSchema schema, Document document) {
```
gnawf, line 310: Unused test code that references deleted classes

temaEmelyan (approved):

PR #389, opened 2022-03-07
make hydration instruction overall actor field non null

temaEmelyan:

### DIFF
--- a/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/NadelExecutionBlueprintFactory.kt
+++ b/engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/NadelExecutionBlueprintFactory.kt
@@ -50,7 +50,7 @@ import graphql.schema.GraphQLType
 internal object NadelExecutionBlueprintFactory {
     fun create(
         engineSchema: GraphQLSchema,
-        services: List<Service>
+        services: List<Service>,
     ): NadelOverallExecutionBlueprint {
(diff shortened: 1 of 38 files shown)


engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/NadelExecutionBlueprintFactory.kt
```diff
@@ -50,7 +50,7 @@ import graphql.schema.GraphQLType
 internal object NadelExecutionBlueprintFactory {
     fun create(
         engineSchema: GraphQLSchema,
-        services: List<Service>
+        services: List<Service>,
```
temaEmelyan, line 53: IDE's auto format

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/NadelFieldInstruction.kt
```diff
@@ -80,9 +80,8 @@ interface NadelGenericHydrationInstruction {
 
     /**
      * The field definition in the overall schema referenced by [queryPathToActorField].
-     * should be non-null in the future
      */
-    val overallActorFieldDef: GraphQLFieldDefinition?
+    val overallActorFieldDef: GraphQLFieldDefinition
```
gnawf, line 84: Great

gnawf (approved):
Very nice, great to see all the tests fixed too 😄.

PR #390, opened 2022-03-07
remove validation hint for overall hydration actor fields

temaEmelyan:


gnawf (approved):
We might want to keep the hints object around even if it's empty for future use cases. But nbd. It doesn't take much effort to add it back in.

PR #391, opened 2022-03-07
Input type validation

faizan-siddiqui:
Add input type compatibility check, please confirm if my understanding of nullability compatibility is correct


engine-nextgen/src/main/kotlin/graphql/nadel/validation/util/NadelInputValidationUtil.kt
```diff
@@ -0,0 +1,85 @@
...
+import graphql.nadel.validation.util.NadelSchemaUtil.getUnderlyingName
+import graphql.schema.GraphQLEnumType
+import graphql.schema.GraphQLInputObjectType
+import graphql.schema.GraphQLInputType
+import graphql.schema.GraphQLNamedType
+import graphql.schema.GraphQLNonNull
+import graphql.schema.GraphQLScalarType
+import graphql.schema.GraphQLTypeUtil
+
+object NadelInputValidationUtil {
```
gnawf, line 12: Can you copy the code in `NadelTypeValidation.isOutputTypeValid` and flip the logic instead?

engine-nextgen/src/main/kotlin/graphql/nadel/validation/util/NadelInputValidationUtil.kt
```diff
@@ -0,0 +1,85 @@
...
+
+    fun isNonNullType(graphQLInputType: GraphQLInputType?): Boolean {
+        return graphQLInputType is GraphQLNonNull
+    }
+
+    /**
+     * Takes in a string and removes the ! if it exists at the end
+     */
+    fun makeArgTypeNullable(underlyingArgType: String): String {
+        return if (underlyingArgType.endsWith("!")) {
```
gnawf, line 79: I would say you should always avoid messing with strings if you can because Strings are fragile. You should use the type system i.e. `GraphQLNonNull` `GraphQLList` and `GraphQLUtil.unwrap` etc

engine-nextgen/src/test/kotlin/graphql/nadel/validation/NadelFieldValidationTest.kt
```diff
@@ -62,6 +64,86 @@ class NadelFieldValidationTest : DescribeSpec({
             assert(error.argument.name == "world")
         }
 
+        it("passes if overall argument value is more strict") {
+            val fixture = NadelValidationTestFixture(
+                    overallSchema = mapOf(
+                            "test" to """
+                        type Query {
+                            echo(world: Boolean!): String
```
gnawf, line 72: Add a test for `world: [Boolean]` in overall and `World: [Boolean!]` in underlying

It should fail but I don't think it will with the current impl

engine-nextgen/src/main/kotlin/graphql/nadel/validation/NadelFieldValidation.kt
```diff
@@ -99,14 +102,23 @@ internal class NadelFieldValidation(
...
-                    )
+                val unwrappedTypeIssues = typeValidation.validate(
+                        NadelServiceSchemaElement(
+                                service = parent.service,
+                                overall = overallArg.type.unwrapAll(),
+                                underlying = underlyingArg.type.unwrapAll(),
+                        )
                 )
+
+                val inputTypeIssues = if (!isMatchingInputTypeNameIgnoringNullableRenameScalar(overallArg.type, underlyingArg.type)) {
```
gnawf, line 113: This should be `val inputType = typeValidation.validateInputType` for code consistency. The caller shouldn't have to generate the error.

engine-nextgen/src/main/kotlin/graphql/nadel/validation/NadelSchemaValidationError.kt
```diff
@@ -151,6 +151,44 @@ sealed interface NadelSchemaValidationError {
...
+    ) : NadelSchemaValidationError {
+        val service: Service get() = parentType.service
+
+        override val message = run {
+            val s = service.name
+            val ofa = makeFieldCoordinates(parentType.overall.name, overallInputArg.name)
+            val ufa = makeFieldCoordinates(parentType.underlying.name, underlyingInputArg.name)
+            val ot = GraphQLTypeUtil.simplePrint(overallInputArg.type)
+            val ut = GraphQLTypeUtil.simplePrint(underlyingInputArg.type)
+            "Overall field argument $ofa has input type $ot but underlying field argument $ufa in service $s has input type $ut"
```
gnawf, line 186: Does this message make sense?

For the argument

```
type Foo {
  baz(arg: String)
}
```

Doesn't it print out

> Overall field argument Foo.arg has input type String

It should be

> Overall field Foo.baz has argument arg of input type String …

gnawf (dismissed):
Can you refactor `NadelInputValidationUtil` to follow the approach in `NadelTypeValidation.isOutputTypeValid`. Couple other minor comments. Good otherwise

engine-nextgen/src/test/kotlin/graphql/nadel/validation/NadelInputValidationTest.kt
```diff
@@ -222,5 +261,121 @@ class NadelInputValidationTest : DescribeSpec({
...
+            val error = errors.assertSingleOfType<IncompatibleFieldInputType>()
+            assert(error.service.name == "test")
+            assert(error.parentType.overall.name == "Role")
+            assert(error.parentType.underlying.name == error.parentType.overall.name)
+            assert(error.overallInputField.type.unwrapAll().name == "Boss")
+            assert(error.underlyingInputField.type.unwrapAll().name == "Manager")
+            assert(error.subject == error.overallInputField)
+        }
+
+        it("passes if all input values exist in underlying type but overall input is less strict") {
```
faizan-siddiqui, line 312: this should have been "fails if..." will fix

build.gradle
```diff
@@ -1,7 +1,7 @@
 import java.text.SimpleDateFormat
 
 plugins {
-    id "org.jetbrains.kotlin.jvm" version "1.5.31"
+    id "org.jetbrains.kotlin.jvm" version "1.6.10"
```
gnawf, line 4: Had to upgrade Kotest because it didn't seem to want to run the data driven tests in Gradle. The new Kotest required Kotlin `1.6.x` or beyond.

temaEmelyan (approved):

faizan-siddiqui (commented):
addressed requested changes

PR #392, opened 2022-03-07
remove exemption for scalar output type mismatch

faizan-siddiqui:
Removing this exemption, we will then filter through any errors for this in the gateway and act accordingly


gnawf:
Do we have an equivalent PR in the AGG for grandfathering the existing violations? Otherwise, if we merged this, it would end up blocking Nadel updates in AGG.

faizan-siddiqui:
> Do we have an equivalent PR in the AGG for grandfathering the existing violations? Otherwise, if we merged this, it would end up blocking Nadel updates in AGG.

We do now

gnawf (approved):

PR #400, opened 2022-03-18
remove underlying actor field from hydration instruction

temaEmelyan:


test/src/test/resources/fixtures/scalars/indexed-hydrating-using-json-data-as-arg.yml
```diff
@@ -18,7 +18,7 @@ overallSchema:
     }
   Baz: |
     type Query {
-      baz(data: JSON!): Foo @hidden
+      baz(data: [JSON!]!): [Foo] @hidden
```
temaEmelyan, line 21: I think this test had a mistake in the schema definitions, the underlying schema is. Now this test fails because we use the overall field. In practice this schema wouldn't pass semantic validation
```
  Baz: |
    type Query {
      baz(data: [JSON!]!): [Foo]
    }
    type Foo {
      id: ID!
    }
    scalar JSON
```

gnawf: Yea nice catch

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/NadelFieldInstruction.kt
```diff
@@ -104,7 +97,6 @@ data class NadelBatchHydrationFieldInstruction(
     override val hydratedFieldDef: GraphQLFieldDefinition,
     override val actorService: Service,
     override val queryPathToActorField: NadelQueryPath,
-    override val actorFieldDef: GraphQLFieldDefinition,
```
gnawf, line 107: Can we rename `overallActorFieldDef` → `actorFieldDef`?

I think it might be a little confusing, but longer term it'll be redundant to explicitly state `overall`.

gnawf (changes requested):

engine-nextgen/src/main/kotlin/graphql/nadel/enginekt/blueprint/NadelFieldInstruction.kt
temaEmelyan: ✅

gnawf (dismissed):

gnawf (approved):

PR #401, opened 2022-03-20
Run validation on tests and fix hydration output validation

gnawf:


.github/workflows/pull_request.yml
```diff
@@ -18,3 +18,10 @@ jobs:
           java-version: '11.0.9'
       - name: build and test
         run: ./gradlew assemble --info && ./gradlew check --info
+      - uses: actions/upload-artifact@v3
```
gnawf, line 21: Uploads the test report if it failed

engine-nextgen/src/main/kotlin/graphql/nadel/validation/NadelHydrationValidation.kt
```diff
@@ -91,40 +74,44 @@ internal class NadelHydrationValidation(
...
+            is GraphQLUnionType -> actorOutputType.types
+            is GraphQLInterfaceType -> overallSchema.getImplementations(actorOutputType)
+            else -> listOf(actorOutputType as GraphQLNamedOutputType)
+        }
+
+        val typeValidation = actorOutputTypes
+            // Find incompatible output types
+            .filter { actorOutputType ->
+                acceptableOutputTypes.none { acceptableOutputType ->
+                    typeValidation.isAssignableTo(lhs = acceptableOutputType, rhs = actorOutputType)
```
gnawf, line 104: Made some changes to the way the output type is validated. Basically we can validate stuff like

```graphql
type Query {
  userById(id: ID!): User
}
type Issue {
  assignee: AbstractUser @hydrated(field: "userById")
}
interface AbstractUser {
}
type User implements AbstractUser {
}
```

i.e. type doesn't exactly match up, but we do a smarter check on whether they're compatible.

engine-nextgen/src/main/kotlin/graphql/nadel/validation/NadelHydrationValidation.kt
```diff
@@ -58,29 +56,14 @@ internal class NadelHydrationValidation(
...
-            val actorServiceQueryType = actorService.underlyingSchema.queryType
-            val actorField = actorServiceQueryType.getFieldAt(hydration.pathToActorField)
+            val actorField = overallSchema.queryType.getFieldAt(hydration.pathToActorField)
             if (actorField == null) {
-                errors.add(MissingHydrationActorField(parent, overallField, hydration, actorServiceQueryType))
+                errors.add(MissingHydrationActorField(parent, overallField, hydration))
                 continue
             }
 
-            val overallActorField = overallSchema.queryType.getFieldAt(hydration.pathToActorField)
```
gnawf, line 68: Removed this as we simply grab from the overall schema now.

engine-nextgen/src/main/kotlin/graphql/nadel/validation/NadelHydrationValidation.kt
```diff
@@ -91,40 +74,44 @@ internal class NadelHydrationValidation(
...
-                            parent,
-                            overallField
-                        )
-                    )
-                overallType = overallSchema.getType(overallTypeName) as GraphQLUnmodifiedType
-            } else {
-                return listOf(FieldWithPolymorphicHydrationMustReturnAUnion(parent, overallField))
             }
-        }
-        val typeValidation = typeValidation.validate(
```
gnawf, line 121: No need to run type validation here as we no longer deal with two type where one is underlying and one is overall.

The validation for the types is already done elsewhere. So in the code here, if they're assignable to each other, then the types are valid.

engine-nextgen/src/main/kotlin/graphql/nadel/validation/NadelSchemaValidationError.kt
```diff
@@ -59,6 +61,21 @@ sealed interface NadelSchemaValidationError {
             .build()
     }
 
+    data class UnionHasExtraType(
```
gnawf, line 64: Not actually used right now.

engine-nextgen/src/main/kotlin/graphql/nadel/validation/NadelSchemaValidationError.kt
```diff
@@ -226,32 +243,14 @@ sealed interface NadelSchemaValidationError {
...
             val of = makeFieldCoordinates(parentType.overall.name, overallField.name)
             val s = hydration.serviceName
-            val af = "${actorServiceQueryType.name}.${hydration.pathToActorField.joinToString(separator = ".")}"
-            "Field $of tried to hydrate from non-existent field $af on service $s"
-        }
-
-        override val subject = overallField
-    }
-
-    data class MissingHydrationActorFieldInOverallSchema(
```
gnawf, line 243: I don't think we actually reference these explicitly, so I don't think removing the type matters.

engine-nextgen/src/main/kotlin/graphql/nadel/validation/NadelTypeValidation.kt
```diff
@@ -98,6 +106,31 @@ internal class NadelTypeValidation(
...
+     * ```
+     *
+     * Note: this assumes both types are from the same schema. This does NOT
+     * deal with differences between overall and underlying schema.
+     */
+    fun isAssignableTo(lhs: GraphQLNamedOutputType, rhs: GraphQLNamedOutputType): Boolean {
+        if (lhs.name == rhs.name) {
+            return true
+        }
+        if (lhs.name == GraphQLID.name && rhs.name == GraphQLString.name) {
```
gnawf, line 125: I'm going to leave this in… I think the other type validation should follow this in the future? Thoughts?

cc: @felipe-gdr

engine-nextgen/src/test/kotlin/graphql/nadel/validation/NadelHydrationValidationTest.kt
```diff
@@ -914,11 +914,295 @@ class NadelHydrationValidationTest : DescribeSpec({
...
+            val errors = validate(fixture)
+            assert(errors.isNotEmpty())
+            val error = errors.singleOfType<HydrationIncompatibleOutputType>()
+            assert(error.parentType.overall.name == "Issue")
+            assert(error.actorField.name == "externalUser")
+            assert((error.actorField.type as GraphQLNamedType).name == "ExternalUser")
+            assert(error.incompatibleOutputType.name == "ExternalUser")
+        }
+
+        it("fails if actor output type does not implement interface") {
```
gnawf, line 995: Some tests against the is hydration type assignable code I added.

test/src/test/kotlin/graphql/nadel/tests/EngineTestHook.kt
```diff
@@ -43,6 +44,10 @@ interface EngineTestHook {
         return builder
     }
 
+    fun isSchemaValid(errors: Set<NadelSchemaValidationError>): Boolean {
```
gnawf, line 47: Hook to bypass validation e.g. to ensure we let the malformed JSW schema through.

See the test `let-jsw-do-jsw-things.yml`

test/src/test/resources/fixtures/basic/schema-transformation-is-applied.yml
```diff
@@ -22,13 +22,18 @@ underlyingSchema:
       hello: String
     }
 
+    type Subscription {
```
gnawf, line 25: All changes beyond here are just fixes to the fixtures to make them valid.

engine-nextgen/src/main/kotlin/graphql/nadel/validation/NadelUnionValidation.kt
```diff
@@ -0,0 +1,45 @@
...
+
+internal class NadelUnionValidation(
+    private val typeValidation: NadelTypeValidation,
+) {
+    fun validate(
+        schemaElement: NadelServiceSchemaElement,
+    ): List<NadelSchemaValidationError> {
+        return emptyList()
+
+        // return if (schemaElement.overall is GraphQLUnionType && schemaElement.underlying is GraphQLUnionType) {
```
gnawf, line 11: Couldn't quite wrap my head around the logic here. Will revisit later.

test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
```diff
@@ -309,3 +324,59 @@ private suspend fun execute(
...
+    val nadel = Nadel.newNadel()
+        .engineFactory {
+            object : NadelExecutionEngine {
+                override fun execute(
+                    executionInput: ExecutionInput,
+                    queryDocument: Document,
+                    instrumentationState: InstrumentationState?,
+                    nadelExecutionParams: NadelExecutionParams,
+                ): CompletableFuture<ExecutionResult> {
+                    throw UnsupportedOperationException("no-op")
```
gnawf, line 341: I mentioned this to @felipe-gdr but in the future I think we should split the creation of the "engine schema" and "services" from the `Nadel` object e.g.

1. Introduce new `NadelSchemaBuilder`
2. Create engine schema and services using `NadelSchemaBuilder` say the output is some `NadelSchemas` object
3. Make Nadel.Builder take in `NadelSchemas` _or_ optionally work as today, but delegate off to `NadelSchemaBuilder`

The issue today is creating a Nadel object is very closely tied to creating a `NextgenEngine` which then creates a blueprint, and the blueprint assumes a valid schema. So for validation, to get the engine schema and services out of the Nadel object we need to stub the engine so no blueprint is instantiated.

gnawf, line 341: I mentioned this to @felipe-gdr but in the future I think we should split the creation of the "engine schema" and "services" from the `Nadel` object e.g.

1. Introduce new `NadelSchemaBuilder`
2. Create engine schema and services using `NadelSchemaBuilder` say the output is some `NadelSchemas` object
3. Make Nadel.Builder take in `NadelSchemas` _or_ optionally work as today, but delegate off to `NadelSchemaBuilder`

The issue today is creating a Nadel object is very closely tied to creating a `NextgenEngine` which then creates a blueprint, and the blueprint assumes a valid schema. So to get the engine schema and services out of the Nadel object we need to stub the engine so no blueprint is instantiated.

engine-nextgen/src/main/kotlin/graphql/nadel/validation/NadelTypeValidation.kt
faizan-siddiqui: so this allows String -> ID, and one way only?

gnawf: Yep. I think the other way around is technically possible too. They're both Strings, but as an input type ID accepts Ints too.

test/src/test/resources/fixtures/renames/types/renamed-type-is-shared-but-not-renamed-in-one-service.yml
```diff
@@ -1,5 +1,6 @@
 name: renamed type is shared but not renamed in one service
-enabled: true
+# Not supported
```
temaEmelyan, line 2: what's the plan with these? are we going to delete them

engine-nextgen/src/main/kotlin/graphql/nadel/validation/NadelUnionValidation.kt
temaEmelyan: We have this same code in the gateway, right? Are we leaving a gap in our validation by commenting this out?

test/src/test/resources/fixtures/renames/types/renamed-type-is-shared-but-not-renamed-in-one-service.yml
gnawf: I think we'll want to support it eventually so I'm keeping the test in. This is the shared rename issue where we'll want the repeatable

```
```

syntax.

engine-nextgen/src/main/kotlin/graphql/nadel/validation/NadelUnionValidation.kt
gnawf: I don't see any validation on it, so don't think it's a big deal.

temaEmelyan (approved):

PR #403, opened 2022-03-21
Merge api and engine-nextgen

gnawf:
This means we're back to one module being published. Just `nadel` which is now found under the `lib/` directory.

Pending #401


lib/src/main/java/graphql/nadel/Nadel.kt
```diff
@@ -43,18 +43,17 @@ import graphql.schema.idl.ScalarInfo.GRAPHQL_SPECIFICATION_SCALARS as graphQLSpe
...
-    val underlyingWiringFactory: WiringFactory,
-    val schemaTransformationHook: SchemaTransformationHook,
+    internal val serviceExecutionFactory: ServiceExecutionFactory,
+    internal val instrumentation: NadelInstrumentation,
+    internal val serviceExecutionHooks: ServiceExecutionHooks,
+    internal val preparsedDocumentProvider: PreparsedDocumentProvider,
+    internal val executionIdProvider: ExecutionIdProvider,
+    internal val overallWiringFactory: WiringFactory,
+    internal val underlyingWiringFactory: WiringFactory,
+    internal val schemaTransformationHook: SchemaTransformationHook,
```
gnawf, line 56: We get benefits like this. Since it's the same module now, we can further restrict visibility for members shared between `api/` and `engine-nextgen/`

lib/src/main/java/graphql/nadel/NextgenEngine.kt
```diff
@@ -7,32 +7,32 @@ import graphql.ExecutionResultImpl.newExecutionResult
 import graphql.GraphQLError
 import graphql.execution.instrumentation.InstrumentationState
 import graphql.language.Document
-import graphql.nadel.enginekt.NadelExecutionContext
```
gnawf, line 10: Removed `kt` from the package name.

lib/src/main/java/graphql/nadel/engine/blueprint/NadelExecutionBlueprintFactory.kt
```diff
@@ -316,9 +316,7 @@ private class Factory(
             NadelBatchHydrationMatchStrategy.MatchIndex
         } else if (hydration.objectIdentifiers?.isNotEmpty() == true) {
             NadelBatchHydrationMatchStrategy.MatchObjectIdentifiers(
-                hydration.objectIdentifiers
-                // todo remove !! once they we merge api/ and nextgen-engine/
-                !!.map { objectIdentifier ->
+                hydration.objectIdentifiers.map { objectIdentifier ->
```
gnawf, line 319: Previously the Kotlin compiler couldn't _ensure_ that the value would be final as the code belonged to a different module e.g. module could be swapped out, different versions might not be final etc.

gnawf:
Also ran `./gradlew test` on this new setup and the same amount of tests are run.

lib/src/main/java/graphql/nadel/NextgenEngine.kt
temaEmelyan: we probably should rename `NextgenEngine` at some point

temaEmelyan (approved):

lib/src/main/java/graphql/nadel/NextgenEngine.kt
gnawf: Yep that's the next thing I'm working on.

So I'm basically rename it to `ExecutionEngine` or something, and basically remove the need to ever reference it. Gonna move stuff like

```kotlin
constructor(
    transforms: List<NadelTransform<out Any>> = emptyList(),
    introspectionRunnerFactory: NadelIntrospectionRunnerFactory = NadelIntrospectionRunnerFactory(::NadelDefaultIntrospectionRunner),
)
```

To `Nadel.Builder` and make `ExecutionEngine` as the default. But in order to do that, I need to separate the overall `GraphQLSchema` and `Service` creation from the `Nadel` object so that validation can work create those without invoking the blueprint, which requires a valid schema.

PR #406, opened 2022-03-23
Separate schema creation from Nadel object creation

gnawf:
Pending #403

### DIFF
--- a/lib/src/main/java/graphql/nadel/Nadel.kt
+++ b/lib/src/main/java/graphql/nadel/Nadel.kt
@@ -19,12 +19,9 @@ import graphql.nadel.instrumentation.parameters.NadelInstrumentationCreateStateP
 import graphql.nadel.instrumentation.parameters.NadelInstrumentationQueryExecutionParameters
 import graphql.nadel.instrumentation.parameters.NadelInstrumentationQueryValidationParameters
 import graphql.nadel.schema.NeverWiringFactory
-import graphql.nadel.schema.OverallSchemaGenerator
 import graphql.nadel.schema.QuerySchemaGenerator
 import graphql.nadel.schema.SchemaTransformationHook
-import graphql.nadel.schema.UnderlyingSchemaGenerator
 import graphql.nadel.util.LogKit
-import graphql.nadel.util.SchemaUtil
 import graphql.parser.InvalidSyntaxException
 import graphql.parser.Parser
 import graphql.schema.GraphQLSchema
@@ -34,12 +31,10 @@ import graphql.validation.Validator
 import org.slf4j.Logger
 import java.io.Reader
 import java.io.StringReader
-import java.util.Objects
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.CompletionException
 import java.util.concurrent.atomic.AtomicReference
 import java.util.function.Function
-import graphql.schema.idl.ScalarInfo.GRAPHQL_SPECIFICATION_SCALARS as graphQLSpecScalars
 
 class Nadel private constructor(
     private val engine: NadelExecutionEngine,
@@ -51,9 +46,6 @@ class Nadel private constructor(
     internal val serviceExecutionHooks: ServiceExecutionHooks,
     internal val preparsedDocumentProvider: PreparsedDocumentProvider,
     internal val executionIdProvider: ExecutionIdProvider,
-    internal val overallWiringFactory: WiringFactory,
-    internal val underlyingWiringFactory: WiringFactory,
-    internal val schemaTransformationHook: SchemaTransformationHook,
 ) {
     fun execute(nadelExecutionInput: NadelExecutionInput): CompletableFuture<ExecutionResult> {
         val executionInput: ExecutionInput = newExecutionInput()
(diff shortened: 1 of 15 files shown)


lib/src/main/java/graphql/nadel/Nadel.kt
```diff
@@ -237,24 +231,49 @@ class Nadel private constructor(
         private var schemaTransformationHook: SchemaTransformationHook = SchemaTransformationHook.Identity
         private var engineFactory: NadelExecutionEngineFactory = NadelExecutionEngineFactory(::NextgenEngine)
 
-        fun dsl(serviceName: String, nsdl: Reader): Builder {
-            Objects.requireNonNull(nsdl)
-            serviceNDSLs[serviceName] = nsdl
+        private var schemaBuilder = NadelSchemas.Builder()
```
gnawf, line 228: Could change this to supply the built `NadelSchemas` object in the future, but I don't think that's necessary right now.

lib/src/main/java/graphql/nadel/Nadel.kt
```diff
@@ -237,24 +231,49 @@ class Nadel private constructor(
...
+
+        fun overallSchema(serviceName: String, nsdl: Reader): Builder {
```
gnawf, line 230: Add `overallSchema` and `underlyingSchema`

Deprecated `dsl` etc.

lib/src/main/java/graphql/nadel/NadelSchemas.kt
```diff
@@ -0,0 +1,162 @@
...
+data class NadelSchemas internal constructor(
+    val engineSchema: GraphQLSchema,
+    val services: List<Service>,
+) {
+    companion object {
+        @JvmStatic
+        fun newBuilder() = Builder()
+    }
+
+    class Builder {
```
gnawf, line 24: People can now use this to create the overall/engine `GraphQLSchema` and the `Service`s.

This separates it from the `Nadel.Builder` so that validation can create this without creating a full Nadel object. The Nadel object is now tied to `NextgenEngine` and also blueprint creation which expects a valid schema. Hence the need to separate them.

lib/src/main/java/graphql/nadel/ServiceExecutionFactory.kt
```diff
@@ -13,15 +9,4 @@ interface ServiceExecutionFactory {
...
-    /**
-     * todo: why does this belong here and not in the Nadel constructor?
-     *
-     * Called to return underlying type definitions for a schema
-     *
-     * @param serviceName the name of the service
-     *
-     * @return a [graphql.schema.idl.TypeDefinitionRegistry] of all the underlying schema types
-     */
-    fun getUnderlyingTypeDefinitions(serviceName: String): TypeDefinitionRegistry
```
gnawf, line 26: Moved this to `Nadel.Builder.underlyingSchema` etc.

Putting this here would suggest that we dynamically get the schema e.g. via introspection. Which is an interesting idea, but for now we're _not_ doing it. So I'm moving it closer to the actual usage.

nadel/build.gradle.kts
```diff
@@ -1,9 +0,0 @@
-plugins {
```
gnawf, line 1: Forgot to delete this in #403

lib/src/main/java/graphql/nadel/Nadel.kt
```diff
@@ -226,75 +218,97 @@ class Nadel private constructor(
...
+            this.preparsedDocumentProvider = preparsedDocumentProvider
+            return this
+        }
+
+        fun executionIdProvider(executionIdProvider: ExecutionIdProvider): Builder {
+            this.executionIdProvider = executionIdProvider
+            return this
+        }
+
+        fun serviceExecutionHooks(serviceExecutionHooks: ServiceExecutionHooks): Builder {
```
gnawf, line 310: Just moved the members around so all the schema builder stuff is together. It's the same API wise.

lib/src/main/java/graphql/nadel/Nadel.kt
```diff
@@ -339,43 +353,31 @@ class Nadel private constructor(
...
                 executionIdProvider = executionIdProvider,
-                overallWiringFactory = overallWiringFactory,
-                underlyingWiringFactory = underlyingWiringFactory,
-                schemaTransformationHook = schemaTransformationHook,
             ).also {
                 nadel = it
             }
         }
 
-        private fun createServices(serviceExecutionFactory: ServiceExecutionFactory): List<Service> {
```
gnawf, line 350: Moved to `NadelSchemas.Builder`

felipe-gdr (dismissed):

lib/src/main/java/graphql/nadel/NadelSchemas.kt
```diff
@@ -0,0 +1,177 @@
...
+                    }
+                },
+            )
+        }
+
+        fun build(): NadelSchemas {
+            require(overallSchemas.isNotEmpty()) { "Nadel schemas must not be empty" }
+            require(underlyingSchemas.isNotEmpty()) { "Underlying schemas must not be empty" }
+
+            require(underlyingSchemas.keys.containsAll(overallSchemas.keys)) {
```
temaEmelyan, line 115: thinking about edge cases which will likely never happen :)
`containsAll` will return true when `underlyingSchemas.keys` is a superset of `overallSchemas.keys`, should it be `equals` instead?

gnawf: Yeah I deliberately wrote it like that, but now that I'm thinking about it again, yea maybe being defensive here is better… could show a mistake was made.

gnawf: @temaEmelyan done

temaEmelyan (approved):
great stuff!

PR #405, opened 2022-03-22
Start documentation

gnawf:
Starting with an overview of Nadel in Markdown, then will move into documenting the code itself.

PR #409, opened 2022-03-25
Update README with directives as old annotations were removed

ndejaco2:

Just updating the README here so no tests required.


ndejaco2:
Seems like this is already covered by https://github.com/atlassian-labs/nadel/pull/405 thanks :)

PR #413, opened 2022-04-04
Fix line numbers and integration issues with the Gateway service

gnawf:


lib/src/main/java/graphql/nadel/NadelSchemas.kt
```diff
@@ -30,10 +31,11 @@ data class NadelSchemas constructor(
         internal var serviceExecutionFactory: ServiceExecutionFactory? = null
 
         // .nadel files
-        internal var overallSchemas = mutableMapOf<String, Reader>()
+        internal var overallSchemaReaders = mutableMapOf<String, Reader>()
 
         // .graphqls files
-        internal var underlyingSchemas = mutableMapOf<String, Reader>()
+        internal var underlyingSchemaReaders = mutableMapOf<String, Reader>()
+        internal var underlyingTypeDefs = mutableMapOf<String, TypeDefinitionRegistry>()
```
gnawf, line 38: So we need to allow this because the Gateway parses the `Reader` by itself earlier and because it's a Reader we can't reuse it in two places. So we need to allow the Gateway to supply the parsed `TypeDefinitionRegistry`.

Maybe we should do the same for overall schemas. But not needed for now.

lib/src/main/java/graphql/nadel/NadelSchemas.kt
```diff
@@ -52,45 +54,69 @@ data class NadelSchemas constructor(
...
         fun overallSchema(serviceName: String, schema: String): Builder = also {
-            overallSchemas[serviceName] = schema.reader()
+            overallSchemaReaders[serviceName] = schema.reader()
         }
 
         fun underlyingSchema(serviceName: String, schema: String): Builder = also {
-            underlyingSchemas[serviceName] = schema.reader()
+            underlyingSchema(
+                serviceName,
+                MultiSourceReader.newMultiSourceReader()
```
gnawf, line 77: Need to use GraphQL Java's `MultiSourceReader` otherwise you don't get line numbers… TIL

lib/src/test/kotlin/graphql/nadel/NadelSchemasTest.kt
```diff
@@ -0,0 +1,256 @@
...
+        .asSequence()
+        .filterNot {
+            it.startsWith("__")
+        }
+        .filterNot {
+            it in allNadelBuiltInTypeNames
+        }
+        .toSet()
+
+class NadelSchemasTest : DescribeSpec({
```
gnawf, line 31: Added unit test for this

lib/src/test/kotlin/graphql/nadel/NadelSchemasTest.kt
```diff
@@ -0,0 +1,256 @@
...
+            assert(schemas.engineSchema.getField(worldCoordinates)?.definition?.sourceLocation?.line == 5)
+            assert(schemas.engineSchema.getField(issueIdCoordinates)?.definition?.sourceLocation?.line == 6)
+
+            val testService = schemas.services.single { it.name == "test" }
+            assert(testService.underlyingSchema.typeMap["Echo"]?.definition?.sourceLocation?.line == 5)
+            assert(testService.underlyingSchema.getField(echoCoordinates)?.definition?.sourceLocation?.line == 2)
+            assert(testService.underlyingSchema.getField(worldCoordinates)?.definition?.sourceLocation?.line == 6)
+
+            val issueService = schemas.services.single { it.name == "issue" }
+            assert(issueService.underlyingSchema.typeMap["Task"]?.definition?.sourceLocation?.line == 4)
```
gnawf, line 253: Tests for line numbers

bbakerman (approved):

PR #414, opened 2022-04-04
Add missing builder function for TypeDefs and fix checks

gnawf:
Should really check the PRs I _thought_ I finished last week.

### DIFF
--- a/lib/src/main/java/graphql/nadel/Nadel.kt
+++ b/lib/src/main/java/graphql/nadel/Nadel.kt
@@ -25,6 +25,7 @@ import graphql.nadel.util.LogKit
 import graphql.parser.InvalidSyntaxException
 import graphql.parser.Parser
 import graphql.schema.GraphQLSchema
+import graphql.schema.idl.TypeDefinitionRegistry
 import graphql.schema.idl.WiringFactory
 import graphql.validation.ValidationError
 import graphql.validation.Validator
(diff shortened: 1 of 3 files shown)


lib/src/main/java/graphql/nadel/Nadel.kt
```diff
@@ -271,6 +277,12 @@ class Nadel private constructor(
             return this
         }
 
+        @JvmName("underlyingTypeDefs")
+        fun underlyingSchemas(value: Map<String, TypeDefinitionRegistry>): Builder = also {
```
gnawf, line 281: Added this to `NadelSchemas.Builder` but forgot to add it here

lib/src/main/java/graphql/nadel/NadelSchemas.kt
```diff
@@ -136,14 +136,15 @@ data class NadelSchemas constructor(
 
         fun build(): NadelSchemas {
             require(overallSchemaReaders.isNotEmpty()) { "Nadel schemas must not be empty" }
-            require(underlyingSchemaReaders.isNotEmpty()) { "Underlying schemas must not be empty" }
+            require(underlyingSchemaReaders.isNotEmpty() || underlyingTypeDefs.isNotEmpty()) { "Underlying schemas must not be empty" }
```
gnawf, line 139: Fixes check if you only supply type defs and no readers. There's a new test for this below too.

gnawf (commented):
I tested this with the AGG and it builds fine now.

bbakerman (approved):

PR #417, opened 2022-04-11
Convert Spock tests to Kotlin and remove Groovy/Spock testing

gnawf:
Something was fucky with my tests earlier and they wouldn't run… Decided to kill off Groovy to make sure Kotest runs properly.


dondonz (approved):
Same weird interop problems happened with Central Schema scripts!

PR #423, opened 2022-05-04
Remove hydration specific execute path

gnawf:
So this code path used to exist because we didn't have all fields in the overall schema. Now we do, so we can pass _all_ the fields to the transforms.

We should remove the separate code path so there's less chance of bugs like today.

Not sure when we should merge. Maybe can't right now? Will double check tomorrow.


gnawf:
Man I swear I ran the tests before pushing, very clearly I didn't.

temaEmelyan (approved):

PR #430, opened 2022-05-20
Don't bypass field validation in extended operation types

felipe-gdr:
We have a couple of services in the central-schema that are using the `extend type Query` and/or `extend type Mutation` syntax.  Event though it is not necessary to extend operation types, this is something we still support.

Another possible fix is to write a validation that prevents people from using `extend type Query/Mutation` -> but this fix in Nadel was quick and simple, and allow for more flexibility in the schema definitions.

- [x ] Add tests that use __typename in queries
- [ x] Does this change work with all nadel transformations (rename, type rename, hydration, etc)? Add tests for this.
- [ x] Is it worth using hints for this change in order to be able to enable a percentage rollout?
- [ x] Do we need to add integration tests for this change in the graphql gateway?
- [ x] Do we need a pollinator check for this?

### DIFF
--- a/lib/src/test/kotlin/graphql/nadel/validation/NadelFieldValidationTest.kt
+++ b/lib/src/test/kotlin/graphql/nadel/validation/NadelFieldValidationTest.kt
@@ -558,5 +558,123 @@ class NadelFieldValidationTest : DescribeSpec({
 
             assert(errors.count { it is MissingUnderlyingField } == 3)
         }
+
+        it("passes if hydrated type is not present in underlying") {
+            val fixture = NadelValidationTestFixture(
+                overallSchema = mapOf(
+                    "test" to """
+                        type Query {
+                            fieldA(id: ID!): TypeA
+                        }
+
+                        type TypeA {
+                            echo: String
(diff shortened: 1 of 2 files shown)


lib/src/test/kotlin/graphql/nadel/validation/NadelFieldValidationTest.kt
```diff
@@ -558,5 +558,123 @@ class NadelFieldValidationTest : DescribeSpec({
...
+            )
+
+            val errors = validate(fixture)
+            assert(errors.size == 1)
+
+            val error = errors.assertSingleOfType<NadelSchemaValidationError>()
+            assert(error.message == "Could not find overall field TypeA.fieldFromExtension on the underlying type TypeA on service test")
+        }
+
+        it("fails if fields in extended Query are missing in underlying") {
```
felipe-gdr, line 652: This is the only test that fails with the previous code. I added a couple more above just as a documentation of how the code behaves in those scenarios.

bbakerman (approved):

PR #431, opened 2022-06-01
fix selection of source fields for batch hydration using input

temaEmelyan:

### DIFF
--- a/test/src/test/resources/fixtures/hydration/complex identified by/complex-identified-by-doesnt-use-all-input-fields-in-the-input-identified-by-argument.yml
+++ b/test/src/test/resources/fixtures/hydration/complex identified by/complex-identified-by-doesnt-use-all-input-fields-in-the-input-identified-by-argument.yml
@@ -1,0 +1,224 @@
+name: "complex identified by doesnt use all input fields in the inputIdentifiedBy argument"
+enabled: true
+overallSchema:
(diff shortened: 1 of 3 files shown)


test/src/test/resources/fixtures/hydration/complex identified by/complex-identified-by-doesnt-use-all-input-fields-in-the-input-identified-by-argument.yml
```diff
@@ -0,0 +1,224 @@
...
+    }
+    
+    type Activity {
+      id: ID
+      issue: Issue @hydrated(
+        service: "Issue"
+        field: "issues"
+        arguments: [{name: "issuesInput" value: "$source.context.issueHydrationInput"}]
+        inputIdentifiedBy: [
+          {sourceId: "context.issueHydrationInput.id" resultId: "issueId"}
```
temaEmelyan, line 16: `issuesInput` expects 2 fields: `id` and `site`. Before the fix, if we only used one field in the `inputIdentifiedBy`, only that one field was queried from `Activity`'s `context.issueHydrationInput` and the input for `issues` was constructed incorrectly. 

The fix is to always query all the fields that are required in the input

bbakerman (approved):

PR #438, opened 2022-08-16
Add support for default argument values in hydration

gnawf:
  
  - [X] Add tests that use __typename in queries. N/A
  - [X] Does this change work with all Nadel transformations (rename, type rename, hydration, etc)? Add tests for this. N/A
  - [X] Is it worth using hints for this change in order to be able to enable a percentage rollout? No
  - [X] Do we need to add integration tests for this change in the GraphQL Gateway? No
  - [X] Do we need a pollinator check for this? No

### DIFF
--- a/test/src/test/kotlin/graphql/nadel/tests/hooks/all-hydration-fields-are-seen-by-transformer.kt
+++ b/test/src/test/kotlin/graphql/nadel/tests/hooks/all-hydration-fields-are-seen-by-transformer.kt
@@ -15,12 +15,13 @@ import graphql.nadel.engine.transform.result.json.JsonNodes
 import graphql.nadel.tests.EngineTestHook
 import graphql.nadel.tests.UseHook
 import graphql.normalized.ExecutableNormalizedField
+import java.util.Collections.synchronizedSet
 
 @UseHook
 class `all-hydration-fields-are-seen-by-transformer` : EngineTestHook {
-    private val isApplicable = mutableListOf<String>()
-    private val transformField = mutableListOf<String>()
-    private val getResultInstructions = mutableListOf<String>()
+    private val isApplicable = synchronizedSet(mutableSetOf<String>())
+    private val transformField = synchronizedSet(mutableSetOf<String>())
+    private val getResultInstructions = synchronizedSet(mutableSetOf<String>())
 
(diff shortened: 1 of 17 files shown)


test/src/test/kotlin/graphql/nadel/tests/hooks/all-hydration-fields-are-seen-by-transformer.kt
```diff
@@ -15,12 +15,13 @@ import graphql.nadel.engine.transform.result.json.JsonNodes
...
+import java.util.Collections.synchronizedSet
 
 @UseHook
 class `all-hydration-fields-are-seen-by-transformer` : EngineTestHook {
-    private val isApplicable = mutableListOf<String>()
-    private val transformField = mutableListOf<String>()
-    private val getResultInstructions = mutableListOf<String>()
+    private val isApplicable = synchronizedSet(mutableSetOf<String>())
+    private val transformField = synchronizedSet(mutableSetOf<String>())
+    private val getResultInstructions = synchronizedSet(mutableSetOf<String>())
```
gnawf, line 24: Ran into some race condition here, so I had to fix it.

test/src/test/resources/fixtures/hydration/basic-hydration-with-default-array-argument-values.yml
```diff
@@ -0,0 +1,114 @@
...
+      name: String
+    }
+  # language=GraphQL
+  service1: |
+    type Query {
+      foo: Foo
+    }
+    type Foo {
+      id: ID
+      bar(test: [String] = ["Hello", "World"]): Bar @hydrated(
```
gnawf, line 20: Hydration is here. It uses `$argument.test` with the default value of `["Hello", "World"]`

test/src/test/resources/fixtures/hydration/basic-hydration-with-default-array-argument-values.yml
```diff
@@ -0,0 +1,114 @@
...
+          }
+        },
+        "extensions": {}
+      }
+  - serviceName: "service2"
+    request:
+      # language=GraphQL
+      query: |
+        query {
+          barById(id: "barId" test: ["Hello", "World"]) {
```
gnawf, line 88: Default value is seen here in the query to the underlying service

bbakerman (approved):
nice one

PR #440, opened 2022-09-02
remove hint removeHydrationSpecificExecutionCode

temaEmelyan:


temaEmelyan:
has been rolled out to 100% for some time now

test/src/test/kotlin/graphql/nadel/tests/hooks/all-hydration-fields-are-seen-by-transformer.kt
```diff
@@ -72,28 +67,34 @@ class `all-hydration-fields-are-seen-by-transformer` : EngineTestHook {
...
     override fun assertResult(result: ExecutionResult) {
-        assert(isApplicable == setOf(
-            "service1.foo",
-            "service1.bar",
-            "service1.name",
-            "service2.bars",
-            "service2.barById",
-            "service2.name",
-        ))
+        assert(
```
temaEmelyan, line 70: these changes are from `ktlintFormat`

bbakerman (approved):

PR #441, opened 2022-09-14
Remove logging when service call throws exception

felipe-gdr:
WIP

---

### DIFF
--- a/lib/src/main/java/graphql/nadel/NextgenEngine.kt
+++ b/lib/src/main/java/graphql/nadel/NextgenEngine.kt
@@ -31,6 +31,8 @@ import graphql.nadel.engine.util.provide
 import graphql.nadel.engine.util.singleOfType
 import graphql.nadel.engine.util.strictAssociateBy
 import graphql.nadel.hooks.ServiceExecutionHooks
+import graphql.nadel.instrumentation.parameters.ExecutionErrorData
+import graphql.nadel.instrumentation.parameters.OnServiceExecutionErrorParameters
 import graphql.nadel.instrumentation.parameters.NadelInstrumentationTimingParameters.RootStep
 import graphql.nadel.util.OperationNameUtil
 import graphql.normalized.ExecutableNormalizedField
@@ -262,6 +264,12 @@ class NextgenEngine @JvmOverloads constructor(
             val errorMessage = "An exception occurred invoking the service '${service.name}'"
             val errorMessageNotSafe = "$errorMessage: ${e.message}"
             val executionId = serviceExecParams.executionId.toString()
+
+            instrumentation.onError(OnServiceExecutionErrorParameters(
+                message = errorMessage,
+                exception = e,
+                errorData = ExecutionErrorData(executionId)
+            ))
 
             newServiceExecutionResult(
                 errors = mutableListOf(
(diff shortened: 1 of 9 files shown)


lib/src/main/java/graphql/nadel/NextgenEngine.kt
```diff
@@ -188,6 +192,7 @@ class NextgenEngine @JvmOverloads constructor(
         topLevelField: ExecutableNormalizedField,
         service: Service,
         executionContext: NadelExecutionContext,
+        instrumentationState: InstrumentationState?,
```
gnawf, line 195: Can we pull the `InstrumentationState` out of `NadelExecutionContext` instead of passing in as another parameter?

lib/src/main/java/graphql/nadel/instrumentation/ChainedNadelInstrumentation.kt
```diff
@@ -49,6 +50,13 @@ class ChainedNadelInstrumentation(
         }
     }
 
+    override fun  onError(parameters: NadelInstrumentationOnErrorParameters) {
```
gnawf, line 53: Nit: double space between `fun` and `onError`

lib/src/main/java/graphql/nadel/instrumentation/parameters/NadelInstrumentationOnErrorParameters.kt
```diff
@@ -0,0 +1,25 @@
...
+    val errorData: Any?,
+    private val instrumentationState: InstrumentationState?,
+) {
+    fun <T : InstrumentationState?> getInstrumentationState(): T? {
+        @Suppress("UNCHECKED_CAST") // trust the caller
+        return instrumentationState as T?
+    }
+}
+
+data class OnServiceExecutionErrorData(
```
gnawf, line 18: Can we do

```kotlin
sealed interface NadelOnErrorData {
    data class OnServiceExecution(
        val executionId: String,
        val serviceName: String,
    ) : NadelOnErrorData
}
```

And change

```diff
-     val errorData: Any?,
+     val errorData: NadelOnErrorData,
```

Just so it's stricter and easier to consume. I hate seeing `Any?` in API and then having to guess what the possible values are.

gnawf (changes requested):
Looks great, just one comment about the data type and parameters.

lib/src/main/java/graphql/nadel/instrumentation/parameters/NadelInstrumentationOnErrorParameters.kt
felipe-gdr: that's awesome!

lib/src/main/java/graphql/nadel/instrumentation/NadelInstrumentation.kt
```diff
@@ -100,4 +101,14 @@ interface NadelInstrumentation {
...
     }
+
+    /**
+     * Called when Nadel encounters an error that can be useful for the caller code - for logging or metrics, for example.
+     *
+     * The nature of the error can be obtained by looking at the value of [NadelInstrumentationOnErrorParameters.errorType].
+     *
+     *  @param parameters to this step
+     */
+    fun onError(parameters: NadelInstrumentationOnErrorParameters) {
```
gnawf, line 112: Maybe by default it could have done the same behavior as before. But not important, since we're definitely going to add custom behavior.

gnawf (approved):
Looks good.

PR #446, opened 2023-01-25
Fix __typename behavior and add missing subselections in result

gnawf:
This addresses the issues we've run into where `__typename` always resolves.

Also ensures there are no absent fields in the result. Fills in the absent fields with value of `null`.


lib/src/main/java/graphql/nadel/NadelResultMerger.kt
```diff
@@ -0,0 +1,156 @@
...
+            .errors(errors)
+            .build()
+    }
+
+    private fun fixData(
+        fields: List<NadelFieldAndService>,
+        engineSchema: GraphQLSchema,
+        data: MutableJsonMap,
+    ): MutableJsonMap? {
+        val requiredFieldMap = buildRequiredFieldMap(fields, engineSchema)
```
gnawf, line 55: So what this does is return a list of fields that should be in the result e.g. given a query

```graphql
query {
  projects {
    project(id: "1") {
      key
    }
    board(id: "1") {
      id
    }
  }
  issueById(id: "1") {
    key
  }
}
```

Then if `projects` is not null then we need both `project` and `board` in the result i.e. the following is a _valid_ result

```json
{
  "data": {
    "projects": {
      "project": {
        "key": "TEST"
      },
      "board": {
        "id": "1"
      }
    },
    "issueById": {
      "key": "TEST-1"
    }
  }
}
```

But, the following are not.

```json
{
  "data": {
    "projects": {
      "project": {
        "key": "TEST"
      }
    },
    "issueById": {
      "key": "TEST-1"
    }
  }
}
```

(above is missing `projects.board`)

```json
{
  "data": {
    "projects": {
      "board": {
        "id": "1"
      }
    },
    "issueById": {
      "key": "TEST-1"
    }
  }
}
```

(above is missing `projects.project`)

```json
{
  "data": {
    "projects": {
      "project": {
        "key": "TEST"
      },
      "board": {
        "id": "1"
      }
    }
  }
}
```

(above is missing `issueById`)

etc.

The `requiredFieldMap` for this query is

```json
{
  "projects": [
    "project",
    "board"
  ],
  "issueById": []
}
```

lib/src/main/java/graphql/nadel/NadelResultMerger.kt
```diff
@@ -0,0 +1,156 @@
...
+
+            if (isNamespacedField(field, engineSchema)) {
+                requiredChildFields.addAll(field.children)
+            }
+        }
+
+        return requiredFields
+    }
+
+    private fun putAndMergeTopLevelData(data: MutableJsonMap, oneData: JsonMap) {
```
gnawf, line 140: This is pulled from GraphQLUtil.kt and is the same

lib/src/main/java/graphql/nadel/NextgenEngine.kt
```diff
@@ -150,28 +149,29 @@ class NextgenEngine @JvmOverloads constructor(
             )
 
             val result: ExecutionResult = try {
-                mergeResults(
```
gnawf, line 153: This change just moves the `mergeResults` to a function call below.

lib/src/main/java/graphql/nadel/NextgenEngine.kt
```diff
@@ -296,12 +296,7 @@ class NextgenEngine @JvmOverloads constructor(
             )
         }
 
-        return serviceExecResult.copy(
-            data = serviceExecResult.data.let { data ->
-                data?.takeIf { transformedQuery.resultKey in data }
-                    ?: mutableMapOf(transformedQuery.resultKey to null)
```
gnawf, line 302: This code was here to ensure that the result always has the top level field.

This has been moved to `NadelResultMerger` and now handles namespaced fields too.

lib/src/main/java/graphql/nadel/util/NamespacedUtil.kt
```diff
@@ -18,4 +22,16 @@ object NamespacedUtil {
...
+        return overallField
+            .getFieldDefinitions(schema)
+            .any {
+                isNamespacedField(it)
+            }
+    }
+
+    fun isNamespacedField(definition: GraphQLFieldDefinition): Boolean {
+        return definition.hasAppliedDirective(namespacedDirectiveDefinition.name)
+    }
```
gnawf, line 36: Moved from `NadelNamespacedFields`

temaEmelyan (approved):

PR #447, opened 2023-01-30
Feature flag new result merger

gnawf:
Feature flagging the new result merger.


lib/src/main/java/graphql/nadel/NextgenEngine.kt
```diff
@@ -296,7 +300,12 @@ class NextgenEngine @JvmOverloads constructor(
             )
         }
 
-        return serviceExecResult
+        return serviceExecResult.copy(
+            data = serviceExecResult.data.let { data ->
+                data.takeIf { transformedQuery.resultKey in data }
+                    ?: mutableMapOf(transformedQuery.resultKey to null)
+            },
+        )
```
gnawf, line 308: Brought this back so the old method would still work.

lib/src/main/java/graphql/nadel/hints/NewResultMergerAndNamespacedTypename.kt
```diff
@@ -0,0 +1,16 @@
+package graphql.nadel.hints
+
+fun interface NewResultMergerAndNamespacedTypename {
```
gnawf, line 3: Changed it so one FF controls both behaviors since they're so tied together.

gnawf (commented):
I also used the test suite from an old version of the repo 84996418abeb532642388dff1c109d5e6e4777e2 and the tests pass on this version of the code.

temaEmelyan (approved):
👍

PR #448, opened 2023-02-13
Performance improvements around result transforming

gnawf:
Removes `JsonNodePath` and thus a bunch of array allocations.

This also stores the actual `JsonNode` to edit in the instruction instead of just its path (which Nadel looks up again later).

Before

```
Benchmark                                             Mode  Cnt  Score   Error  Units
LargeRenamedResponseBenchmarkRunner.benchMarkAvgTime  avgt   15  7.082 ± 0.187  ms/o
```

After

```
Benchmark                                             Mode  Cnt  Score   Error  Units
LargeRenamedResponseBenchmarkRunner.benchMarkAvgTime  avgt   15  2.469 ± 0.046  ms/op
```


lib/src/main/java/graphql/nadel/engine/util/GraphQLUtil.kt
```diff
@@ -225,7 +226,33 @@ fun ExecutableNormalizedField.copyWithChildren(children: List<ExecutableNormaliz
         }
 }
 
-val ExecutableNormalizedField.queryPath: NadelQueryPath get() = NadelQueryPath(listOfResultKeys)
+val ExecutableNormalizedField.queryPath: NadelQueryPath
+    get() = NadelQueryPath(
```
gnawf, line 230: This is more performant.

bbakerman (approved):
This is deep change in ma ny ways - however given Nadels excellent test coverage - the gact that you can refactor here significantly and yet have it pass all the tests it s testement to the approach

Great stuff

PR #451, opened 2023-04-10
bump gql java

temaEmelyan:


gnawf (approved):
Seems fine, but any reason for that particular version? Seems to be a week old. I'm guessing you made the change last week and only just opened the PR?

Not a big deal, doesn't seem like there's much in the newer GraphQL Java commits anyway.

PR #455, opened 2023-06-21
Better union exclusion

gnawf:


lib/src/main/java/graphql/nadel/validation/util/NadelGetReachableTypes.kt
```diff
@@ -223,6 +226,15 @@ internal fun getReachableTypeNames(
...
         }
+
+        override fun visitGraphQLAppliedDirective(
+            node: GraphQLAppliedDirective,
+            context: TraverserContext<GraphQLSchemaElement>,
+        ): TraversalControl {
+            // Don't look into applied directives. Could be a shared directive.
+            // As long as the schema compiled then we don't care.
+            return ABORT
+        }
```
gnawf, line 237: We need this otherwise it'll start looking at directive argument types etc.

We don't want to validate those as we have tons of shared directives.

As long as the `GraphQLSchema`  compiles then we are happy.

lib/src/main/java/graphql/nadel/validation/NadelTypeValidation.kt
```diff
@@ -229,16 +232,30 @@ internal class NadelTypeValidation(
...
+                    .values
+                    .asSequence()
+                    .filterIsInstance<GraphQLFieldsContainer>()
+                    .flatMap {
+                        it.fieldDefinitions
+                    }
+                    .filter {
+                        it.type.unwrapAll().name == union.name
+                    }
+                    .all(min = 1) {
```
gnawf, line 252: I'm saying here we need _at least_ one `@hydrated` field that uses this.

Maybe we could drop this requirement… but I kept it in to satisfy some tests regarding not allowing synthetic unions. (no hydrations referenced in that test).

temaEmelyan (approved):

bbakerman (approved):
Nice one

PR #465, opened 2023-10-19
Pulls latest graphql-java version, which contains oneOf, and add some tests

felipe-gdr:

- [n/a ] Add tests that use __typename in queries
- [n/a ] Does this change work with all nadel transformations (rename, type rename, hydration, etc)? Add tests for this.
- [n/a] Is it worth using hints for this change in order to be able to enable a percentage rollout?
- [n/a ] Do we need to add integration tests for this change in the graphql gateway?
- [n/a ] Do we need a pollinator check for this?

### DIFF
--- a/test/src/test/resources/fixtures/oneOf/one-of-fails-when-2-values-are-passed.yml
+++ b/test/src/test/resources/fixtures/oneOf/one-of-fails-when-2-values-are-passed.yml
@@ -6,8 +6,8 @@ overallSchema:
       search(by: SearchInput): String
     }
     input SearchInput @oneOf {
-      name: String!
-      id: ID!
+      name: String
+      id: ID
     }
 underlyingSchema:
   MyService: |-
@@ -15,8 +15,8 @@ underlyingSchema:
       search(by: SearchInput): String
     }
     input SearchInput @oneOf {
-      name: String!
-      id: ID!
+      name: String
+      id: ID
     }
 query: |
   query myQuery {
--- a/test/src/test/resources/fixtures/oneOf/one-of-fails-when-2-values-are-passed.yml
+++ b/test/src/test/resources/fixtures/oneOf/one-of-fails-when-2-values-are-passed.yml
@@ -1,0 +1,38 @@
+name: "oneOf fails when 2 values are passed"
+enabled: true
+overallSchema:
+  MyService: |
+    type Query {
+      search(by: SearchInput): String
+    }
+    input SearchInput @oneOf {
+      name: String!
+      id: ID!
+    }
+underlyingSchema:
+  MyService: |-
+    type Query {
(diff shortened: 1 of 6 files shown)


test/src/test/resources/fixtures/oneOf/one-of-fails-when-2-values-are-passed.yml
```diff
@@ -0,0 +1,38 @@
+name: "oneOf fails when 2 values are passed"
+enabled: true
+overallSchema:
+  MyService: |
+    type Query {
+      search(by: SearchInput): String
```
dondonz, line 6: You prob already had this in mind - when the new graphql-java build version containing your PR https://github.com/graphql-java/graphql-java/pull/3353 is merged in, can you add a test with the non-null input case?

felipe-gdr: Yep, that's a good idea!

dondonz (dismissed):
Thanks for doing the upgrade and bugfix!

felipe-gdr:
@gnawf 

> Do we need validation to ensure @oneOf also exists in the underlying schema etc?

Yep, that's a fair concern, but I don't think any extra validation would be needed. 
The enforcement of the `@oneOf` spec will be done at the gateway level, so underlying services are guaranteed to receive a valid query, even if the implementation doesn't yet support `@oneOf`.
Also, `@oneOf` is never applied in queries, only in schema elements. So its usage can't result in queries that will be invalid in the underlying schema.

> Do we need a test with variables?
That's not a bad idea. I'm gonna add a simple one for reference. I want to avoid writing tests to cover all the edge cases, though, as this should be well covered in the graphql-java tests

> Is it permitted to specify multiple argument values from variables and only one of them is non-null?
That's not permitted. Exactly one field should be present, and it should have a non-null value.
The RFC has a bunch of examples that describe the expected behaviour in different scenarios: https://github.com/graphql/graphql-spec/pull/948/files (there's a table with the examples)

test/src/test/resources/fixtures/oneOf/one-of-fails-when-invalid-variables-are-passed.yml
```diff
@@ -0,0 +1,39 @@
...
+    }
+    input SearchInput @oneOf {
+      name: String
+      id: ID
+    }
+query: |
+  query myQuery($name: String, $id: ID) {
+    search(by: {name: $name, id: $id})
+  }
+variables:
```
felipe-gdr, line 25: @gnawf here's a test with variables ;-)

gnawf (approved):
Yea fair point about the tests. Ok this looks good.

PR #467, opened 2023-10-30
Pull graphql-java with oneof fix and add more tests

felipe-gdr:

- [n ] Add tests that use __typename in queries
- [n ] Does this change work with all nadel transformations (rename, type rename, hydration, etc)? Add tests for this.
- [n ] Is it worth using hints for this change in order to be able to enable a percentage rollout?
- [y ] Do we need to add integration tests for this change in the graphql gateway?
- [n ] Do we need a pollinator check for this?

### DIFF
--- a/lib/build.gradle.kts
+++ b/lib/build.gradle.kts
@@ -7,7 +7,7 @@ plugins {
     id("com.bnorm.power.kotlin-power-assert")
 }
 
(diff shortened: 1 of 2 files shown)


gnawf (approved):
Cool

sbarker2 (approved):

PR #474, opened 2023-11-08
Adding When condition input types

sbarker2:


gnawf (changes requested):

lib/src/main/java/graphql/nadel/schema/NadelDirectives.kt
```diff
@@ -482,6 +525,15 @@ object NadelDirectives {
...
-        return ValuesResolver.valueToInternalValue(graphQLArgument.argumentValue, graphQLArgument.type, GraphQLContext.getDefault(), Locale.getDefault()) as T
+        return ValuesResolver.valueToInternalValue(
+            graphQLArgument.argumentValue,
+            graphQLArgument.type,
+            GraphQLContext.getDefault(),
+            Locale.getDefault()
+        ) as T
+    }
+
+    inline fun parseType(sdl: String): InputObjectTypeDefinition {
```
gnawf, line 536: This needs to be generic (see my other comment for more).

gnawf (approved):

PR #477, opened 2023-12-05
Adding conditional hydration validation

sbarker2:

### DIFF
--- a/lib/src/main/java/graphql/nadel/validation/NadelHydrationWhenConditionValidation.kt
+++ b/lib/src/main/java/graphql/nadel/validation/NadelHydrationWhenConditionValidation.kt
@@ -2,6 +2,7 @@ package graphql.nadel.validation
 
 import graphql.Scalars
 import graphql.nadel.dsl.UnderlyingServiceHydration
+import graphql.nadel.engine.util.unwrapAll
 import graphql.nadel.engine.util.unwrapNonNull
 import graphql.schema.GraphQLFieldDefinition
 import graphql.schema.GraphQLFieldsContainer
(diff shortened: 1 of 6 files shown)


gnawf (changes requested):
Some changes, but should be good to go after.

lib/src/main/java/graphql/nadel/validation/NadelHydrationWhenConditionValidation.kt
```diff
@@ -0,0 +1,71 @@
...
+        hydration: UnderlyingServiceHydration,
+    ): NadelSchemaValidationError? {
+        if (hydration.conditionalHydration == null) {
+            return null
+        }
+
+        val whenConditionSourceFieldName: String = hydration.conditionalHydration?.get("sourceField") as String
+        val whenConditionSourceField: GraphQLFieldDefinition = (parent.overall as GraphQLFieldsContainer).getField(whenConditionSourceFieldName)
+
+        if (whenConditionSourceField.type !is GraphQLScalarType){
```
gnawf, line 24: This would fail if the field type is non-nullable.

We need to let fields be non-nullable.

lib/src/main/java/graphql/nadel/validation/NadelHydrationWhenConditionValidation.kt
```diff
@@ -0,0 +1,71 @@
...
+            return return NadelSchemaValidationError.WhenConditionUnsupportedFieldType(whenConditionSourceFieldName, "", overallField)
```
gnawf, line 25: Did you leave the type name here blank on purpose?

You should be able to get a type name using something like `GraphQLTypeUtil.simplePrint(whenConditionSourceField.type)`

gnawf (changes requested):

lib/src/main/java/graphql/nadel/validation/NadelHydrationWhenConditionValidation.kt
```diff
@@ -0,0 +1,82 @@
...
+        hydration: UnderlyingServiceHydration,
+        isBatchHydration: Boolean,
+        actorFieldName: String,
+    ): NadelSchemaValidationError? {
+        if (hydration.conditionalHydration == null) {
+            return null
+        }
+
+        val whenConditionSourceFieldName: String = hydration.conditionalHydration?.get("sourceField") as String
+        val whenConditionSourceField = (parent.overall as GraphQLFieldsContainer).getField(whenConditionSourceFieldName)
```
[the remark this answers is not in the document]
sbarker2: isn't that just every other hydration test?

gnawf (approved):

PR #488, opened 2023-12-13
when condition check across alll hydrations

sbarker2:


lib/src/main/java/graphql/nadel/validation/NadelHydrationValidation.kt
```diff
@@ -274,16 +282,34 @@ internal class NadelHydrationValidation(
...
+        hydrations: List<UnderlyingServiceHydration>,
+        parent: NadelServiceSchemaElement,
+        overallField: GraphQLFieldDefinition,
+    ): NadelSchemaValidationError.SomeHydrationsHaveMissingConditions? {
+        //if one hydration has a condition, then they all must, so lets check the first one
+        var shouldHaveConditions = true
+        if (hydrations.first().conditionalHydration == null) {
+            shouldHaveConditions = false
+        }
+        if (hydrations.any { (it.conditionalHydration == null) == shouldHaveConditions }) { //i.e. (if it has no condition but it should have condition) OR  (if it has condition but shouldn't have condition)
```
gnawf, line 310: Hmm, I am not so sure about this logic.

I think this test case would break it

```
@hydrated()
@hydrated(when: )
@hydrated()
```

You can't just check any of them, you need to check all of them.

i.e. logic should be if any `when` exists, then all must have `when`

lib/src/main/java/graphql/nadel/validation/NadelSchemaValidationError.kt
```diff
@@ -430,21 +431,32 @@ sealed interface NadelSchemaValidationError {
...
+        override val subject = overallField
+    }
+
+    data class SomeHydrationsHaveMissingConditions(
+        val parentType: NadelServiceSchemaElement,
+        val overallField: GraphQLFieldDefinition,
     ) : NadelSchemaValidationError {
-        override val message = "When condition source field \"${sourceFieldName}\" of type \"${sourceFieldTypeName}\" " +
-            "needs to be of type String or ID in order to use the ${predicateType} predicate."
+        val of = makeFieldCoordinates(parentType.overall.name, overallField.name)
```
gnawf, line 458: This should not be a property. Refer to all the other message builders, they use `run` to limit the scope of variables used in making the message.

lib/src/main/java/graphql/nadel/validation/NadelSchemaValidationError.kt
```diff
@@ -430,21 +431,32 @@ sealed interface NadelSchemaValidationError {
...
+        override val message = "Some hydrations of field ${of} are missing a when condition"
```
gnawf, line 459: Don't use `${}` for one property (your IDE should have a warning for this).

lib/src/main/java/graphql/nadel/validation/NadelHydrationValidation.kt
```diff
@@ -274,16 +282,34 @@ internal class NadelHydrationValidation(
...
+                            actorField.name
+                        )
+                    )
                 }
                 return emptyList()
             }
         }
     }
+
+    private fun validateConditionsOnAllHydrations(
```
gnawf, line 300: Shouldn't this code belong in `NadelHydrationWhenConditionValidation` ?

gnawf (changes requested):

lib/src/main/java/graphql/nadel/validation/NadelHydrationValidation.kt
sbarker2: ahh yeah, good catch. I was tryna keep it a small amount of lines but it backfired a little.
changed the logic now to handle that

gnawf (approved):

PR #493, opened 2023-12-17
Fix validation and outdated schema problem

gnawf:


lib/src/main/java/graphql/nadel/schema/NadelDirectives.kt
```diff
@@ -223,7 +223,10 @@ object NadelDirectives {
                 val identifiedByValues = resolveArgumentValue<List<Any>>(inputIdentifiedBy)
                 val identifiedBy = createObjectIdentifiers(identifiedByValues)
 
-                val conditionalHydration = buildConditionalHydrationObject(directive.getArgument("when"))?.result
+                val conditionalHydration = directive.getArgument("when")
+                    ?.let {
```
gnawf, line 227: Handle case where schema does not define `when` and the value is null.

lib/src/main/java/graphql/nadel/validation/NadelHydrationWhenConditionValidation.kt
```diff
@@ -26,17 +25,19 @@ internal class NadelHydrationWhenConditionValidation() {
...
+        val whenConditionSourceFieldType = whenConditionSourceField.type.unwrapNonNull()
+        if (whenConditionSourceFieldType !is GraphQLScalarType) {
             return NadelSchemaValidationError.WhenConditionUnsupportedFieldType(
                 whenConditionSourceFieldName,
                 GraphQLTypeUtil.simplePrint(whenConditionSourceField.type),
-                overallField
+                overallField,
             )
         }
-        val whenConditionSourceFieldTypeName: String = (whenConditionSourceField.type as GraphQLScalarType).name
```
gnawf, line 39: Avoid manual casting. There is a bug where a non-nullable type causes this to crash.

sbarker2: Thanks for fixing this 👍

lib/src/main/java/graphql/nadel/schema/NadelDirectives.kt
sbarker2: makes sense, thanks for picking it up 👍

sbarker2 (approved):

sbarker2 (approved):

PR #503, opened 2023-12-19
Prevent multiple source fields even if list is not leaf field

gnawf:


lib/src/test/kotlin/graphql/nadel/validation/NadelHydrationValidationTest2.kt
```diff
@@ -170,6 +170,97 @@ class NadelHydrationValidationTest2 {
         assertTrue(errors.single().subject.name == "data")
     }
 
+    @Test
+    fun `prohibit multiple source fields where list not the leaf`() {
```
gnawf, line 174: This should actually be an allowed feature, but it's not supported right now.

lib/src/main/java/graphql/nadel/validation/NadelHydrationValidation.kt
```diff
@@ -120,30 +121,31 @@ internal class NadelHydrationValidation(
...
                         .arguments
                         .asSequence()
-                        .mapNotNull { argument ->
-                            argument.remoteArgumentSource.pathToField
-                        }
-                        .any { path ->
-                            parentType.getFieldAt(path)?.type?.unwrapNonNull()?.isList == true
+                        .mapNotNull { it.remoteArgumentSource.pathToField }
+                }
+                .toList()
```
gnawf, line 132: Pulled the paths out to a list

lib/src/main/java/graphql/nadel/validation/NadelHydrationValidation.kt
```diff
@@ -120,30 +121,31 @@ internal class NadelHydrationValidation(
...
+
+            val parentType = parent.underlying as GraphQLFieldsContainer
+            val anyListSourceInputField = pathsToSourceFields
+                .any { pathToField ->
+                    parentType
+                        .getFieldsAlong(pathToField)
+                        .any { field ->
+                            field.type.unwrapNonNull().isList
```
gnawf, line 140: This is the actual change.

e.g. for `$source.contexts.userId` where `contexts` is a list but `userId` is not a list

Previously we only checked the end field to see if it was a list, this checks everything.

sbarker2 (approved):

PR #507, opened 2024-01-03
Fix source input edge case

gnawf:


test/src/test/resources/fixtures/new hydration/new batching/new-batching-absent-source-input.yml
```diff
@@ -0,0 +1,215 @@
...
+          }
+        }
+      variables: { }
+    # language=JSON
+    response: |-
+      {
+        "data": {
+          "activity": [
+            {
+              "__typename__batch_hydration__content": "Activity"
```
andimarek, line 113: that is not a legal answer, is it? I mean there is no fragment/type selector involved and therefore every key must be present.

gnawf: Yeah I was creating tests with different theories with how the bug was occurring.

I just left it in anyway. I don't think there's much harm.

(I eventually just pulled in the data from the logs and got an exact replication).

sbarker2 (approved):

bbakerman (approved):

PR #509, opened 2024-01-17
Clean up conditional hydration names

gnawf:


lib/build.gradle.kts
```diff
@@ -40,6 +40,8 @@ dependencies {
     testImplementation("io.kotest:kotest-runner-junit5:5.8.0")
     testImplementation("io.kotest:kotest-framework-datatest:5.8.0")
     testImplementation("io.mockk:mockk:1.13.8")
+
+    testImplementation("com.tngtech.archunit:archunit:1.2.1")
```
gnawf, line 44: To mandate Nadel prefix.

lib/src/main/java/graphql/nadel/dsl/NadelHydrationConditionDefinition.kt
```diff
@@ -0,0 +1,16 @@
+package graphql.nadel.dsl
+
+data class NadelHydrationConditionDefinition(
```
gnawf, line 3: Used to be named `WhenConditionDefinition`

Three things

1. Added `Nadel` prefix
3. Removed redundant `When`
4. Mentioned `Hydration`

lib/src/main/java/graphql/nadel/dsl/NadelHydrationDefinition.kt
```diff
@@ -10,7 +10,7 @@ data class UnderlyingServiceHydration(
     val isBatched: Boolean,
     val batchSize: Int,
     val timeout: Int,
-    val conditionalHydration: WhenConditionResultDefinition?
```
gnawf, line 13: `Hydration` is redundant since we're in a hydration class.

sbarker2 (approved):

PR #510, opened 2024-01-17
Clean up more conditional hydration names

gnawf:


lib/src/main/java/graphql/nadel/validation/NadelHydrationConditionValidation.kt
gnawf, line 1: This is the same class, but too many renames so it thinks it's different now.

sbarker2 (approved):

PR #513, opened 2024-01-18
add null test cases to conditional hydration tests

sbarker2:
note: the batch null test cases will be in the batch PR


gnawf (approved):
Great

PR #515, opened 2024-02-12
Adding ability to generate queries containing @defer (but not with defer execution yet)

sbarker2:
- Hide new functionality behind a feature flag (Nadel hints)
- Nadel can generate queries containing @defer and send them to underlying services

Actual defer execution will come in a later PR


lib/src/main/java/graphql/nadel/incremental/NadelIncrementalServiceExecutionResult.kt
```diff
@@ -0,0 +1,14 @@
+package graphql.nadel.incremental
+
+import graphql.nadel.ServiceExecutionResult
+import org.reactivestreams.Publisher
+
+class NadelIncrementalServiceExecutionResult(
```
felipe-gdr, line 6: Do we need to introduce this class now? We're not dealing with incremental responses yet. 
This PR is only handling the generation of queries containing `@defer`.

You could change the title of this PR accordingly, so it clearly indicates the stage of the implementation that is being done.

lib/src/main/java/graphql/nadel/ServiceExecutionResult.kt
```diff
@@ -1,6 +1,6 @@
 package graphql.nadel
 
-class ServiceExecutionResult @JvmOverloads constructor(
+open class ServiceExecutionResult @JvmOverloads constructor(
```
felipe-gdr, line 3: This will be extended in a future PR

felipe-gdr (approved):

PR #517, opened 2024-02-28
Fix polymorphic hydration parallelism

gnawf:

### DIFF
--- a/lib/src/main/java/graphql/nadel/engine/transform/hydration/batch/NadelNewBatchHydrator.kt
+++ b/lib/src/main/java/graphql/nadel/engine/transform/hydration/batch/NadelNewBatchHydrator.kt
@@ -207,14 +207,26 @@ internal class NadelNewBatchHydrator(
         val sourceObjectsMetadata = getSourceObjectsMetadata(sourceObjects)
         val sourceInputsByInstruction = groupSourceInputsByInstruction(sourceObjectsMetadata)
 
-        val resultsByInstruction = sourceInputsByInstruction
-            .mapValues { (instruction, sourceInputs) ->
-                executeQueries(
-                    executionBlueprint = executionBlueprint,
-                    instruction = instruction,
-                    sourceInputs = sourceInputs,
-                )
-            }
+        val resultsByInstruction = coroutineScope {
+            // It's important to understand that this is a List, not a Sequence.
+            // We want to kick off ALL the hydrations at the same time, _then_ wait for them after.
+            val deferredResults = sourceInputsByInstruction
+                .entries
+                .map { (instruction, sourceInputs) ->
+                    instruction to async {
+                        executeQueries(
+                            executionBlueprint = executionBlueprint,
+                            instruction = instruction,
+                            sourceInputs = sourceInputs,
+                        )
+                    }
+                }
+
+            deferredResults
+                .associate { (instruction, deferred) ->
+                    instruction to deferred.await()
+                }
+        }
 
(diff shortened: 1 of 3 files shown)


test/src/test/kotlin/graphql/nadel/tests/hooks/polymorphic-hydration-hook-using-alias-helper.kt
```diff
@@ -33,12 +46,10 @@ private class PolymorphicHydrationHookUsingAliasHelper : NadelExecutionHooks {
...
-        return instruction.actorFieldDef.name.contains("pet") && hydrationArgumentValue.startsWith(
-            "pet", ignoreCase = true
-        ) ||
-            instruction.actorFieldDef.name.contains("human") && hydrationArgumentValue.startsWith(
-            "human", ignoreCase = true
-        )
+        return (instruction.actorFieldDef.name.contains("pet")
+            && hydrationArgumentValue.startsWith("pet", ignoreCase = true))
+            || (instruction.actorFieldDef.name.contains("human")
+            && hydrationArgumentValue.startsWith("human", ignoreCase = true))
```
gnawf, line 52: Reformat

lib/src/main/java/graphql/nadel/engine/transform/hydration/batch/NadelNewBatchHydrator.kt
```diff
@@ -207,14 +207,26 @@ internal class NadelNewBatchHydrator(
...
-            .mapValues { (instruction, sourceInputs) ->
-                executeQueries(
-                    executionBlueprint = executionBlueprint,
-                    instruction = instruction,
-                    sourceInputs = sourceInputs,
-                )
-            }
+        val resultsByInstruction = coroutineScope {
+            // It's important to understand that this is a List, not a Sequence.
+            // We want to kick off ALL the hydrations at the same time, _then_ wait for them after.
```
gnawf, line 212: Left a comment here, but this is also captured in a test.

bbakerman (approved):

sbarker2 (approved):

test/src/test/kotlin/graphql/nadel/tests/hooks/polymorphic-hydration-hook-using-alias-helper.kt
```diff
@@ -57,3 +68,46 @@ class `batch-polymorphic-hydration-actor-fields-are-in-the-same-service` : Polym
...
+                    override fun getServiceExecution(serviceName: String): ServiceExecution {
+                        val defaultServiceExecution = defaultServiceExecutionFactory.getServiceExecution(serviceName)
+
+                        return ServiceExecution { params ->
+                            coroutineScope
+                                .future {
+                                    servicesExecuted.add(serviceName)
+                                    if (params.isHydrationCall) {
+                                        delay(Random.nextInt(400, 800).milliseconds)
+                                        assertTrue(servicesExecuted == hashSetOf("people", "pets", "foo"))
```
felipe-gdr-atlassian, line 92: so by checking that all 3 services have started this wrapped execution code before any of the actual executions started (which happens in the `thenCompose` below), we can assert that the calls happened in parallel?

gnawf: There's random delay but that's just to ensure the service execution order doesn't matter. Either way whenever the first service finishes, the set MUST have all the service names in it.

felipe-gdr-atlassian (approved):

PR #518, opened 2024-03-03
defer classes to talk to gateway

sbarker2:


felipe-gdr (approved):

lib/src/main/java/graphql/nadel/ServiceExecutionResult.kt
```diff
@@ -1,7 +1,16 @@
 package graphql.nadel
 
+import graphql.incremental.DelayedIncrementalPartialResult
+import org.reactivestreams.Publisher
+
 open class ServiceExecutionResult @JvmOverloads constructor(
```
felipe-gdr, line 6: can we make `ServiceExecutionResult` `sealed`, and _not_ make `NadelncrementalServiceExecutionResult` `open`?

sbarker2: oh yeah, forgot to change that back

bbakerman (dismissed):

lib/src/main/java/graphql/nadel/ServiceExecutionResult.kt
sbarker2: @felipe-gdr Hmm, I'm not sure if this is what we want actually. Sealed types cannot be instantiated, but it needs to be instantiated here in IntrospectionService.kt

https://github.com/atlassian-labs/nadel/blob/9897a5db6024cd179c8b1840ee9043448618cb7c/lib/src/main/java/graphql/nadel/engine/blueprint/IntrospectionService.kt#L48C1-L51C18

felipe-gdr: Ah right
[A sealed class is [abstract](https://kotlinlang.org/docs/classes.html#abstract-classes) by itself, it cannot be instantiated directly](https://kotlinlang.org/docs/sealed-classes.html#location-of-direct-subclasses)

We could create a `ServiceExecutionResultImpl` class that extends it and instantiate that instead.

sbarker2: Changed, thanks

felipe-gdr (approved):

PR #519, opened 2024-03-04
don't call underlying services on empty queries

temaEmelyan:


lib/src/main/java/graphql/nadel/NextgenEngine.kt
```diff
@@ -362,6 +366,32 @@ internal class NextgenEngine(
...
+        }
+    }
+
+    private fun onlyTopLevelTypenameField(executableNormalizedField: ExecutableNormalizedField): Boolean {
+        if (executableNormalizedField.fieldName == TypeNameMetaFieldDef.name) {
+            return true
+        }
+        val operationType = engineSchema.getTypeAs<GraphQLObjectType>(executableNormalizedField.singleObjectTypeName)
+        val topLevelFieldDefinition =
+            operationType.fieldDefinitions.firstOrNull { it.name == executableNormalizedField.name }
```
gnawf, line 387: `getField(executableNormalizedField.name)` ?

test/src/test/kotlin/graphql/nadel/tests/transforms/RemoveFieldTestTransform.kt
```diff
@@ -78,11 +77,11 @@ class RemoveFieldTestTransform : NadelTransform<GraphQLError> {
         state: GraphQLError,
         nodes: JsonNodes,
     ): List<NadelResultInstruction> {
-        val parentNodes = JsonNodeExtractor.getNodesAt(
-            data = result.data,
-            queryPath = underlyingParentField?.queryPath ?: NadelQueryPath.root,
-            flatten = true,
-        )
+        val parentNodes = JsonNodes(result.data)
```
gnawf, line 80: This should use the `nodes` param instead of creating a new one.

I mean it's test code so it doesn't matter but in reality we want to share one `JsonNodes` because it caches some JSON traversals.

gnawf (dismissed):

gnawf (approved):

PR #520, opened 2024-03-05
add defer types to test fixture

sbarker2:
Note: the tests arent asserting the incremental responses yet for deferred requests, it is just using the initial responses at the moment, the work for the incremental responses is in progress and will be in a follow-up PR. 
The aim of this PR is just to add all the types of the test fixture.

### DIFF
--- a/test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
+++ b/test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
@@ -180,10 +180,9 @@ private suspend fun execute(
 
                                 if (indexOfCall != null) {
                                     val serviceCall = serviceCalls.removeAt(indexOfCall)
-                                    if( serviceCall.incrementalResponse != null){
+                                    if (serviceCall.incrementalResponse != null) {
                                         serviceCall.incrementalResponse.initialResponse //for now, just return initial response
-                                    }
-                                    else {
+                                    } else {
                                         serviceCall.response!!
                                     }
                                 } else {
--- a/test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
+++ b/test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
@@ -180,7 +180,12 @@ private suspend fun execute(
 
                                 if (indexOfCall != null) {
                                     val serviceCall = serviceCalls.removeAt(indexOfCall)
-                                    serviceCall.response
+                                    if( serviceCall.incrementalResponse != null){
+                                        serviceCall.incrementalResponse.initialResponse //for now, just return initial response
+                                    }
+                                    else {
+                                        serviceCall.response!!
+                                    }
                                 } else {
                                     fail(
                                         """Unable to match service call 
--- a/test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
+++ b/test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
@@ -168,6 +168,18 @@ private suspend fun execute(
                             )
                             printSyncLine(actualQuery)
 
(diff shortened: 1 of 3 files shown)


test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
```diff
@@ -180,7 +180,11 @@ private suspend fun execute(
 
                                 if (indexOfCall != null) {
                                     val serviceCall = serviceCalls.removeAt(indexOfCall)
-                                    serviceCall.response
+                                    if (serviceCall.incrementalResponse != null) {
+                                        serviceCall.incrementalResponse.initialResponse //for now, just return initial response
+                                    } else {
+                                        serviceCall.response!!
+                                    }
```
felipe-gdr, line 187: I think your next PR will cover this, but, ideally, the engine would fail with a descriptive message when fixtures haven't defined neither `incrementalResponse` or `response`.

test/src/test/resources/fixtures/defer/defer-with-label.yml
```diff
@@ -48,25 +48,62 @@ serviceCalls:
...
-response: |-
-  {
-    "data": {
-      "defer": {
-        "hello": "world"
-      }
-    },
-    "extensions": {}
-  }
+#response: |-
```
felipe-gdr, line 79: delete these commented out lines of code?

test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
```diff
@@ -180,7 +180,11 @@ private suspend fun execute(
 
                                 if (indexOfCall != null) {
                                     val serviceCall = serviceCalls.removeAt(indexOfCall)
-                                    serviceCall.response
+                                    if (serviceCall.incrementalResponse != null) {
```
felipe-gdr, line 183: let's fail the test when both `response` and `incrementalResponse` are defined in the fixture? It never makes sense to define both together

sbarker2: 👍

test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
```diff
@@ -168,6 +168,18 @@ private suspend fun execute(
                             )
                             printSyncLine(actualQuery)
 
+                            fun failWithFixtureContext(message: String): Nothing {
```
sbarker2, line 171: put into function because the fixture context is now useful for multiple fail messages

test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
```diff
@@ -180,17 +192,20 @@ private suspend fun execute(
 
                                 if (indexOfCall != null) {
                                     val serviceCall = serviceCalls.removeAt(indexOfCall)
-                                    serviceCall.response
+                                    val response = serviceCall.response
```
sbarker2, line 195: setting response to a variable, because kotlin thinks the state can change due to the way it was defined:
```
val response: JsonMap? by lazy {
...
}
```

test/src/test/kotlin/graphql/nadel/tests/TestFixture.kt
```diff
@@ -88,3 +94,20 @@ data class ExpectedException(
         )
     }
 }
+
+data class IncrementalResponse(
+    @JsonProperty("initialResponse")
```
felipe-gdr, line 99: I'm happy to merge this as it is. But as we discussed, it's worth trying to use the actual graphql-java types here to make deserialization easier.

felipe-gdr (approved):

PR #521, opened 2024-03-06
Adding defer to engine tests

sbarker2:
this PR is to enable us to run engine tests for the `@defer` directive according to the `graphql-spec`

### DIFF
--- a/test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
+++ b/test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
@@ -212,9 +212,10 @@ private suspend fun execute(
                                                 .incrementalItems(
                                                     incrementalDataVal.map{
                                                         DeferPayload.newDeferredItem()
-                                                            .label(it["label"] as String)
+                                                            .data(it["data"])
                                                             .path(it["path"] as List<Object>)
                                                             .apply {
+                                                                if (it["label"] != null) it["label"] as String
                                                                 if (it["extensions"] != null) extensions(it["extensions"] as Map<Any, Any>)
                                                                 if (it["errors"] != null) errors(it["errors"] as List<GraphQLError>)
                                                             }
@@ -222,9 +223,6 @@ private suspend fun execute(
                                                     }
                                                 )
                                                 .build()
-                                        }
-                                        serviceCall.incrementalResponse.delayedResponses.toTypedArray().map {
-                                            transformData(it)
                                         }
 
                                         val incrementalItemPublisher: Publisher<DelayedIncrementalPartialResult> = flowOf(*serviceCall.incrementalResponse.delayedResponses.toTypedArray()).map {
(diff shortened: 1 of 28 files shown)


test/src/test/kotlin/graphql/nadel/tests/transforms/RemoveFieldTestTransform.kt
```diff
@@ -39,7 +40,7 @@ class RemoveFieldTestTransform : NadelTransform<GraphQLError> {
             ?: return null
 
         if (objectType.getField(overallField.name)?.getDirective("toBeDeleted") != null) {
-            return ValidationError(ValidationErrorType.WrongType)
+            return newValidationError().validationErrorType(ValidationErrorType.WrongType).build()
```
sbarker2, line 43: `ValidationError()` is private in `graphql-java` now so using the builder

test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
```diff
@@ -180,7 +189,79 @@ private suspend fun execute(
...
+                                                }
+                                                .incrementalItems(
+                                                    incrementalDataVal.map{
+                                                        DeferPayload.newDeferredItem()
+                                                            .label(it["label"] as String)
+                                                            .path(it["path"] as List<Object>)
+                                                            .apply {
+                                                                if (it["extensions"] != null) extensions(it["extensions"] as Map<Any, Any>)
+                                                                if (it["errors"] != null) errors(it["errors"] as List<GraphQLError>)
+                                                            }
```
sbarker2, line 220: only label and path are required as per spec here:
https://github.com/graphql/graphql-spec/blob/c630301560d9819d33255d3ba00f548e8abbcdc6/spec/Section%207%20--%20Response.md?plain=1#L376C19-L377C53

gnawf:
Are we asserting the new `incrementalResponse` now or in another PR? Because that isn't happening here.

sbarker2:
> Are we asserting the new `incrementalResponse` now or in another PR? Because that isn't happening here.

the assertion will be in its own PR  👍

test/src/test/kotlin/graphql/nadel/tests/EngineTests.kt
```diff
@@ -180,7 +189,77 @@ private suspend fun execute(
...
+                                        |   service : '${serviceName}' 
+                                        |   service : '${serviceName}' 
+                                        |   query : '${actualQuery}' 
+                                        |   variables : '${actualVariables}' 
+                                        |   operation : '${actualOperationName}' 
+                                        """.trimMargin()
+                                        )
+                                    }
+                                    else if (serviceCall.incrementalResponse != null) {
+                                        fun transformData(executionResult: JsonMap): DelayedIncrementalPartialResult{
```
felipe-gdr, line 205: I think this function needs a more specific name: `transformDelayedIncrementalPartialResult` or `transformDelayed`...
I'd also move its definition out of the `getServiceExecution` function for better readability.
The other transformation steps could also be extracted into their own functions: `transformExecutionResult`, `transformIncrementalExecutionResult`, etc.

felipe-gdr (approved):

PR #525, opened 2024-03-18
Add timeout to coroutines - Draft

sbarker2:


lib/src/main/java/graphql/nadel/NextgenEngine.kt
```diff
@@ -101,14 +103,38 @@ internal class NextgenEngine(
...
-        return coroutineScope.async {
-            executeCoroutine(
-                executionInput,
-                queryDocument,
-                instrumentationState,
-                nadelExecutionParams.nadelExecutionHints,
-            )
-        }.asCompletableFuture()
+
+        val FF_APPLY_COROUTINE_TIMEOUT = true //TODO: make this legit
```
sbarker2, line 107: will add timeout value and FF to `NadelExecutionParams` in a bit

sbarker2: done

PR #528, opened 2024-03-25
Can specify max nodes for introspection

gnawf:


temaEmelyan (approved):

temaEmelyan (approved):

test/src/test/resources/fixtures/introspection/can-specify-max-introspection-response-size.yml
```diff
@@ -0,0 +1,52 @@
+name: "can specify max introspection response size"
```
felipe-gdr, line 1: What are we testing here?
Should we be using a query that exceeds the limit and assert an error message is returned?

gnawf: So that's something I need to follow up on…

In the GraphQL Java implementation once it hits the limit it just returns `null` without error.

https://github.com/graphql-java/graphql-java/pull/3525/files

felipe-gdr: I see. So this test case will actually result in max result nodes limits to be reached. Yeah, that's not super obvious from just looking at the test data. Maybe a comment to make it more clear?

PR #532, opened 2024-04-17
Switch to suspending functions all the way to the top

gnawf:


lib/src/main/java/graphql/nadel/Nadel.kt
```diff
@@ -48,7 +55,24 @@ class Nadel private constructor(
...
+        }
+    }
+
+    fun executeAsync(executionInput: NadelExecutionInput): CompletableFuture<ExecutionResult> {
+        return coroutineScope.future {
+            execute(executionInput)
+        }
+    }
+
+    @JvmName("executeSuspending")
```
felipe-gdr, line 74: Should this really be part of the API? I can't think of reasons why clients would call `execute` in a blocking way.

Also, is `executeSuspending` conveying the right meaning? What does "suspending" mean in Java code? Perhaps `executeBlocking` would be a better name.

lib/src/main/java/graphql/nadel/Nadel.kt
```diff
@@ -75,75 +99,79 @@ class Nadel private constructor(
...
-                        val abortException = throwable.cause as AbortExecutionException
-                        abortException.toExecutionResult()
-                    } else if (throwable is RuntimeException) {
-                        throw throwable
-                    } else {
-                        throw RuntimeException(throwable)
+            val result = try {
+                parseValidateAndExecute(executionInput, instrumentationState, hints)
+                    .also {
+                        executionInstrumentation.onCompleted(it, null)
```
felipe-gdr, line 105: `onCompleted` was being called from `whenComplete` before. Is calling it from `also` going to result in it executing right after the function returns but before the future resolves?

felipe-gdr: Also: do we have test coverage that asserts these instrumentation methods are being call as expected?

lib/src/main/java/graphql/nadel/Nadel.kt
gnawf: Well this isn't blocking, this is the suspending function in Kotlin. In a Kotlin codebase, this IS the entry point. The other functions are for Java compat.

Suspending functions can ONLY be invoked from Kotlin code. Java code can't invoke suspending functions.

1.  We had an `execute` function with the same name.
2. So we "renamed" that to `executeJava` in Kotlin, but Java code will see this as `execute` due to the `@JvmName` annotation.

I think the `@JvmName` is actually a bit redundant because the Kotlin compiler will change the function signature anyway, but I decided to keep it in to make it more explicit that we've kind of swapped these functions around.

Can explain this more at our sync.

lib/src/main/java/graphql/nadel/Nadel.kt
gnawf: So if I understand `whenComplete` then this is the exact same behavior.

`.also` isn't some special function. This code is exactly the same as
 
```
val temp1 = parseValidateAndExecute(executionInput, instrumentationState, hints)
executionInstrumentation.onCompleted(temp1, null)
temp1 // return
```

Here's the code definition for `.also`

```
public inline fun <T> T.also(block: (T) -> Unit): T {
    block(this)
    return this
}
```

gnawf: I'll check on the tests.

lib/src/main/java/graphql/nadel/Nadel.kt
felipe-gdr: Right - yeah. I was focused on the return type and missed the `suspend` keyword. Makes sense.
Should we get rid of the `@JvmName`, this should make the method essentially inaccessible from Java right, and just add some comments?

lib/src/main/java/graphql/nadel/Nadel.kt
felipe-gdr: Ah, I see. So calling a suspended function will actually block the execution. 
Whereas this
```
async {
parseValidateAndExecute(executionInput, instrumentationState, hints)
}
```
would return a `Defer` which is kind of like a CF in Java

felipe-gdr (approved):

lib/src/main/java/graphql/nadel/Nadel.kt
gnawf: Ok I introduced some more tests here https://github.com/atlassian-labs/nadel/pull/539

PR #533, opened 2024-05-01
Defer hydrations

gnawf:


lib/src/main/java/graphql/nadel/engine/NadelDeferSupport.kt
```diff
@@ -0,0 +1,167 @@
...
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.channels.Channel.Factory.UNLIMITED
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.consumeAsFlow
+import kotlinx.coroutines.launch
+import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.atomic.AtomicInteger
+
+class NadelDeferSupport internal constructor(
+    private val delayedResultsChannel: Channel<DelayedIncrementalPartialResult> = Channel(UNLIMITED),
```
felipe-gdr, line 20: What does `UNLIMITED` mean here? Is it the maximum size of the buffer? Should we adopt a reasonable limit instead?

lib/src/test/kotlin/graphql/nadel/engine/NadelDeferSupportTest.kt
```diff
@@ -0,0 +1,208 @@
...
+import kotlinx.coroutines.flow.lastOrNull
+import kotlinx.coroutines.flow.toList
+import kotlinx.coroutines.test.runTest
+import org.junit.jupiter.api.Test
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+class NadelDeferSupportTest {
+    @Test
+    fun `channel closes once initial result comes in and there are no pending defer jobs`() {
```
felipe-gdr, line 19: Is this description accurate? Wouldn't the channel close only once the defer coroutine scope is complete, which will happen when all pending defer calls are finished?
This description implies that nested defers are not supported, is that the case?

lib/src/test/kotlin/graphql/nadel/engine/NadelDeferSupportTest.kt
```diff
@@ -0,0 +1,208 @@
...
+        // Then
+        firstAsync.join()
+        secondAsync.join()
+
+        lockingJob.complete(true)
+
+        val results = channel.consumeAsFlow().toList()
+        val lastResult = results.last()
+        assertTrue((lastResult.incremental?.single() as DeferPayload).getData<String>() == "Bye world")
+        assertFalse(lastResult.hasNext())
```
felipe-gdr, line 83: can we assert that `hasNext` on the first payload is `true`, just in case?

lib/src/test/kotlin/graphql/nadel/engine/NadelDeferSupportTest.kt
```diff
@@ -0,0 +1,208 @@
...
+        val subject = NadelDeferSupport(channel)
+        val firstLock = CompletableDeferred<Boolean>()
+        val secondLock = CompletableDeferred<Boolean>()
+
+        // When
+        subject.defer {
+            // Wait until test tells us to continue
+            firstLock.join()
+
+            subject.defer {
```
felipe-gdr, line 98: oh, so we do support nested defers. Cool.

lib/src/test/kotlin/graphql/nadel/engine/NadelDeferSupportTest.kt
```diff
@@ -0,0 +1,208 @@
...
+        // Then
+        firstLock.complete(true)
+
+        val item = channel.receive()
+        assertTrue(item.hasNext())
+    }
+
+    @Test
+    fun `hasNext is true if there is another job still running`() = runTest {
+        // Channel that stores the oldest item
```
felipe-gdr, line 122: I can't understand what this comment means 🤔 There's only one channel where all the items are published to, right?

lib/src/main/java/graphql/nadel/engine/NadelDeferSupport.kt
```diff
@@ -0,0 +1,167 @@
...
+            )
+        }
+    }
+
+    /**
+     * Assumes that the [Flow] will correctly set [DelayedIncrementalPartialResult.hasNext] correctly.
+     *
+     * i.e. that the last element in the [Flow] will have `hasNext` set to `false`.
+     */
+    fun defer(serviceResults: Flow<DelayedIncrementalPartialResult>): Job {
```
felipe-gdr, line 63: Is this version of the `defer` function supposed to be used in the scenario where the underlying services return delayed responses? 
cc @sbarker2

lib/src/main/java/graphql/nadel/engine/NadelDeferSupport.kt
```diff
@@ -0,0 +1,167 @@
...
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.channels.Channel.Factory.UNLIMITED
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.consumeAsFlow
+import kotlinx.coroutines.launch
+import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.atomic.AtomicInteger
+
+class NadelDeferSupport internal constructor(
```
felipe-gdr, line 19: This class will also be used for `@stream` support. We should name all constructs shared by `@stream` and `@defer` with a common name or prefix. Maybe `Delayed` or `Incremental`?

lib/src/main/java/graphql/nadel/engine/NadelDeferSupport.kt
gnawf: Yeah it's the size of the buffer, we can drop it down.

lib/src/test/kotlin/graphql/nadel/engine/NadelDeferSupportTest.kt
gnawf: So yeah as you discovered later on in the test, nested defer is supported.

This particular description is accurate. Defer calls are meant to be made before the execution result finishes and goes back to the user.

lib/src/test/kotlin/graphql/nadel/engine/NadelDeferSupportTest.kt
gnawf: Done

lib/src/test/kotlin/graphql/nadel/engine/NadelDeferSupportTest.kt
gnawf: Oh yeah that comment seems outdated, I'll remove it. I think before I had `Channel(1, DROP_OLDEST)` or something, but I changed it.

lib/src/main/java/graphql/nadel/engine/NadelDeferSupport.kt
gnawf: Yep I've renamed it, thanks.

lib/src/test/kotlin/graphql/nadel/engine/NadelDeferSupportTest.kt
gnawf: i.e. if no defer jobs are pending by the time the first execution job is done, then we can conclude there is no defer. So we close the channel as no other results are coming in.

Also, you effectively cannot submit a defer job _after_ the initial response is done (unless it's nested) because you'd have to be submitting it async to the main coroutine scope of the request.

felipe-gdr: Ah yep, makes sense. Thanks for explaining 👍

lib/src/main/java/graphql/nadel/engine/NadelDeferSupport.kt
gnawf: Dropped it down to 100 which should be much more than we need, and yet still a sane limit.

lib/src/main/java/graphql/nadel/Nadel.kt
```diff
@@ -259,11 +260,17 @@ class Nadel private constructor(
         private var transforms = emptyList<NadelTransform<out Any>>()
         private var introspectionRunnerFactory = NadelIntrospectionRunnerFactory(::NadelDefaultIntrospectionRunner)
 
+        private var schemas: NadelSchemas? = null
         private var schemaBuilder = NadelSchemas.Builder()
 
         private var maxQueryDepth = Integer.MAX_VALUE
         private var maxFieldCount = Integer.MAX_VALUE
 
+        fun schemas(schemas: NadelSchemas): Builder {
```
gnawf, line 269: So that our tests can build the schemas separately.

Tbh we probably should remove all the `overallSchema` methods etc. and make the consumer build a `NadelSchemas` themselves and pass it in.

lib/src/main/java/graphql/nadel/NextgenEngine.kt
```diff
@@ -117,14 +120,17 @@ internal class NextgenEngine(
                 )
             }
 
+            val incrementalResultSupport = NadelIncrementalResultSupport()
```
gnawf, line 123: New defer support class that gets stored in `NadelExecutionContext` for transforms etc to access.

lib/src/main/java/graphql/nadel/NextgenEngine.kt
```diff
@@ -167,7 +173,16 @@ internal class NextgenEngine(
             }
 
             beginExecuteContext?.onCompleted(result, null)
-            return result
+            incrementalResultSupport.onInitialResultComplete()
+
+            return if (incrementalResultSupport.hasDeferredResults()) {
+                IncrementalExecutionResultImpl.Builder()
+                    .from(result)
+                    .incrementalItemPublisher(incrementalResultSupport.resultFlow().asPublisher())
```
gnawf, line 181: FYI the `Publisher` can only be consumed once.

lib/src/main/java/graphql/nadel/NextgenEngine.kt
```diff
@@ -176,11 +191,25 @@ internal class NextgenEngine(
         }
     }
 
-    internal suspend fun executeTopLevelField(
+    internal suspend fun executeHydration(
```
gnawf, line 194: Split up the functions.

lib/src/main/java/graphql/nadel/engine/NadelExecutionContext.kt
```diff
@@ -19,6 +20,8 @@ data class NadelExecutionContext internal constructor(
     val hints: NadelExecutionHints,
     val instrumentationState: InstrumentationState?,
     internal val timer: NadelInstrumentationTimer,
+    internal val deferSupport: NadelIncrementalResultSupport,
+    internal val hydrationDetails: ServiceExecutionHydrationDetails? = null,
```
gnawf, line 24: Stores hydration details so that transforms can access it.

In the future this will enable us to handle deferring nested hydrations.

lib/src/main/java/graphql/nadel/engine/NadelIncrementalResultSupport.kt
```diff
@@ -0,0 +1,181 @@
...
+ * todo: we do not handle the case where defer jobs finish before [onInitialResultComplete]
+ */
+class NadelIncrementalResultSupport internal constructor(
+    private val delayedResultsChannel: Channel<DelayedIncrementalPartialResult> = Channel(
+        capacity = 100,
+        onBufferOverflow = BufferOverflow.DROP_LATEST,
+        onUndeliveredElement = {
+            log.error("Dropping incremental result because of buffer overflow")
+        },
+    ),
```
gnawf, line 30: Default channel buffers 100 elements if nobody is consuming them.

Should be high enough.

lib/src/main/java/graphql/nadel/engine/transform/hydration/NadelHydrationTransform.kt
```diff
@@ -296,4 +393,73 @@ internal class NadelHydrationTransform(
...
+
+        return if (executionContext.hints.deferSupport() && overallField.deferredExecutions.isNotEmpty()) {
+            // We currently don't support defer if the hydration is inside a List
+            return !areAnyParentFieldsOutputtingLists(overallField, executionBlueprint)
+        } else {
+            false
+        }
+    }
+
+    private fun areAnyParentFieldsOutputtingLists(
```
gnawf, line 407: In _this_ PR we don't allow deferring in lists. It's been implemented in https://github.com/atlassian-labs/nadel/pull/535

lib/src/main/java/graphql/nadel/engine/transform/hydration/NadelHydrationTransform.kt
```diff
@@ -187,88 +273,99 @@ internal class NadelHydrationTransform(
...
+            fieldToHydrate = fieldToHydrate,
+            parentNode = parentNode,
+            executionBlueprint = executionBlueprint,
+        )
 
-        when (instruction.hydrationStrategy) {
-            is NadelHydrationStrategy.OneToOne -> {
-                // Should not have more than one query for one to one
-                val result = actorQueryResults.emptyOrSingle()
+        return NadelPreparedHydration {
```
gnawf, line 292: Code inside here should be the same, it's just wrapped in a callback now, instead of being executed immediately.

lib/src/main/java/graphql/nadel/engine/transform/hydration/NadelHydrationTransform.kt
```diff
@@ -154,30 +166,104 @@ internal class NadelHydrationTransform(
...
+        parentNodes: List<JsonNode>,
+        state: State,
+        executionBlueprint: NadelOverallExecutionBlueprint,
+        overallField: ExecutableNormalizedField,
+        executionContext: NadelExecutionContext,
+    ): List<NadelResultInstruction> {
+        return coroutineScope {
+            parentNodes
+                .map {
+                    prepareHydration(
```
gnawf, line 179: So the way this works is we now split the hydration into two steps

1. Preparing the hydration
2. Executing the prepared hydration

The first step involves gathering things like the source inputs etc. This is done before the Nadel execution framework removes all the artificial fields etc.

Then for non-deferred hydration we execute step 2 in sync. Otherwise if it's deferred, we execute step 2 in a background coroutine and then build a delayed result and send it back to the user.

lib/src/main/java/graphql/nadel/engine/transform/hydration/NadelHydrationTransform.kt
```diff
@@ -154,30 +166,104 @@ internal class NadelHydrationTransform(
...
+        // Prepare the hydrations before we go async
+        // We need to do this because if we run it async below, we cannot guarantee that our artificial fields have not yet been removed
+        val hydrations = parentNodes.map {
+            prepareHydration(
+                parentNode = it,
+                state = state,
+                executionBlueprint = executionBlueprint,
+                fieldToHydrate = overallField,
+                executionContext = executionContext,
+            )
```
gnawf, line 213: Here we prepare the hydrations.

lib/src/main/java/graphql/nadel/engine/transform/hydration/NadelHydrationTransform.kt
```diff
@@ -154,30 +166,104 @@ internal class NadelHydrationTransform(
...
         }
 
-        return jobs.awaitAll().flatten()
+        executionContext.deferSupport.defer {
+            val instructionSequence = hydrations
+                .map {
+                    async {
+                        it.hydrate()
```
gnawf, line 220: Then we execute them in the background coroutine by using `deferSupport.defer`

lib/src/main/java/graphql/nadel/engine/transform/hydration/NadelHydrationTransform.kt
```diff
@@ -187,88 +273,99 @@ internal class NadelHydrationTransform(
 
         // Do nothing if there is no hydration instruction associated with this result
         if (instructions.isEmpty()) {
-            return emptyList()
+            return NadelPreparedHydration {
```
gnawf, line 268: This code should mainly be the same. All the code as just be wrapped in a `NadelPreparedHydration`

lib/src/main/java/graphql/nadel/engine/NadelDeferSupport.kt
gnawf: Yep we're using it in https://github.com/atlassian-labs/nadel/pull/540/files#diff-90d64a63cac8ddb0ce02278176c2fbc733244bdecab27b930281841f66d2eb7fR325-R329 now

lib/src/main/java/graphql/nadel/engine/NadelExecutionContext.kt
```diff
@@ -19,9 +20,15 @@ data class NadelExecutionContext internal constructor(
     val hints: NadelExecutionHints,
     val instrumentationState: InstrumentationState?,
     internal val timer: NadelInstrumentationTimer,
+    internal val incrementalResultSupport: NadelIncrementalResultSupport,
+    internal val hydrationDetails: ServiceExecutionHydrationDetails? = null,
 ) {
     private val serviceContexts = ConcurrentHashMap<String, CompletableFuture<Any?>>()
 
+    @Deprecated("Use incrementalSupport instead", ReplaceWith("incrementalResultSupport"))
+    internal val deferSupport: NadelIncrementalResultSupport
```
gnawf, line 29: I renamed it, but this is just so other branches can still use the code.

I think Steven's branch relies on it. I'll delete this when we merge those PRs.

lib/src/main/java/graphql/nadel/engine/NadelIncrementalResultSupport.kt
```diff
@@ -0,0 +1,181 @@
...
+
+            delayedResultsChannel.send(
+                // Copy of result but with the correct hasNext according to the info we know
+                result.copy(hasNext = hasNext)
+            )
+        }
+    }
+
+    /**
+     * Assumes that the [Flow] will correctly set [DelayedIncrementalPartialResult.hasNext] correctly.
```
felipe-gdr-atlassian, line 73: @sbarker2 @Juliano-Prado This is the method that the "non-hydrated" defer code path will use. You don't have to worry about what the value of `hasNext` is, as this should have been set correctly by the underlying service, though.

lib/src/main/java/graphql/nadel/engine/NadelIncrementalResultSupport.kt
```diff
@@ -0,0 +1,181 @@
...
+                    // Here we'll stipulate that the last element of the Flow sets hasNext=false
+                    val hasNext = if (result.hasNext()) {
+                        true
+                    } else {
+                        outstandingJobHandle.decrementAndGetJobCount() > 0
+                    }
+
+                    delayedResultsChannel.send(
+                        // Copy of result but with the correct hasNext according to the info we know
+                        result.copy(hasNext = hasNext)
```
felipe-gdr-atlassian, line 90: A significant amount of the time, the value of `hasNext` won't have changed. Is it worth checking it before calling `copy`, or is its implementation optimized for this scenario (i.e. nothing changed)?

sbarker2 (approved):

lib/src/test/kotlin/graphql/nadel/engine/NadelIncrementalResultSupportTest.kt
```diff
@@ -0,0 +1,214 @@
...
+                // Must be at end of runTest
+                testCompleted = true
+            }
+        } catch (e: UnsupportedOperationException) {
+            assertTrue(e.message == "Hello")
+        }
+
+        // Coroutines code will throw uncaught exceptions even though we threw it deliberately
+        // Just ensure that we actually ran all the asserts and the tests is fine
+        assertTrue(testCompleted)
```
felipe-gdr-atlassian, line 212: Amazing test coverage @gnawf . Can we add some tests for the `Flow` version of the `defer` function though?

felipe-gdr-atlassian (approved):

felipe-gdr-atlassian (changes requested):
great stuff! Just marking as "request changes" so we add tests for the Flow version of the defer function 👍

felipe-gdr-atlassian (changes requested):
great stuff! Just marking as "request changes" so we add tests for the Flow version of the defer function 👍

gnawf:
@felipe-gdr-atlassian good point, I'll update it with more tests. I wrote the `Flow` function after I wrote the initial class & tests as something for Steven to piggy back off.

lib/src/main/java/graphql/nadel/engine/NadelIncrementalResultSupport.kt
felipe-gdr: Looking at the implementation of `copy` in `graphql.nadel.engine.util.GraphQLUtilKt#copy(graphql.incremental.DelayedIncrementalPartialResult, java.util.List<? extends graphql.incremental.IncrementalPayload>, java.util.Map<java.lang.Object,? extends java.lang.Object>, boolean)`
it will create a new object every time. Is it worth preventing this if `hasNext` hasn't changed?

felipe-gdr (approved):

lib/src/main/java/graphql/nadel/engine/NadelIncrementalResultSupport.kt
gnawf: Hey I'll address this in another PR :)

PR #540, opened 2024-05-07
Merge deferred results

sbarker2:
Merging deferred results to publisher and send back to client (part of the defer implementation for non-hydrated queries)

NOTE: This does not apply the transformers yet - this work is in progress, but just note at this point, queries with transformers (for example renames) will not work.


lib/src/main/java/graphql/nadel/engine/NadelExecutionContext.kt
```diff
@@ -25,10 +25,6 @@ data class NadelExecutionContext internal constructor(
 ) {
     private val serviceContexts = ConcurrentHashMap<String, CompletableFuture<Any?>>()
 
-    @Deprecated("Use incrementalSupport instead", ReplaceWith("incrementalResultSupport"))
-    internal val deferSupport: NadelIncrementalResultSupport
-        get() = incrementalResultSupport
-
```
sbarker2, line 31: Pretty sure I was the only one using this so we can remove the deprecated field now

test/src/test/kotlin/graphql/nadel/tests/next/fixtures/defer/DeferWithLabelTestSnapshot.kt
```diff
@@ -0,0 +1,102 @@
...
+                |       slow
+                |     }
+                |   }
+                | }
+                """.trimMargin(),
+                variables = "{}",
+                response = """
+                | {
+                |   "defer": {
+                |     "hello": "helloString"
```
felipe-gdr, line 36: The initial response should contain the `hasNext` field with value `true`. See example here:
https://github.com/graphql/graphql-over-http/blob/main/rfcs/IncrementalDelivery.md#content-type-multipartmixed

test/src/test/kotlin/graphql/nadel/tests/next/fixtures/defer/DeferWithLabelTestSnapshot.kt
```diff
@@ -0,0 +1,102 @@
...
+     */
+    override val response: ExpectedNadelResponse = ExpectedNadelResponse(
+            response = """
+            | {
+            |   "data": {
+            |     "defer": {
+            |       "hello": "helloString"
+            |     }
+            |   },
+            |   "hasNext": true
```
felipe-gdr, line 80: Not sure why `hasNext` is present on the final expected response, but not on the service response.

test/src/test/kotlin/graphql/nadel/tests/next/fixtures/defer/MultipleDeferDirectivesTestSnapshot.kt
```diff
@@ -0,0 +1,133 @@
...
+                variables = "{}",
+                response = """
+                | {
+                |   "defer": {
+                |     "fastField": "123"
+                |   }
+                | }
+                """.trimMargin(),
+                delayedResponses = listOfJsonStrings(
+                    """
```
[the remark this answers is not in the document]
felipe-gdr: also, how deterministic is this? Not sure that `slowString` will always be the last field to be resolved here.

test/src/test/kotlin/graphql/nadel/tests/next/fixtures/defer/MultipleFieldsinSingleDeferDirectiveTestSnapshot.kt
```diff
@@ -0,0 +1,104 @@
...
+                |       "data": {
+                |         "slowField": "slowString",
+                |         "anotherSlowField": 123456789
+                |       }
+                |     }
+                |   ]
+                | }
+                """.trimMargin(),
+            ),
+        )
```
felipe-gdr, line 103: Some more test cases you could add:
- defer on a field from a type returned as part of a list
- defer on calls to multiple underlying services 
- error cases - what if the initial response results on an error. What if one or more incremental response results in an error?
- extensions - on initial payload and incremental payloads

felipe-gdr (changes requested):
Great stuff - I marked as "requested changes" because of the missing `hasNext` on service call response and so we can add more test scenarios.

test/src/test/kotlin/graphql/nadel/tests/next/fixtures/defer/DeferWithLabelTestSnapshot.kt
gnawf: Hmm that's probably my fault. This is odd, it doesn't even contain the `data` field.

PR #541, opened 2024-05-14
put shared renamed types in unions behind a hint

temaEmelyan:

that's a follow up on https://github.com/atlassian-labs/nadel/pull/537
I put all changes that were done there behind a hint so we can have a safer rollout
also in that PR I missed one branch, fixed that and covered with a test

### DIFF
--- a/lib/src/main/java/graphql/nadel/Service.kt
+++ b/lib/src/main/java/graphql/nadel/Service.kt
@@ -17,4 +17,9 @@ open class Service(
      * These are the GraphQL definitions that a service contributes to the OVERALL schema.
      */
     val definitionRegistry: NadelDefinitionRegistry,
-)
+
+    ) {
+    override fun toString(): String {
+        return "Service{name='$name'}"
+    }
+}
--- a/lib/src/main/java/graphql/nadel/engine/transform/NadelTypeRenameResultTransform.kt
+++ b/lib/src/main/java/graphql/nadel/engine/transform/NadelTypeRenameResultTransform.kt
@@ -77,16 +77,17 @@ internal class NadelTypeRenameResultTransform : NadelTransform<State> {
                 underlyingTypeName = underlyingTypeName,
             )
 
-            val typeName: String = if (
-                !executionContext.hints.sharedTypeRenames(service) ||
-                overallField.objectTypeNames.contains(overallTypeName)
-            ) {
-                overallTypeName
-            } else {
-                overallField.objectTypeNames.singleOrNull {
-                    executionBlueprint.getRename(it)?.underlyingName == underlyingTypeName
+            val typeName: String = if (executionContext.hints.sharedTypeRenames(service)) {
+                if (overallField.objectTypeNames.contains(overallTypeName)) {
+                    overallTypeName
+                } else {
+                    overallField.objectTypeNames.singleOrNull {
+                        executionBlueprint.getRename(it)?.underlyingName == underlyingTypeName
+                    } ?: overallTypeName
                 }
-            } ?: overallTypeName
+            } else {
+                overallTypeName
+            }
             NadelResultInstruction.Set(
                 subject = parentNode,
                 key = NadelResultKey(overallField.resultKey),
--- a/lib/src/main/java/graphql/nadel/engine/transform/NadelTypeRenameResultTransform.kt
+++ b/lib/src/main/java/graphql/nadel/engine/transform/NadelTypeRenameResultTransform.kt
@@ -77,13 +77,16 @@ internal class NadelTypeRenameResultTransform : NadelTransform<State> {
                 underlyingTypeName = underlyingTypeName,
             )
 
-            val typeName: String = if (overallField.objectTypeNames.contains(overallTypeName)) {
+            val typeName: String = if (
+                !executionContext.hints.sharedTypeRenames(service) ||
+                overallField.objectTypeNames.contains(overallTypeName)
+            ) {
                 overallTypeName
             } else {
-                overallField.objectTypeNames.single {
+                overallField.objectTypeNames.singleOrNull {
                     executionBlueprint.getRename(it)?.underlyingName == underlyingTypeName
                 }
-            }
+            } ?: overallTypeName
             NadelResultInstruction.Set(
                 subject = parentNode,
                 key = NadelResultKey(overallField.resultKey),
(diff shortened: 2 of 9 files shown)


lib/src/main/java/graphql/nadel/Service.kt
```diff
@@ -17,4 +17,9 @@ open class Service(
      * These are the GraphQL definitions that a service contributes to the OVERALL schema.
      */
     val definitionRegistry: NadelDefinitionRegistry,
-)
+
+    ) {
+    override fun toString(): String {
+        return "Service{name='$name'}"
```
temaEmelyan, line 23: this is just so we don't see stuff like `Service@323edf`

lib/src/main/java/graphql/nadel/engine/transform/NadelTypeRenameResultTransform.kt
```diff
@@ -77,13 +77,16 @@ internal class NadelTypeRenameResultTransform : NadelTransform<State> {
                 underlyingTypeName = underlyingTypeName,
             )
 
-            val typeName: String = if (overallField.objectTypeNames.contains(overallTypeName)) {
+            val typeName: String = if (
+                !executionContext.hints.sharedTypeRenames(service) ||
+                overallField.objectTypeNames.contains(overallTypeName)
+            ) {
```
temaEmelyan, line 83: so, if hints are off, we will always use the old behaviour, just get the `overallTypeName`

lib/src/main/java/graphql/nadel/engine/transform/NadelTypeRenameResultTransform.kt
```diff
@@ -77,13 +77,16 @@ internal class NadelTypeRenameResultTransform : NadelTransform<State> {
...
                 overallTypeName
             } else {
-                overallField.objectTypeNames.single {
```
temaEmelyan, line 83: that `single` blew up here during the rollout, handling it safer by defaulting to `overallTypeName`

lib/src/main/java/graphql/nadel/engine/transform/query/NadelQueryTransformer.kt
```diff
@@ -168,8 +168,14 @@ class NadelQueryTransformer private constructor(
...
     private fun getUnderlyingTypeNames(objectTypeNames: Collection<String>): List<String> {
-        return objectTypeNames.map {
-            executionBlueprint.getUnderlyingTypeName(service, overallTypeName = it)
+        return if (executionContext.hints.sharedTypeRenames(service)) {
+            objectTypeNames.map {
+                executionBlueprint.getUnderlyingTypeName(overallTypeName = it)
+            }
+        } else {
+            objectTypeNames.map {
+                executionBlueprint.getUnderlyingTypeName(service, overallTypeName = it)
```
temaEmelyan, line 177: ok, so if something is renamed but the type is declared and renamed in another service
`executionBlueprint.getUnderlyingTypeName(service, overallTypeName = it)` will not actually get you the underlying name as it's only looking at the types owned by `service`. this thing `executionBlueprint.getUnderlyingTypeName(overallTypeName = it)` looks at the whole overall schema tho, for every rename, there can only be 1 underlying name and 1 overall name, so it's always 1 to 1 mapping

lib/src/main/java/graphql/nadel/engine/transform/NadelTypeRenameResultTransform.kt
```diff
@@ -77,13 +77,16 @@ internal class NadelTypeRenameResultTransform : NadelTransform<State> {
                 underlyingTypeName = underlyingTypeName,
             )
 
-            val typeName: String = if (overallField.objectTypeNames.contains(overallTypeName)) {
+            val typeName: String = if (
+                !executionContext.hints.sharedTypeRenames(service) ||
```
gnawf, line 81: Sorry idk if my brain is fried but I've been staring at this if statement for a 5 mins trying to figure it out and I can't lol.

Can we change it to something clearer like

```
val typeName: String = if (executionContext.hints.sharedTypeRenames(service)) {
    if (overallField.objectTypeNames.contains(overallTypeName)) {
        overallTypeName
    } else {
        overallField.objectTypeNames.singleOrNull {
            executionBlueprint.getRename(it)?.underlyingName == underlyingTypeName
        } ?: overallTypeName
    }
} else {
    overallTypeName
}
```

lib/src/main/java/graphql/nadel/engine/transform/query/NadelQueryTransformer.kt
```diff
@@ -168,8 +168,14 @@ class NadelQueryTransformer private constructor(
     }
 
     private fun getUnderlyingTypeNames(objectTypeNames: Collection<String>): List<String> {
-        return objectTypeNames.map {
-            executionBlueprint.getUnderlyingTypeName(service, overallTypeName = it)
+        return if (executionContext.hints.sharedTypeRenames(service)) {
+            objectTypeNames.map {
+                executionBlueprint.getUnderlyingTypeName(overallTypeName = it)
```
gnawf, line 173: This… doesn't feel right.

Sorry I haven't really been following this work, but renames are technically supposed to be per service.

Tbh I don't really remember the full story, but a long time ago I wanted to implement repeated `@renamed` with a `service` argument.

lib/src/main/java/graphql/nadel/engine/transform/query/NadelQueryTransformer.kt
```diff
@@ -168,8 +168,14 @@ class NadelQueryTransformer private constructor(
     }
 
     private fun getUnderlyingTypeNames(objectTypeNames: Collection<String>): List<String> {
-        return objectTypeNames.map {
-            executionBlueprint.getUnderlyingTypeName(service, overallTypeName = it)
+        return if (executionContext.hints.sharedTypeRenames(service)) {
```
temaEmelyan, line 171: this is the bit I missed the first time around. This is covered by the test where I added fragments that used renamed typenames
```graphql
  query {
    nodes {
      __typename
      ... on Issue {
        id
      }
      ... on JiraComment {
        id
      }
    }
  }
```
I didn't realise that this thing runs every time before we send the query to the underlying

lib/src/main/java/graphql/nadel/engine/transform/NadelTypeRenameResultTransform.kt
temaEmelyan: thanks, done

gnawf (approved):

PR #562, opened 2024-07-09
Defer payloads in groups

gnawf:
Early work

### DIFF
--- a/lib/src/main/java/graphql/nadel/engine/NadelIncrementalResultAccumulator.kt
+++ b/lib/src/main/java/graphql/nadel/engine/NadelIncrementalResultAccumulator.kt
@@ -160,7 +160,7 @@ class NadelIncrementalResultAccumulator(
 /**
  * Similar to [java.io.File.walkTopDown] but for ENFs.
  */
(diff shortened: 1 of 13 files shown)


lib/src/main/java/graphql/nadel/engine/NadelIncrementalResultAccumulator.kt
```diff
@@ -0,0 +1,179 @@
...
+import graphql.incremental.DeferPayload
+import graphql.incremental.DelayedIncrementalPartialResult
+import graphql.incremental.DelayedIncrementalPartialResultImpl
+import graphql.nadel.engine.transform.query.NadelQueryPath
+import graphql.nadel.engine.util.MutableJsonMap
+import graphql.normalized.ExecutableNormalizedField
+import graphql.normalized.ExecutableNormalizedOperation
+import graphql.normalized.incremental.NormalizedDeferredExecution
+
+class NadelIncrementalResultAccumulator(
```
felipe-gdr, line 13: Can you add some docs on this class that briefly describes the problem that it is solving?

gnawf: Done

felipe-gdr (approved):

PR #565, opened 2024-07-24
Add support for defer payload label

gnawf:


felipe-gdr (approved):

lib/src/main/java/graphql/nadel/engine/transform/hydration/NadelHydrationTransform.kt
```diff
@@ -213,6 +213,9 @@ internal class NadelHydrationTransform(
             )
         }
 
+        // This isn't really right… but we start with this
```
Juliano-Prado, line 216: it would be great to have better details of current issues/limitation

Juliano-Prado (approved):

lib/src/main/java/graphql/nadel/engine/transform/hydration/NadelHydrationTransform.kt
gnawf: Ah sorry yeah I forgot we discussed this, will fix in subsequent PR

PR #572, opened 2024-08-13
Fix top level hydration

gnawf:
So the bug was that we couldn't find the service that owned a hydrated field, because we looked at the ownership of the parent type of the hydrated field, which for root level hydrations was the operation type.

Then on top of that, we didn't handle multple root level fields.

This PR fixes both.

### DIFF
--- a/lib/src/main/java/graphql/nadel/NextgenEngine.kt
+++ b/lib/src/main/java/graphql/nadel/NextgenEngine.kt
@@ -29,7 +29,6 @@ import graphql.nadel.engine.transform.result.NadelResultTransformer
 import graphql.nadel.engine.util.MutableJsonMap
 import graphql.nadel.engine.util.beginExecute
 import graphql.nadel.engine.util.compileToDocument
-import graphql.nadel.engine.util.copy
 import graphql.nadel.engine.util.getOperationKind
 import graphql.nadel.engine.util.newExecutionResult
 import graphql.nadel.engine.util.newGraphQLError
@@ -64,7 +63,6 @@ import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.delay
-import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.future.asCompletableFuture
 import kotlinx.coroutines.future.asDeferred
@@ -298,11 +296,11 @@ internal class NextgenEngine(
             executionContext.incrementalResultSupport.defer(
                 result.incrementalItemPublisher
                     .asFlow()
-                    .onEach {delayedIncrementalResult ->
+                    .onEach { delayedIncrementalResult ->
                         // Transform
                         delayedIncrementalResult.incremental
                             ?.filterIsInstance<DeferPayload>()
-                            ?.forEach {deferPayload ->
+                            ?.forEach { deferPayload ->
                                 resultTransformer
                                     .transform(
                                         executionContext = executionContext,
@@ -313,7 +311,8 @@ internal class NextgenEngine(
                                         service = service,
                                         result = result,
                                         deferPayload = deferPayload,
-                                    ) }
+                                    )
+                            }
                     }
             )
         }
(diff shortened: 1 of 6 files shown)


lib/src/main/java/graphql/nadel/NextgenEngine.kt
```diff
@@ -343,9 +343,16 @@ internal class NextgenEngine(
...
             serviceExecutionContext = serviceExecutionContext,
             hydrationDetails = executionHydrationDetails,
-            executableNormalizedField = transformedQuery,
+            // Prefer non __typename field first, otherwise we just get first
+            executableNormalizedField = topLevelFields
+                .asSequence()
+                .filterNot {
+                    it.fieldName == TypeNameMetaFieldDef.name
+                }
+                .firstOrNull() ?: topLevelFields.first(),
```
gnawf, line 352: This gets the first non `__typename` field, to maintain the old behavior.

(Refer to above comment).

lib/src/main/java/graphql/nadel/NextgenEngine.kt
```diff
@@ -343,9 +343,16 @@ internal class NextgenEngine(
             serviceContext = executionContext.getContextForService(service).await(),
             serviceExecutionContext = serviceExecutionContext,
             hydrationDetails = executionHydrationDetails,
-            executableNormalizedField = transformedQuery,
+            // Prefer non __typename field first, otherwise we just get first
+            executableNormalizedField = topLevelFields
```
gnawf, line 347: So this `ServiceExecutionParameters.executableNormalizedField` actually needs to be deprecated and removed but we are keeping it for now for compatibility.

It's currently used to extract the cloud ID for routing.

The reason why we went from `ExecutableNormalizedField` to `List<ExecutableNormalizedField>` is because the root level hydration transform injects a `__typename` field.

```graphql
{
  batch_hydration__myIssues__myIssueKeys: myIssueKeys
  __typename__batch_hydration__myIssues: __typename
}
```

You could argue that we don't need a `__typename` for the operation types, but it's not a big deal to just keep it in.

Plus in the future we may end up grouping more root level fields into one execution, so this sort of lays the ground work for that.

lib/src/main/java/graphql/nadel/NextgenEngine.kt
```diff
@@ -388,35 +395,45 @@ internal class NextgenEngine(
...
+        if (hints.shortCircuitEmptyQuery(service) && isOnlyTopLevelFieldTypename(topLevelFields)) {
+            return engineSchemaIntrospectionService.serviceExecution
         }
+
+        return service.serviceExecution
     }
 
-    private fun onlyTopLevelTypenameField(executableNormalizedField: ExecutableNormalizedField): Boolean {
-        if (executableNormalizedField.fieldName == TypeNameMetaFieldDef.name) {
+    private fun isOnlyTopLevelFieldTypename(topLevelFields: List<ExecutableNormalizedField>): Boolean {
```
gnawf, line 421: This code here was modified a bit.

Context: we wanted to add code that would shortcut query execution IF there was only one field and it was `__typename`

Note: this code is protected by a FF `hints.shortCircuitEmptyQuery(service)` and it's mostly off because it hasn't been fully rolled out.

lib/src/main/java/graphql/nadel/engine/blueprint/NadelExecutionBlueprintFactory.kt
```diff
@@ -561,11 +560,25 @@ private class Factory(
...
         val renameInstruction = makeTypeRenameInstruction(overallType as? GraphQLDirectiveContainer ?: return null)
-        val service = definitionNamesToService[overallType.name]
-            ?: error("Unknown service for type: ${overallType.name}")
         val underlyingName = renameInstruction?.underlyingName ?: overallType.name
+
+        val fieldCoordinates = makeFieldCoordinates(overallType, childField)
+
+        val service = definitionNamesToService[overallType.name]
+            ?: coordinatesToService[fieldCoordinates]
+            ?: error("Unable to determine service for $fieldCoordinates")
```
gnawf, line 580: Main fix is here.
Before we would do type name -> service.
This doesn't work for operation types.
So we now fall back to field name -> service.

andimarek-atlassian (dismissed):

lib/src/main/java/graphql/nadel/NextgenEngine.kt
```diff
@@ -408,39 +414,49 @@ internal class NextgenEngine(
             )
         }
 
-        val transformedData: MutableJsonMap = serviceExecResult.data.let { data ->
-            data.takeIf { transformedQuery.resultKey in data }
-                ?: mutableMapOf(transformedQuery.resultKey to null)
-        }
+        val transformedData: MutableJsonMap = serviceExecResult.data
```
russellyou, line 417: nit, why can't we use immutable approach like

serviceExecResult.data.filter((k,v) -> topLevelFields.keySet.contain(k)).toMutableMap

russellyou (dismissed):

russellyou (approved):

PR #579, opened 2024-08-19
Apply transforms to defer payloads

sbarker2:


lib/src/main/java/graphql/nadel/NextgenEngine.kt
```diff
@@ -383,7 +390,23 @@ internal class NextgenEngine(
 
         if (serviceExecResult is NadelIncrementalServiceExecutionResult) {
```
felipe-gdr, line 391: I think you could move all that to `executeTopLevelField` to avoid having to pass the extra params down, and also to keep both calls to `transform` (for the initial and incremental data) close together

lib/src/main/java/graphql/nadel/engine/transform/hydration/NadelHydrationTransform.kt
```diff
@@ -145,6 +145,7 @@ internal class NadelHydrationTransform(
         return makeTypeNameField(
             aliasHelper = state.aliasHelper,
             objectTypeNames = objectTypeNames,
+            deferredExecutions = linkedSetOf(),
```
felipe-gdr, line 148: Does that mean that, for now, hydrations (batched and non-batched) won't work when nested inside deferred fields?

lib/src/main/java/graphql/nadel/engine/transform/result/NadelResultTransformer.kt
```diff
@@ -26,9 +33,23 @@ internal class NadelResultTransformer(private val executionBlueprint: NadelOvera
         service: Service,
         result: ServiceExecutionResult,
     ): ServiceExecutionResult {
+        if (result is NadelIncrementalServiceExecutionResult) {
```
felipe-gdr, line 36: Can we remove this whole `if` block? It seems like transformations for incremental data is being done on the overloaded version of this function now.

lib/src/main/java/graphql/nadel/engine/transform/result/json/JsonNodes.kt
```diff
@@ -46,6 +48,22 @@ class NadelCachingJsonNodes(
      * Extracts the nodes at the given query selection path.
      */
     private fun getNodesAt(rootNode: JsonNode, queryPath: NadelQueryPath, flatten: Boolean = false): List<JsonNode> {
+        // if incremental payload, remove prefix
+        if (prefix != null && !removedPrefix) {
```
felipe-gdr, line 52: Can this quasi-recursive call be replaced with something like this?
```
        val queryPath = if (prefix != null && queryPathArg.startsWith(prefix.toList()) ) {
            queryPathArg.removePrefix(prefix)
        } else {
            queryPathArg
        }
```

test/src/test/kotlin/graphql/nadel/tests/next/fixtures/defer/transforms/RenameInnerFieldInsideDeferSnapshot.kt
```diff
@@ -0,0 +1,111 @@
+// @formatter:off
+package graphql.nadel.tests.next.fixtures.defer.transforms
```
felipe-gdr, line 2: Can we add tests with lists as well? I want to ensure that the path manipulation logic we're introducing works well with list indexes, etc.

lib/src/main/java/graphql/nadel/engine/transform/result/NadelResultTransformer.kt
```diff
@@ -73,6 +83,65 @@ internal class NadelResultTransformer(private val executionBlueprint: NadelOvera
...
+                                underlyingFields.first().parent,
+                                result,
+                                step.state,
+                                nodes,
+                            )
+                        },
+                    )
+                }
+
+                asyncInstructions.add(
```
felipe-gdr, line 129: should this be inside the `for` loop?

lib/src/main/java/graphql/nadel/engine/transform/result/json/JsonNodes.kt
```diff
@@ -34,18 +34,27 @@ interface JsonNodes {
...
     override fun getNodesAt(queryPath: NadelQueryPath, flatten: Boolean): List<JsonNode> {
         val rootNode = JsonNode(data)
-        return getNodesAt(rootNode, queryPath, flatten)
+
+        return if (pathPrefix == null) {
+            getNodesAt(rootNode, queryPath, flatten)
+        } else if (queryPath.startsWith(pathPrefix.segments)) {
+            getNodesAt(rootNode, queryPath.removePrefix(pathPrefix.segments), flatten)
+        } else {
+            emptyList()
```
felipe-gdr, line 49: should we `Assert.shouldNotHappen` here instead of returning an empty list?

lib/src/main/java/graphql/nadel/engine/transform/result/json/JsonNodes.kt
```diff
@@ -34,18 +36,27 @@ interface JsonNodes {
  */
 class NadelCachingJsonNodes(
     private val data: JsonMap,
+    private val pathPrefix: NadelQueryPath? = null, // for incremental (defer) payloads, we pass in the prefix we need to remove from path
```
gnawf, line 39: Move this into a Javadoc comment `/**`

lib/src/main/java/graphql/nadel/engine/transform/result/json/JsonNodes.kt
```diff
@@ -34,18 +36,27 @@ interface JsonNodes {
...
     override fun getNodesAt(queryPath: NadelQueryPath, flatten: Boolean): List<JsonNode> {
         val rootNode = JsonNode(data)
-        return getNodesAt(rootNode, queryPath, flatten)
+
+        return if (pathPrefix == null) {
+            getNodesAt(rootNode, queryPath, flatten)
+        } else if (queryPath.startsWith(pathPrefix.segments)) {
+            getNodesAt(rootNode, queryPath.removePrefix(pathPrefix.segments), flatten)
+        } else {
+            assertShouldNeverHappen("The supplied path prefix did not match the start of the query path")
```
gnawf, line 51: Change this back to `emptyList`

lib/src/main/java/graphql/nadel/engine/util/GraphQLUtil.kt
```diff
@@ -365,7 +367,18 @@ fun ExecutionIdProvider.provide(executionInput: ExecutionInput): ExecutionId {
...
+fun NadelIncrementalServiceExecutionResult.copy(
+    data: MutableJsonMap = this.data,
+    errors: MutableList<MutableJsonMap> = this.errors,
+    extensions: MutableJsonMap = this.extensions,
+    incremental: List<IncrementalPayload>? = this.incremental,
+    incrementalItemPublisher: Publisher<DelayedIncrementalPartialResult> = this.incrementalItemPublisher,
+    hasNext: Boolean = this.hasNext,
+): ServiceExecutionResult {
+    return newIncrementalServiceExecutionResult(data, errors, extensions, incremental, incrementalItemPublisher, hasNext)
+}
```
gnawf, line 379: Should just make `NadelIncrementalServiceExecutionResult` a `data` class and remove this.

lib/src/main/java/graphql/nadel/engine/util/GraphQLUtil.kt
```diff
@@ -381,6 +394,17 @@ fun newServiceExecutionResult(
...
+fun newIncrementalServiceExecutionResult(
+    data: MutableJsonMap = mutableMapOf(),
+    errors: MutableList<MutableJsonMap> = mutableListOf(),
+    extensions: MutableJsonMap = mutableMapOf(),
+    incremental: List<IncrementalPayload>?,
+    incrementalItemPublisher: Publisher<DelayedIncrementalPartialResult>,
+    hasNext: Boolean,
+): ServiceExecutionResult {
+    return NadelIncrementalServiceExecutionResult(data, errors, extensions, incremental, incrementalItemPublisher, hasNext)
+}
```
gnawf, line 406: And then we should be able to remove this

lib/src/main/java/graphql/nadel/engine/transform/result/NadelResultTransformer.kt
```diff
@@ -110,6 +167,13 @@ internal class NadelResultTransformer(private val executionBlueprint: NadelOvera
         mutableErrors.add(newError)
     }
 
+    private fun processGraphQLErrors(
+        instruction: NadelResultInstruction.AddError,
+        errors: List<GraphQLError>?,
+    ) {
+        errors?.asMutable()?.add(instruction.error)
```
gnawf, line 174: Need to handle the case where these are null in the future.

lib/src/main/java/graphql/nadel/engine/transform/result/json/JsonNodes.kt
```diff
@@ -11,6 +13,9 @@ import java.util.concurrent.ConcurrentHashMap
  *
  * Use [NadelCachingJsonNodes] for the most part because that is faster.
  * It is the default implementation.
+ *
+ * @param data The JSON map data.
+ * @param pathPrefix For incremental (defer) payloads, this is the prefix that needs to be removed from the path.
```
gnawf, line 18: Wait this comment doesn't belong here…

Should be on `JsonNodes.Companion.invoke` or so

gnawf (approved):

lib/src/main/java/graphql/nadel/engine/transform/hydration/NadelHydrationTransform.kt
felipe-gdr: Should this be 
```        
        return makeTypeNameField(
            aliasHelper = state.aliasHelper,
            objectTypeNames = objectTypeNames,
            deferredExecutions = field.deferredExecutions,
        )
        ```
instead?

felipe-gdr (approved):
Thanks @sbarker2 
Can you just please have a look at this comment before merging:

https://github.com/atlassian-labs/nadel/pull/579/files#r1735501506

PR #582, opened 2024-08-28
Adding tests for the other transforms with defer

sbarker2:
Adding tests for the other transforms with defer.

This PR adds tests for these transforms:
- deep rename
- type rename
- skipinclude

There are already existing tests for these transforms:
- rename
- hydration


test/src/test/kotlin/graphql/nadel/tests/next/fixtures/defer/transforms/DeferredDeepRenameTest.kt
```diff
@@ -0,0 +1,72 @@
+package graphql.nadel.tests.next.fixtures.defer.transforms
+
+import graphql.nadel.NadelExecutionHints
+import graphql.nadel.tests.next.NadelIntegrationTest
+
+open class DeferredDeepRenameTest : NadelIntegrationTest(
+    query = """
+      query {
+        details {
```
felipe-gdr, line 9: This query is not using `@defer` anywhere

felipe-gdr (changes requested):
One of the tests is not using `@defer` at all

felipe-gdr (approved):

PR #583, opened 2024-08-29
Fix delayed result snapshot

gnawf:
The test snapshots were wrong because we weren't deep cloning the result before Nadel mutated it.

Pending #579


test/src/test/kotlin/graphql/nadel/tests/next/fixtures/defer/transforms/DeferredFieldIsRenamedTestSnapshot.kt
```diff
@@ -55,7 +55,8 @@ public class DeferredFieldIsRenamedTestSnapshot : TestSnapshot() {
                     |         "defer"
                     |       ],
                     |       "data": {
-                    |         "overallString": "string for the deferred renamed field"
```
gnawf, line 58: Because we didn't clone the result before Nadel mutated it, the test snapshots wrote down the final result that already had the field renamed.

felipe-gdr (dismissed):

gnawf:
Hey @felipe-gdr had to push to add more fixes to tests introduced in #586

sbarker2 (approved):

felipe-gdr (approved):

PR #584, opened 2024-08-29
Fix renamed shared input type bug

gnawf:


test/src/test/kotlin/graphql/nadel/tests/next/fixtures/rename/RenamedSharedInputTypeTestSnapshot.kt
```diff
@@ -0,0 +1,98 @@
...
+ *
+ * Refer to [graphql.nadel.tests.next.UpdateTestSnapshots
+ */
+@Suppress("unused")
+public class RenamedSharedInputTypeTestSnapshot : TestSnapshot() {
+    override val calls: List<ExpectedServiceCall> = listOf(
+            ExpectedServiceCall(
+                service = "confluence_something",
+                query = """
+                | query (${'$'}v0: PathType!) {
```
gnawf, line 27: Prior to this fix the query was

```
| query (${'$'}v0: ConfluenceLegacyPathType!) {
```

lib/src/main/java/graphql/nadel/engine/blueprint/NadelExecutionBlueprintFactory.kt
```diff
@@ -771,44 +771,68 @@ private class SharedTypesAnalysis(
...
+        val overallOutputTypeName = overallField.type.unwrapAll().name
+        val underlyingOutputTypeName = underlyingField.type.unwrapAll().name
+        val outputTypeRenameInstruction = getTypeRenameInstructionOrNull(
+            overallTypeName = overallOutputTypeName,
+            underlyingTypeName = underlyingOutputTypeName,
+            serviceDefinedTypes = serviceDefinedTypes,
+            service = service,
+        )
+
+        val argumentTypeRenameInstructions = overallField.inputValueDefinitions
```
andimarek-atlassian, line 786: so we never considered input type renames here?

andimarek-atlassian (approved):

lib/src/main/java/graphql/nadel/engine/blueprint/NadelExecutionBlueprintFactory.kt
gnawf: Not shared ones. This code is looking for references to types used by the service, but NOT defined by the service itself.

As we discussed elsewhere, it should really reuse the validation code that uses a schema traverser to pick up all referenced types.

PR #588, opened 2024-09-06
Avoid hasNext=true on the last incremental payload

felipe-gdr:
The computation of `hasNext` is somewhat flaky, as we can see in the occasional failure of `HydrationDeferGroupingTest`.

The failure occurs because sometimes the last deferred payload contains `hasNext=true`. 
This is possible since not all results from an outstanding job will be converted to a deferred payload and sent to the results channel.

So we might have something like this
1. defer task A starts -> outstanding jobs count is 1
2. defer task B starts  -> outstanding jobs count is 2
3. defer task A completes and results on a non-null defer payload after being processed by the accumulator -> outstanding jobs count is 1, so hasNext=true
4. defer task B completes and results in null, so no events are emitted and the publisher finishes producing.

The last (and only) event that was actually emitted has hasNext=true, so clients will hang waiting for a completion event forever.

Note: this PR doesn't fix the flaky test (yet)

### DIFF
--- a/lib/src/main/java/graphql/nadel/engine/NadelIncrementalResultAccumulator.kt
+++ b/lib/src/main/java/graphql/nadel/engine/NadelIncrementalResultAccumulator.kt
@@ -115,14 +115,6 @@ class NadelIncrementalResultAccumulator(
             }
 
         if (readyAccumulators.isEmpty()) {
-            if(!hasNext) {
-                // We have to return hasNext=false to indicate to clients that there's no more data coming.
-                // Note: the spec allows an empty payload which only contains hastNext=false to be returned.
-                return DelayedIncrementalPartialResultImpl.newIncrementalExecutionResult()
-                    .incrementalItems(emptyList())
-                    .hasNext(false)
-                    .build()
-            }
             return null
         }
 
--- a/lib/src/main/java/graphql/nadel/engine/NadelIncrementalResultAccumulator.kt
+++ b/lib/src/main/java/graphql/nadel/engine/NadelIncrementalResultAccumulator.kt
@@ -40,7 +40,7 @@ import graphql.normalized.incremental.NormalizedDeferredExecution
  * the user.
  */
 class NadelIncrementalResultAccumulator(
-    private val operation: ExecutableNormalizedOperation,
+    operation: ExecutableNormalizedOperation,
 ) {
     data class DeferAccumulatorKey(
         val incrementalPayloadPath: List<Any>,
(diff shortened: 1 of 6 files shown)


lib/src/main/java/graphql/nadel/engine/NadelIncrementalResultAccumulator.kt
```diff
@@ -115,6 +115,14 @@ class NadelIncrementalResultAccumulator(
             }
 
         if (readyAccumulators.isEmpty()) {
+            if(!hasNext) {
+                // We have to return hasNext=false to indicate to clients that there's no more data coming.
+                // Note: the spec allows an empty payload which only contains hastNext=false to be returned.
+                return DelayedIncrementalPartialResultImpl.newIncrementalExecutionResult()
+                    .incrementalItems(emptyList())
+                    .hasNext(false)
+                    .build()
```
gnawf, line 124: I think this belongs in `NadelIncrementalResultSupport`

This class just accumulates data. This is some incremental result logic.

lib/src/main/java/graphql/nadel/engine/NadelIncrementalResultSupport.kt
```diff
@@ -145,21 +143,6 @@ class NadelIncrementalResultSupport internal constructor(
         initialCompletionLock.complete(Unit)
     }
 
-    fun close() {
-        coroutineScope.cancel()
-    }
```
gnawf, line 150: HUH, this needs to be called. I think it's a bug that it's not.

i.e. we should cancel the jobs launched on the defer scope if the incremental result is cancelled e.g. on request closed early.

We don't have to address that in this PR.

I _suspect_ we can edit `resultFlow` to something like

```kotlin
    fun resultFlow(): Flow<DelayedIncrementalPartialResult> {
        return resultFlow
            .onCompletion {
                close()
            }
    }
```

Where `onCompletion` is invoked on cancellation.

test/src/test/kotlin/graphql/nadel/tests/next/NadelIntegrationTest.kt
```diff
@@ -379,7 +379,9 @@ abstract class NadelIntegrationTest(
             // Note: there exists a IncrementalExecutionResult.getIncremental but that is part of the initial result
             assertTrue(result is IncrementalExecutionResult)
 
-            // Fuck why delayed & incremental?? Shouldn't incremental == delayed? Why is there an optional synchronous incremental??
```
gnawf, line 382: Sorry, should've deleted that

lib/src/main/java/graphql/nadel/engine/NadelIncrementalResultSupport.kt
```diff
@@ -145,19 +152,17 @@ class NadelIncrementalResultSupport internal constructor(
...
-        subject: DelayedIncrementalPartialResult,
-        hasNext: Boolean,
-    ): DelayedIncrementalPartialResult {
-        return if (subject.hasNext() == hasNext) {
-            subject
-        } else {
-            subject.copy(hasNext = hasNext)
-        }
+    // We have to return hasNext=false to indicate to clients that there's no more data coming.
+    // Note: the spec allows an empty payload which only contains hastNext=false to be returned.
```
gnawf, line 160: Nit: Javadoc to make comment sticky to function

gnawf (approved):
Nice, thanks.

PR #591, opened 2024-09-25
Partition calls

felipe-gdr:

- [ yes ] Add tests that use __typename in queries
- [ yes ] Does this change work with all nadel transformations (rename, type rename, hydration, etc)? Add tests for this.
- [ FF can be in AGG] Is it worth using hints for this change in order to be able to enable a percentage rollout?
- [ yes ] Do we need to add integration tests for this change in the graphql gateway?
- [ maybe ] Do we need a pollinator check for this?


sbarker2 (approved):
Nice work 👍

sbarker2:
Love the comprehensive test coverage

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionMutationPayloadMerger.kt
```diff
@@ -0,0 +1,87 @@
...
+                field = overallField
+            )
+        )
+    }
+
+    /**
+     * A GraphQL type is considered to be a "mutation payload" (for the purposes of this transform)  if
+     * it has a `success` field of type `Boolean!` plus an `errors` field of type List and any other number
+     * of fields of type List
+     */
```
temaEmelyan, line 72: I think atlassian public schema defines Payload as
```graphql
"""
The general shape of a mutation response.
"""
interface Payload {
    "Was this mutation successful"
    success: Boolean!

    "A list of errors if the mutation was not successful"
    errors: [MutationError!]
}
```
would it be reasonable to check if the type extends `Payload`

temaEmelyan (approved):

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelFieldPartition.kt
```diff
@@ -0,0 +1,167 @@
...
+
+class NadelFieldPartition(
+    private val pathToPartitionArg: List<String>,
+    private val fieldPartitionContext: Any,
+    private val graphQLSchema: GraphQLSchema,
+    private val partitionKeyExtractor: (
+        scalarValue: ScalarValue<*>,
+        inputValueDef: GraphQLInputValueDefinition,
+        context: Any,
+    ) -> String?,
```
gnawf, line 22: Would like to see this made into an interface so it's easier to parse e.g.

```kotlin
interface NadelPartitionKeyExtractor {
  fun getPartitionKey(
        scalarValue: ScalarValue<*>,
        inputValueDef: GraphQLInputValueDefinition,
        context: Any,
    ): String
}
```

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelFieldPartition.kt
```diff
@@ -0,0 +1,167 @@
...
+import graphql.normalized.ExecutableNormalizedField
+import graphql.normalized.NormalizedInputValue
+import graphql.schema.GraphQLArgument
+import graphql.schema.GraphQLFieldDefinition
+import graphql.schema.GraphQLInputObjectType
+import graphql.schema.GraphQLInputValueDefinition
+import graphql.schema.GraphQLSchema
+import graphql.schema.GraphQLTypeUtil
+
+class NadelFieldPartition(
```
gnawf, line 14: Should be `internal` so code outside library can't access it.

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelFieldPartitionException.kt
```diff
@@ -0,0 +1,16 @@
+package graphql.nadel.engine.transform.partition
+
+import graphql.ErrorClassification
+import graphql.nadel.error.NadelGraphQLErrorException
+
+class NadelPartitionGraphQLErrorException(
+    message: String,
+    path: List<Any>? = null,
+    errorClassification: ErrorClassification? = partitioningErrorClassification,
```
gnawf, line 9: Don't think `errorClassification` is ever variable, so we should hard code it into the super call i.e.

```
class NadelPartitionGraphQLErrorException(
    message: String,
    path: List<Any>? = null,
) : NadelGraphQLErrorException(message, path, partitioningErrorClassification)
```

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionListMerger.kt
```diff
@@ -0,0 +1,45 @@
...
+import graphql.nadel.engine.transform.result.NadelResultInstruction
+import graphql.nadel.engine.transform.result.json.JsonNode
+import graphql.nadel.engine.util.unwrapNonNull
+import graphql.normalized.ExecutableNormalizedField
+import graphql.schema.GraphQLList
+import graphql.schema.GraphQLObjectType
+import graphql.schema.GraphQLOutputType
+import graphql.schema.GraphQLScalarType
+
+object NadelPartitionListMerger {
```
gnawf, line 12: Should be `internal`

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionListMerger.kt
```diff
@@ -0,0 +1,45 @@
...
+                } else {
+                    check(it is List<*>) { "Expected a list, but got ${it::class.simpleName}" }
+                    it
+                }
+            }.flatten()
+
+        val thisNodesDataCast = thisNodesData?.let {
+            check(it is List<*>) { "Expected a list, but got ${it::class.simpleName}" }
+            it
+        } ?: emptyList<NadelResultInstruction.Set>()
```
gnawf, line 34: Are the generics here correct?

Should just be `emptyList<Any?>()` right?

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionListMerger.kt
```diff
@@ -0,0 +1,45 @@
...
+import graphql.schema.GraphQLList
+import graphql.schema.GraphQLObjectType
+import graphql.schema.GraphQLOutputType
+import graphql.schema.GraphQLScalarType
+
+object NadelPartitionListMerger {
+    fun mergeDataFromList(
+        dataFromPartitionCalls: List<Any?>,
+        thisNodesData: Any?,
+        parentNodes: List<JsonNode>,
```
gnawf, line 16: Maybe invoke `.single()` at the call site and pass in `parentNode: JsonNode` (singular)

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionMutationPayloadMerger.kt
```diff
@@ -0,0 +1,87 @@
...
+import graphql.nadel.engine.transform.result.NadelResultInstruction
+import graphql.nadel.engine.transform.result.json.JsonNode
+import graphql.nadel.engine.util.unwrapNonNull
+import graphql.normalized.ExecutableNormalizedField
+import graphql.schema.GraphQLList
+import graphql.schema.GraphQLObjectType
+import graphql.schema.GraphQLOutputType
+import graphql.schema.GraphQLScalarType
+
+object NadelPartitionMutationPayloadMerger {
```
gnawf, line 12: Should be `internal`

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionMutationPayloadMerger.kt
```diff
@@ -0,0 +1,87 @@
...
+            && this.getField("errors") != null
+            && this.fields.filter { it.name != "success" }.all { it.type.unwrapNonNull() is GraphQLList }
+    }
+}
+
+fun Any?.safeToBoolean(): Boolean {
+    return (this as? Boolean) ?: false
+}
+
+fun Any?.safeToList(): List<Any> {
```
gnawf, line 85: Should be private unless we put it in a util class.

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionTransformHook.kt
```diff
@@ -0,0 +1,54 @@
...
+     */
+    fun getFieldPartitionContext(
+        executionContext: NadelExecutionContext,
+        serviceExecutionContext: NadelServiceExecutionContext,
+        executionBlueprint: NadelOverallExecutionBlueprint,
+        services: Map<String, Service>,
+        service: Service,
+        overallField: ExecutableNormalizedField,
+        hydrationDetails: ServiceExecutionHydrationDetails?,
+    ): Any? {
```
gnawf, line 35: Think we should make a parent class like `NadelFieldPartitionContext` so we don't have to pass around `Any`

e.g.

```
abstract class NadelFieldPartitionContext
```

Similar to how `NadelServiceExecutionContext` does it.

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionTransformHook.kt
```diff
@@ -0,0 +1,54 @@
...
+     */
+    fun getPartitionKeyExtractor(): ((ScalarValue<*>, GraphQLInputValueDefinition, Any?) -> String?) {
+        return {_, _, _-> null}
+    }
+
+    /**
+     * This hook allows you to perform some non-side effect action on the field partitions (like logging, etc.).
+     * This is called before the field partitions calls are executed, and whether the partition requests are successful or not.
+     */
+    fun willPartitionCallback(executionContext: NadelExecutionContext, fieldPartitions: Map<String, ExecutableNormalizedField>) {
```
gnawf, line 51: Maybe `onPartition`

I'm used to `on{Event}` names for these kind of functions

nbd

lib/src/main/java/graphql/nadel/util/MapUtil.kt
```diff
@@ -0,0 +1,38 @@
+package graphql.nadel.util
+
+internal object MapUtil {
```
gnawf, line 3: Is this used? in IDE I can't find references

lib/src/main/java/graphql/nadel/util/MapUtil.kt
```diff
@@ -0,0 +1,38 @@
...
+                tempOriginal[value] =
+                    (tempOriginal[value] as List<*>) + tempPartitioned.flatMap { it[value] as List<*> }
+            } else {
+                return null
+            }
+        }
+        return immutableMap
+    }
+
+    fun <K, V> Map<K, V>.toMutableMapRecursively(): MutableMap<K, V> {
```
gnawf, line 28: If it's from Jackson it should always be mutable.

So you just force cast it to the mutable version of the type.

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionTransform.kt
```diff
@@ -0,0 +1,266 @@
...
+
+        val fieldPartitions = checkNotNull(state.fieldPartitions) { "Expected fieldPartitions to be set" }
+
+        val primaryPartition = fieldPartitions.values.first()
+
+        // TODO: throw error if operation is Subscription?
+        val rootType = executionContext.query.operation.getType(executionBlueprint.engineSchema)
+
+        val partitionCalls = coroutineScope {
+            fieldPartitions.values.drop(1).map {
```
gnawf, line 143: Could put `fieldPartitions.values.drop(1)` right below the definition of `primaryPartition` so it's easier to understand i.e.

```kotlin
val primaryPartition = fieldPartitions.values.first()
val otherPartitions = fieldPartitions.values.drop(1)

```

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionTransform.kt
```diff
@@ -0,0 +1,266 @@
...
+        field: ExecutableNormalizedField,
+        state: State,
+    ): NadelTransformFieldResult {
+
+        if (state.errors != null) {
+            // At this point we know we can't partition the field, but removing it could result
+            // in a query with no top-level fields, which would be invalid.
+            // Adding __typename as an artificial field also wouldn't work because that field
+            // belongs to all operation types, which causes issues further down the execution.
+            return NadelTransformFieldResult.unmodified(field)
```
gnawf, line 132: I feel like we should just stop the execution here?

Like could perhaps we just `throw SomeNadelException` or so and it should include that in the response if the thrown type is a `GraphQLError` I think.

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionTransform.kt
```diff
@@ -0,0 +1,266 @@
...
+        }
+
+        val fieldPartitions = checkNotNull(state.fieldPartitions) { "Expected fieldPartitions to be set" }
+
+        val primaryPartition = fieldPartitions.values.first()
+
+        // TODO: throw error if operation is Subscription?
+        val rootType = executionContext.query.operation.getType(executionBlueprint.engineSchema)
+
+        val partitionCalls = coroutineScope {
```
gnawf, line 142: As mentioned easier on call, this `coroutineScope` is only going to return once the child `async` jobs are done.

So we need to launch jobs off a `supervisorJob` earlier.

Like maybe

```diff
diff --git a/lib/src/main/java/graphql/nadel/NextgenEngine.kt b/lib/src/main/java/graphql/nadel/NextgenEngine.kt
index f57a6a5b..54f28ae2 100644
--- a/lib/src/main/java/graphql/nadel/NextgenEngine.kt
+++ b/lib/src/main/java/graphql/nadel/NextgenEngine.kt
@@ -70,6 +70,7 @@ import kotlinx.coroutines.future.await
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.reactive.asFlow
 import kotlinx.coroutines.reactive.asPublisher
+import kotlinx.coroutines.supervisorScope
 import java.util.concurrent.CompletableFuture
 import graphql.normalized.ExecutableNormalizedOperationFactory.Options.defaultOptions as executableNormalizedOperationFactoryOptions
 
@@ -144,7 +145,7 @@ internal class NextgenEngine(
         queryDocument: Document,
         instrumentationState: InstrumentationState?,
         executionHints: NadelExecutionHints,
-    ): ExecutionResult {
+    ): ExecutionResult = supervisorScope {
         try {
             val timer = NadelInstrumentationTimer(
                 instrumentation,
@@ -177,6 +178,7 @@ internal class NextgenEngine(
                 timer,
                 incrementalResultSupport,
                 resultTracker,
+                executionCoroutine = this,
             )
 
             val beginExecuteContext = instrumentation.beginExecute(
@@ -226,7 +228,7 @@ internal class NextgenEngine(
             // todo: maybe pass in the incremental version that's built below into here
             resultTracker.complete(result)
 
-            return if (incrementalResultSupport.hasDeferredResults()) {
+            if (incrementalResultSupport.hasDeferredResults()) {
                 IncrementalExecutionResultImpl.Builder()
                     .from(result)
                     .incrementalItemPublisher(incrementalResultSupport.resultFlow().asPublisher())
@@ -236,7 +238,7 @@ internal class NextgenEngine(
             }
         } catch (e: Throwable) {
             when (e) {
-                is GraphQLError -> return newExecutionResult(error = e)
+                is GraphQLError -> newExecutionResult(error = e)
                 else -> throw e
             }
         }
diff --git a/lib/src/main/java/graphql/nadel/engine/NadelExecutionContext.kt b/lib/src/main/java/graphql/nadel/engine/NadelExecutionContext.kt
index a9c26e76..4e6e0bc8 100644
--- a/lib/src/main/java/graphql/nadel/engine/NadelExecutionContext.kt
+++ b/lib/src/main/java/graphql/nadel/engine/NadelExecutionContext.kt
@@ -11,6 +11,7 @@ import graphql.nadel.hooks.CreateServiceContextParams
 import graphql.nadel.hooks.NadelExecutionHooks
 import graphql.nadel.result.NadelResultTracker
 import graphql.normalized.ExecutableNormalizedOperation
+import kotlinx.coroutines.CoroutineScope
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.ConcurrentHashMap
 
@@ -24,7 +25,8 @@ data class NadelExecutionContext internal constructor(
     internal val incrementalResultSupport: NadelIncrementalResultSupport,
     internal val resultTracker: NadelResultTracker,
     internal val hydrationDetails: ServiceExecutionHydrationDetails? = null,
-    internal val isPartitionedCall: Boolean = false
+    internal val isPartitionedCall: Boolean = false,
+    internal val executionCoroutine: CoroutineScope,
 ) {
     private val serviceContexts = ConcurrentHashMap<String, CompletableFuture<Any?>>()
```

Then we can do thing like

```kotlin
val partitionCalls = otherPartitions.map {
    executionContext.executionCoroutine.async {
        val topLevelField = NFUtil.createField(
            executionBlueprint.engineSchema,
            rootType,
            // TODO: queryPath contains field aliases, not field names, which results in an error.
            field.queryPath,
            it.normalizedArguments,
            it.children
        )

        engine.executePartitionedCall(topLevelField, service, state.executionContext)
    }
}
```

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionTransform.kt
```diff
@@ -0,0 +1,266 @@
...
+        overallField: ExecutableNormalizedField,
+        underlyingParentField: ExecutableNormalizedField?,
+        result: ServiceExecutionResult,
+        state: State,
+        nodes: JsonNodes,
+    ): List<NadelResultInstruction> {
+        val parentNodes = nodes.getNodesAt(
+            queryPath = underlyingParentField?.queryPath ?: NadelQueryPath.root,
+            flatten = true,
+        )
```
gnawf, line 178: Maybe a `.singleOrNull() ?: return emptyList()`

So we do things like `parentNodes.first()` later

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionTransform.kt
```diff
@@ -0,0 +1,266 @@
...
+
+        if (parentNodes.size != 1) {
+            // TODO: Support multiple parent nodes (interfaces, unions)
```
gnawf, line 181: Multiple parent nodes should only be if it's a list in the result. Don't think abstract types would cause that.

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionMutationPayloadMerger.kt
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * A GraphQL type is considered to be a "mutation payload" (for the purposes of this transform)  if
+     * it has a `success` field of type `Boolean!` plus an `errors` field of type List and any other number
+     * of fields of type List
+     */
+    fun GraphQLOutputType.isMutationPayloadLike(): Boolean {
+        return this is GraphQLObjectType
+            && this.getField("success")?.let { it.type.unwrapNonNull() as? GraphQLScalarType }?.name == "Boolean"
+            && this.getField("errors") != null
+            && this.fields.filter { it.name != "success" }.all { it.type.unwrapNonNull() is GraphQLList }
```
gnawf, line 77: Could be

```
&& this.fields.all { it.name == "success" || it.type.unwrapNonNull() is GraphQLList }
```

Since `.filter` and other transform functions return an entire new `List` unless the transform is on a `Sequence`

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionListMerger.kt
```diff
@@ -0,0 +1,45 @@
...
+        overallField: ExecutableNormalizedField,
+    ): List<NadelResultInstruction.Set> {
+        val parentNode = parentNodes.first()
+
+        val listDataFromPartitionCalls = dataFromPartitionCalls
+            .mapNotNull {
```
gnawf, line 22: Could be

```kotlin
.flatMap {
  if (it == null) {
    emptyList()
  } else {
    check(it is List<*>) { "Expected a list, but got ${it::class.simpleName}" }
    it
  }
}
```

Also I think that `ClassCastException` will print out the expected/actual type. So could just straight up cast it.

```kotlin
.flatMap {
  it as List<*>?
}
```

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionTransform.kt
```diff
@@ -0,0 +1,266 @@
...
+                    NadelPartitionGraphQLErrorException(
+                        "The call for field '${overallField.resultKey}' was not partitioned due to the following error: '${it.message}'",
+                        path = overallField.queryPath.segments,
+                    )
+                )
+            } + nullifyField
+        }
+
+        // TODO: handle HTTP errors
+        val resultFromPartitionCalls = state.partitionCalls.awaitAll()
```
gnawf, line 203: Nice

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelFieldPartition.kt
```diff
@@ -0,0 +1,167 @@
...
+    private val pathToPartitionArg: List<String>,
+    private val fieldPartitionContext: Any,
+    private val graphQLSchema: GraphQLSchema,
```
gnawf, line 17: Maybe be explicit about it being the `engineSchema` since there's a few schema objects.

gnawf (commented):
Left some comments, we can address them later too.

In general looks good.

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionMutationPayloadMerger.kt
felipe-gdr: I considered it, but eventually decided not to do it because the type for the use case we're targeting doesn't actually implement `Payload`. But you make a good point, that type should really be implementing this interface. They're already adhering to the contract, so it shouldn't be a big change for them 👍

felipe-gdr:
> Left some comments, we can address them later too.
> 
> In general looks good.

Thanks for the great suggestions 🙏 

I'll merge it as is to unblock subsequent changes quickly and raise follow-up PRs to address your feedback.

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionTransform.kt
felipe-gdr: > if it's a list in the result

Hm, when would that happen? I'm deailng with fields that have type of `List` and `parentNodes` is always 1 🤔

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionTransformHook.kt
felipe-gdr: Yep, much better. That's a well known convention

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionTransformHook.kt
felipe-gdr: Love it. It will make the code much more readable.

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelFieldPartition.kt
felipe-gdr: Nice.

lib/src/main/java/graphql/nadel/util/MapUtil.kt
felipe-gdr: Ha, left over from some experiment I was doing.

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionTransform.kt
felipe-gdr: Nice

PR #594, opened 2024-10-14
Add validation for partition directive

felipe-gdr:
Review https://github.com/atlassian-labs/nadel/pull/591 first ;-)


test/src/test/kotlin/graphql/nadel/tests/next/fixtures/partition/MutationPartitionWithInvalidReturnTypeTest.kt
felipe-gdr, line 1: This scenario is no longer reproducible, as it results in validation errors thus the query never gets executed.

PR #596, opened 2024-10-16
Improvements in the partition code after PR feedback

felipe-gdr:
Most of @gnawf 's suggestions are addressed here. More PRs will come to address the remaining ones and also @temaEmelyan 's suggestion about the `Payload` interface 👍


lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionKeyExtractor.kt
```diff
@@ -0,0 +1,29 @@
...
+     * Function to extract the partition key from the given scalar value.
+     *
+     * A `null` return value indicates that the scalar value should not be partitioned.
+     */
+    fun getPartitionKey(
+        scalarValue: ScalarValue<*>,
+        inputValueDef: GraphQLInputValueDefinition,
+        context: NadelFieldPartitionContext,
+    ): String? {
+        return null
```
gnawf, line 20: Should this have a default? doesn't seem right

I think the `noop` version below should hardcode null, but the user should be otherwise forced to add an impl given this is the only function in this interface

gnawf (approved):
I think the interface definition seems wrong, but otherwise fine

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionKeyExtractor.kt
felipe-gdr: Yep makes sense 👍

gnawf (approved):

temaEmelyan (approved):

temaEmelyan (approved):

PR #597, opened 2024-10-21
Move partition instruction to Blueprint

felipe-gdr:
2 benefits out of this:
- we move part of the computation out of request processing
- don't rely on `overallField.getFieldDefinitions()` calls in the transform, which can cause problems when used in conjunction with hydrations and renames

### DIFF
--- a/lib/src/main/java/graphql/nadel/engine/blueprint/NadelExecutionBlueprintFactory.kt
+++ b/lib/src/main/java/graphql/nadel/engine/blueprint/NadelExecutionBlueprintFactory.kt
@@ -189,7 +189,7 @@ private class Factory(
                         when (val mappingDefinition = getFieldMappingDefinition(field)) {
                             null -> {
                                 getUnderlyingServiceHydrations(field)
-                                    .map { makeHydrationFieldInstruction(type, field, it) }
+                                    .map { makeHydrationFieldInstruction(type, field, it) } + makePartitionInstruction(type, field)
                             }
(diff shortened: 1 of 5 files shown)


lib/src/main/java/graphql/nadel/engine/blueprint/NadelExecutionBlueprintFactory.kt
```diff
@@ -471,6 +471,21 @@ private class Factory(
...
+    ): List<NadelPartitionInstruction> {
+        val partitionDefinition = NadelDirectives.createPartitionDefinition(field)
+            ?: return emptyList()
+
+        return listOf(
+            NadelPartitionInstruction(
+                location = makeFieldCoordinates(parentType, field),
+                pathToPartitionArg = partitionDefinition.pathToSplitPoint,
+            )
+        )
```
gnawf, line 486: Shouldn't this store more info like whether the output type is Payload like? etc.

Though maybe leave the changes here minimal because I'm exploring merging validation & blueprint code together in another branch.

gnawf (approved):

lib/src/main/java/graphql/nadel/engine/blueprint/NadelExecutionBlueprintFactory.kt
felipe-gdr: True. All this static information should live in the Blueprint.
I'll wait for your refactor to make the change.

PR #601, opened 2024-10-21
Refactor dsl code

gnawf:
Also deleted `@hydratedFrom` because nobody uses that.

We'll revisit it later because we do want to make hydration definitions easier.


felipe-gdr (approved):

felipe-gdr (dismissed):
Nice one!

PR #603, opened 2024-10-22
Throw error from transform instead of returning error transform instr…

felipe-gdr:


test/src/test/kotlin/graphql/nadel/tests/next/fixtures/partition/PartitionWithConflictingRoutingValuesTestSnapshot.kt
```diff
@@ -21,45 +21,7 @@ private suspend fun main() {
 @Suppress("unused")
 public class PartitionWithConflictingRoutingValuesTestSnapshot : TestSnapshot() {
     override val calls: List<ExpectedServiceCall> = listOf(
-            ExpectedServiceCall(
```
felipe-gdr, line 24: the service call on this test is no longer executed, since we're throwing an error on the `transformField` method

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionTransform.kt
```diff
@@ -189,17 +183,6 @@ internal class NadelPartitionTransform(
             newValue = null
         )
 
-        if (state.errors != null) {
```
felipe-gdr, line 192: Errors were handled in the `transformField` method, which executed just before.

temaEmelyan (approved):

PR #605, opened 2024-10-24
Fix some partition things

felipe-gdr:
Some general changes in the partition code

### DIFF
--- a/lib/src/main/java/graphql/nadel/engine/transform/partition/NadelFieldPartitionContext.kt
+++ b/lib/src/main/java/graphql/nadel/engine/transform/partition/NadelFieldPartitionContext.kt
@@ -5,6 +5,4 @@ package graphql.nadel.engine.transform.partition
  *
  * An instance is created for each field that is subject to partitioning.
  */
-abstract class NadelFieldPartitionContext {
-    internal object None : NadelFieldPartitionContext()
-}
+abstract class NadelFieldPartitionContext
--- a/lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionKeyExtractor.kt
+++ b/lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionKeyExtractor.kt
@@ -17,19 +17,4 @@ interface NadelPartitionKeyExtractor {
         inputValueDef: GraphQLInputValueDefinition,
         context: NadelFieldPartitionContext,
     ): String?
-
-    companion object {
-        /**
-         * A no-op partition key extractor that does not partition any scalar values.
-         */
-        val noop: NadelPartitionKeyExtractor = object : NadelPartitionKeyExtractor {
-            override fun getPartitionKey(
-                scalarValue: ScalarValue<*>,
-                inputValueDef: GraphQLInputValueDefinition,
-                context: NadelFieldPartitionContext,
-            ): String? {
-                return null
-            }
-        }
-    }
 }
--- a/lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionKeyExtractor.kt
+++ b/lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionKeyExtractor.kt
@@ -17,4 +17,16 @@ interface NadelPartitionKeyExtractor {
         inputValueDef: GraphQLInputValueDefinition,
         context: NadelFieldPartitionContext,
     ): String?
+
+    companion object {
+        internal val noop = object : NadelPartitionKeyExtractor {
+            override fun getPartitionKey(
+                scalarValue: ScalarValue<*>,
+                inputValueDef: GraphQLInputValueDefinition,
+                context: NadelFieldPartitionContext,
+            ): String? {
+                return null
+            }
+        }
+    }
 }
--- a/lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionTransformHook.kt
+++ b/lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionTransformHook.kt
@@ -30,17 +30,13 @@ interface NadelPartitionTransformHook {
         service: Service,
         overallField: ExecutableNormalizedField,
         hydrationDetails: ServiceExecutionHydrationDetails?,
-    ): NadelFieldPartitionContext {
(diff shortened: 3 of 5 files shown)


lib/src/main/java/graphql/nadel/engine/transform/partition/NadelFieldPartitionContext.kt
```diff
@@ -5,6 +5,4 @@ package graphql.nadel.engine.transform.partition
  *
  * An instance is created for each field that is subject to partitioning.
  */
-abstract class NadelFieldPartitionContext {
-    internal object None : NadelFieldPartitionContext()
```
felipe-gdr, line 9: There was no use for this object, I'd rather rely on the hook callback to return `null` to indicate no partition is needed for a certain field.

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionKeyExtractor.kt
```diff
@@ -17,19 +17,4 @@ interface NadelPartitionKeyExtractor {
         inputValueDef: GraphQLInputValueDefinition,
         context: NadelFieldPartitionContext,
     ): String?
-
-    companion object {
-        /**
-         * A no-op partition key extractor that does not partition any scalar values.
-         */
-        val noop: NadelPartitionKeyExtractor = object : NadelPartitionKeyExtractor {
```
felipe-gdr, line 25: No reason for a default noop. If the consumer code wants to implement the hook, it should really provide a real implementation, otherwise there's no point in implementing the hook at all.

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionTransformHook.kt
```diff
@@ -30,17 +30,13 @@ interface NadelPartitionTransformHook {
         service: Service,
         overallField: ExecutableNormalizedField,
         hydrationDetails: ServiceExecutionHydrationDetails?,
-    ): NadelFieldPartitionContext {
```
felipe-gdr, line 33: No point in having defaults. Those functions are necessary for partitioning to work at all

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionTransformHook.kt
```diff
@@ -30,17 +30,13 @@ interface NadelPartitionTransformHook {
...
+    ): NadelFieldPartitionContext?
 
     /**
      * This hook allows you to provide a partition key for a given scalar value.
      * Arguments with the same partition key will be sent on the same request.
      */
-    fun getPartitionKeyExtractor(): NadelPartitionKeyExtractor {
-        return NadelPartitionKeyExtractor.noop
-    }
+    fun getPartitionKeyExtractor(): NadelPartitionKeyExtractor
```
felipe-gdr, line 39: Now, the `onPartition` function hook below deserves a default impl, since it's optional.

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelFieldPartitionContext.kt
gnawf: I did omit the `None` when I commented in https://github.com/atlassian-labs/nadel/pull/591#discussion_r1802440977 😉

gnawf (dismissed):

lib/src/main/java/graphql/nadel/engine/transform/partition/NadelPartitionKeyExtractor.kt
gnawf: I assume the remaining `internal` impl is just for testing?

Edit: ah I see it's used in the default no-op impl

lib/src/main/java/graphql/nadel/hooks/NadelExecutionHooks.kt
```diff
@@ -106,7 +113,23 @@ interface NadelExecutionHooks {
     }
 
     fun partitionTransformerHook(): NadelPartitionTransformHook {
```
gnawf, line 115: Could also make this null, and omit the transform entirely if it's null

e.g.

```diff
diff --git a/lib/src/main/java/graphql/nadel/engine/plan/NadelExecutionPlanFactory.kt b/lib/src/main/java/graphql/nadel/engine/plan/NadelExecutionPlanFactory.kt
index 5608c3e0..ee6845f3 100644
--- a/lib/src/main/java/graphql/nadel/engine/plan/NadelExecutionPlanFactory.kt
+++ b/lib/src/main/java/graphql/nadel/engine/plan/NadelExecutionPlanFactory.kt
@@ -109,7 +109,9 @@ internal class NadelExecutionPlanFactory(
                 transforms = listOfTransforms(
                     NadelSkipIncludeTransform(),
                     NadelServiceTypeFilterTransform(),
-                    NadelPartitionTransform(engine, executionHooks.partitionTransformerHook()),
+                    executionHooks.partitionTransformerHook()?.let { partitionHook ->
+                        NadelPartitionTransform(engine, partitionHook)
+                    },
                     *transforms.toTypedArray(),
                     NadelDeepRenameTransform(),
                     NadelTypeRenameResultTransform(),
@@ -121,10 +123,10 @@ internal class NadelExecutionPlanFactory(
             )
         }
 
-        private fun listOfTransforms(vararg elements: NadelTransform<out Any>): List<NadelTransform<Any>> {
-            return elements.map {
+        private fun listOfTransforms(vararg elements: NadelTransform<out Any>?): List<NadelTransform<Any>> {
+            return elements.mapNotNull {
                 @Suppress("UNCHECKED_CAST") // Ssh it's okay
-                it as NadelTransform<Any>
+                it as NadelTransform<Any>?
             }
         }
     }
```

gnawf (approved):

lib/src/main/java/graphql/nadel/hooks/NadelExecutionHooks.kt
felipe-gdr: This would make it harder to feature flag, as this code path is executed only during Nadel instance creation

PR #608, opened 2024-10-31
fix empty query short circuit

temaEmelyan:

### DIFF
--- a/test/src/test/resources/fixtures/field removed/renamed-top-level-field-is-not-removed-short-circuit-hint-is-on.yml
+++ b/test/src/test/resources/fixtures/field removed/renamed-top-level-field-is-not-removed-short-circuit-hint-is-on.yml
@@ -1,0 +1,60 @@
+name: "renamed top level field is not removed short circuit hint is on"
+enabled: true
+# language=GraphQL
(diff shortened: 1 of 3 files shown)


test/src/test/resources/fixtures/field removed/renamed-top-level-field-is-not-removed-short-circuit-hint-is-on.yml
```diff
@@ -0,0 +1,60 @@
+name: "renamed top level field is not removed short circuit hint is on"
+enabled: true
+# language=GraphQL
+overallSchema:
+  CommentService: |
+    directive @toBeDeleted on FIELD_DEFINITION
+    type Query {
+      commentById(id: ID): Comment @renamed(from: "commentByIdUnderlying")
```
temaEmelyan, line 8: even tho there were old tests with renames, they were only testing a scenario where this field is removed. the bug triggers when the field is not removed and the hint is on. I tried this test on master and it blows up but is passing with the new fex

bbakerman (approved):

gnawf (approved):

PR #610, opened 2024-11-04
Rename hydration terms

gnawf:
image

### DIFF
--- a/docs/Primer.MD
+++ b/docs/Primer.MD
@@ -249,9 +249,9 @@ Is transformed to:
 data class NadelHydrationFieldInstruction(
     override val location: FieldCoordinates, // User.friend
     val hydratedFieldDef: GraphQLFieldDefinition, // Schema definition for User.friend
-    val actorService: Service, // Users service i.e. service owning users.friendById
-    val queryPathToActorField: NadelQueryPath, // users.friendById
-    val actorFieldDef: GraphQLFieldDefinition, // Schema definition for users.friendById
+    val backingService: Service, // Users service i.e. service owning users.friendById
+    val queryPathToBackingField: NadelQueryPath, // users.friendById
+    val backingFieldDef: GraphQLFieldDefinition, // Schema definition for users.friendById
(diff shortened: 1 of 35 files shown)


lib/src/main/java/graphql/nadel/definition/hydration/NadelHydrationDefinition.kt
```diff
@@ -33,7 +33,7 @@ class NadelHydrationDefinition(
                 "This allows you to hydrate new values into fields"
                 directive @hydrated(
                     "The target service"
-                    service: String!
+                    service: String
```
Juliano-Prado, line 36: is it right, service now can be null?

gnawf: Yeah sorry I kind of snuck that into this PR. That field is going to be removed in the future. It's going to be inferred from the service it's declared in.

docs/Primer.MD
```diff
@@ -249,9 +249,9 @@ Is transformed to:
 data class NadelHydrationFieldInstruction(
     override val location: FieldCoordinates, // User.friend
     val hydratedFieldDef: GraphQLFieldDefinition, // Schema definition for User.friend
-    val actorService: Service, // Users service i.e. service owning users.friendById
-    val queryPathToActorField: NadelQueryPath, // users.friendById
-    val actorFieldDef: GraphQLFieldDefinition, // Schema definition for users.friendById
+    val backingService: Service, // Users service i.e. service owning users.friendById
+    val queryPathToBackingField: NadelQueryPath, // users.friendById
+    val backingFieldDef: GraphQLFieldDefinition, // Schema definition for users.friendById
     val actorInputValueDefs: List<NadelHydrationActorInputDef>, // [{ name: "ids" value: "$source.friendId" }]
```
Juliano-Prado, line 255: what about this actorInput? Does not make sense to rename to backingInputValueDefs?

gnawf: Updated, thanks for pointing that out

Juliano-Prado (approved):
LGTM

PR #612, opened 2024-11-04
Build blueprint from validation

gnawf:


lib/src/main/java/graphql/nadel/validation/NadelEnumValidation.kt
```diff
@@ -6,35 +6,34 @@ import graphql.schema.GraphQLEnumType
 import graphql.schema.GraphQLEnumValueDefinition
 
 internal class NadelEnumValidation {
+    context(NadelValidationContext)
     fun validate(
         schemaElement: NadelServiceSchemaElement,
-    ): List<NadelSchemaValidationError> {
+    ): NadelSchemaValidationResult {
```
gnawf, line 12: Commenting here about `NadelSchemaValidationResult` as this is the first reference in the diff.

`NadelSchemaValidationResult` is the root class for all schema validation results.
`NadelSchemaValidationError` extends from `NadelSchemaValidationResult`
We also have other `NadelSchemaValidationResult` implementations that actually return data
e.g. `NadelValidatedFieldResult` , `NadelValidatedTypeResult` and `NadelReachableServiceTypesResult` etc.

This is how we return more than just errors.

You'll also notice that we removed the `List` from here. That's because one implementation of `NadelReachableServiceTypesResult` is a wrapper around `List`.
This is done to make passing around the result easier.

lib/src/main/java/graphql/nadel/validation/NadelSchemaHydrationValidationError.kt
gnawf, line 1: Hydration errors are a mess. Tbh the other errors are a mess too.

This file tries to standardise the errors e.g. names, error messages etc.

e.g. most hydration error classes should provide at least this much information

    parentType: NadelServiceSchemaElement,
    virtualField: GraphQLFieldDefinition,
    hydration: NadelHydrationDefinition,

and if so, then the error message should be in the standard format

"Field $parentTypeName.$fieldName tried to hydrate from Query.$backingField but $reason"

lib/src/main/java/graphql/nadel/validation/NadelSchemaValidationError.kt
gnawf, line 1: Moved hydration errors to another file

lib/src/main/java/graphql/nadel/util/IteratorUtil.kt
gnawf, line 1: Pulled from PQS.

Used in NadelTypeWrappingValidation which is also pulled from PQS.

lib/src/main/java/graphql/nadel/engine/blueprint/NadelExecutionBlueprintFactory.kt
gnawf, line 1: Fluff changes like reformat & renames.

Don't think there's anything important at the time of writing this comment.

felipe-gdr (dismissed):

felipe-gdr (dismissed):

felipe-gdr (approved):

PR #616, opened 2024-11-13
Add execution hint to migrate to new blueprint

gnawf:
Adds a FF so we can switch between the new and old blueprint.


bbakerman (approved):
!!

PR #617, opened 2024-11-13
Concrete NadelServiceSchemaElement type

gnawf:
Made `NadelServiceSchemaElement` more specific so we don't have to keep casting the type.


bbakerman (dismissed):
Nice one

bbakerman (approved):

PR #619, opened 2024-11-15
Faster validation

gnawf:
So I ran this against Central Schema using dcd34ebc3ebb5baf87acd4ec885a1214a0165c57

The current validation takes 2 _seconds_. With the improvements on this branch it brings it down to 80ms.

The big majority of time saves are in precomputing info instead of computing it for each service validated etc.

For reference, the current blueprint creation method takes 30ms.

### DIFF
--- a/lib/src/main/java/graphql/nadel/NadelDefinitionRegistry.kt
+++ b/lib/src/main/java/graphql/nadel/NadelDefinitionRegistry.kt
@@ -78,6 +78,10 @@ class NadelDefinitionRegistry {
         return sdlDefinitions.filterIsInstance<T>()
     }
 
+    fun getDefinitions(name: String): List<AnySDLDefinition> {
+        return definitionsByName[name] ?: return emptyList()
+    }
+
     inline fun <reified T : AnySDLDefinition> getDefinitionsOfType(): List<T> {
         return getDefinitionsOfType(T::class.java)
     }
--- a/lib/src/main/java/graphql/nadel/engine/blueprint/NadelFastSchemaTraverser.kt
+++ b/lib/src/main/java/graphql/nadel/engine/blueprint/NadelFastSchemaTraverser.kt
@@ -1,0 +1,272 @@
+package graphql.nadel.engine.blueprint
+
+import graphql.nadel.engine.util.unwrapAll
+import graphql.schema.GraphQLAppliedDirective
+import graphql.schema.GraphQLAppliedDirectiveArgument
+import graphql.schema.GraphQLArgument
+import graphql.schema.GraphQLCompositeType
+import graphql.schema.GraphQLDirective
+import graphql.schema.GraphQLDirectiveContainer
+import graphql.schema.GraphQLEnumType
+import graphql.schema.GraphQLEnumValueDefinition
+import graphql.schema.GraphQLFieldDefinition
+import graphql.schema.GraphQLFieldsContainer
+import graphql.schema.GraphQLInputFieldsContainer
+import graphql.schema.GraphQLInputObjectField
+import graphql.schema.GraphQLInputObjectType
+import graphql.schema.GraphQLInterfaceType
+import graphql.schema.GraphQLList
+import graphql.schema.GraphQLModifiedType
+import graphql.schema.GraphQLNamedSchemaElement
+import graphql.schema.GraphQLNamedType
+import graphql.schema.GraphQLNonNull
+import graphql.schema.GraphQLNullableType
+import graphql.schema.GraphQLObjectType
+import graphql.schema.GraphQLOutputType
+import graphql.schema.GraphQLScalarType
+import graphql.schema.GraphQLSchema
+import graphql.schema.GraphQLSchemaElement
+import graphql.schema.GraphQLType
+import graphql.schema.GraphQLTypeReference
+import graphql.schema.GraphQLUnionType
+import graphql.schema.GraphQLUnmodifiedType
+
+/**
+ * Significantly faster than normal [graphql.schema.SchemaTraverser] as it's simpler.
+ *
+ * Validating central schema with old [graphql.schema.SchemaTraverser] takes 306ms and this impl takes 74ms.
+ *
+ * That's 4x faster.
+ */
+internal class NadelFastSchemaTraverser {
+    fun traverse(
+        schema: GraphQLSchema,
+        roots: List<String>,
+        visitor: Visitor,
+    ) {
+        val queue = roots
+            .mapTo(mutableListOf()) {
+                (null as GraphQLNamedSchemaElement?) to (schema.typeMap[it] ?: schema.getDirective(it))!!
+            }
+        val visited = roots.toMutableSet()
+
+        fun queueType(parent: GraphQLNamedSchemaElement, type: GraphQLNamedType) {
+            if (!visited.contains(type.name)) {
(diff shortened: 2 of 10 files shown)


lib/src/main/java/graphql/nadel/validation/NadelNamespaceValidation.kt
```diff
@@ -1,18 +1,14 @@
...
 
-class NadelNamespaceValidation(
-) {
+class NadelNamespaceValidation {
     context(NadelValidationContext)
     fun validate(
         schemaElement: NadelServiceSchemaElement,
     ): NadelSchemaValidationResult {
-        if (!isNamespacedOperationType(typeName = schemaElement.overall.name)) {
+        if (schemaElement.overall.name !in namespaceTypeNames) {
```
gnawf, line 11: `NadelNamespaceValidation` is run for every type.

The old `isNamespacedOperationType` would iterate through all root fields, which was very slow.

This procomputes it and puts that information in the `NadelValidationContext` so that we only iterate through the root fields once.

lib/src/main/java/graphql/nadel/validation/NadelFieldValidation.kt
```diff
@@ -181,6 +180,6 @@ internal class NadelFieldValidation(
 
     context(NadelValidationContext)
     private fun isCombinedType(type: GraphQLNamedSchemaElement): Boolean {
-        return isCombinedType(engineSchema, type)
+        return type.name in combinedTypeNames
```
gnawf, line 183: This is run on _every_ field.

The old `isCombinedType` would iterate through all root fields, which was very slow.

This procomputes it and puts that information in the NadelValidationContext so that we only iterate through the root fields once.

lib/src/main/java/graphql/nadel/validation/NadelSchemaValidation.kt
```diff
@@ -29,10 +37,17 @@ class NadelSchemaValidation(
...
         val context = NadelValidationContext(
-            engineSchema,
-            servicesByName,
-            fieldContributorMap,
+            engineSchema = engineSchema,
+            servicesByName = servicesByName,
+            fieldContributor = fieldContributorMap,
+            hydrationUnions = getHydrationUnions(engineSchema),
+            namespaceTypeNames = namespaceTypes,
+            combinedTypeNames = namespaceTypes + operationTypes.map { it.name },
```
gnawf, line 50: Store some more precomputed info like `hydrationUnions` , `namespaceTypeNames` and `combinedTypeNames`

lib/src/main/java/graphql/nadel/validation/NadelTypeValidation.kt
```diff
@@ -159,16 +155,14 @@ internal class NadelTypeValidation {
...
         }
 
-        return engineSchema
-            .typeMap
-            .asSequence()
-            .filter { (key) ->
-                key in namesUsed
+        return namesUsed
+            .map {
+                engineSchema.typeMap[it]!!
```
gnawf, line 160: Does the same thing, but iterates much faster.

Old method:

For each type in the schema (very large set).
Filter out the ones this service used (constant time).

New method:

For each type name the service uses (smaller set).
Look up the type (constant time).

lib/src/main/java/graphql/nadel/validation/NadelTypeValidation.kt
```diff
@@ -202,47 +196,39 @@ internal class NadelTypeValidation {
...
-                        it.fieldDefinitions
-                    }
-                    .filter {
-                        it.type.unwrapAll().name == union.name
-                    }
-                    .all(min = 1) {
-                        it.hasAppliedDirective(hydratedDirectiveDefinition.name)
-                    }
+            .filter {
+                it.name in hydrationUnions
```
gnawf, line 204: Use precomputed info

lib/src/main/java/graphql/nadel/validation/NadelTypeValidation.kt
```diff
@@ -270,19 +255,17 @@ internal class NadelTypeValidation {
             }
 
         // If it can be reached by using your service, you must own it to return it!
-        val referencedTypes = getReachableTypeNames(engineSchema, service, definitionNames + matchingImplementsNames)
-
-        return (definitionNames + referencedTypes).toSet()
+        return getReachableTypeNames(service, definitionNames + matchingImplementsNames)
```
gnawf, line 255: Turns out `getReachableTypeNames` is a superset of,

```
return (definitionNames + referencedTypes).toSet()
```

So it was useless to add them.

lib/src/main/java/graphql/nadel/validation/util/NadelCombinedTypeUtil.kt
```diff
@@ -42,10 +42,9 @@ object NadelCombinedTypeUtil {
      * type `Query` then the result is `[time]`.
      */
     fun getFieldsThatServiceContributed(service: Service, overallType: GraphQLNamedSchemaElement): Set<String> {
-        return service.definitionRegistry.definitions
+        return service.definitionRegistry.getDefinitions(overallType.name)
```
gnawf, line 41: Direct lookup `O(1)` instead of using filter `O(N)`

lib/src/main/java/graphql/nadel/validation/util/NadelGetReachableTypes.kt
```diff
@@ -267,21 +268,11 @@ internal fun getReachableTypeNames(
         }
     }
 
-    SchemaTraverser { element ->
-        // Handle combined types
-        if (element is GraphQLNamedSchemaElement && isCombinedType(overallSchema, element)) {
```
gnawf, line 272: This part was moved into `visitGraphQLFieldDefinition`

lib/src/main/java/graphql/nadel/NadelDefinitionRegistry.kt
```diff
@@ -78,6 +78,10 @@ class NadelDefinitionRegistry {
         return sdlDefinitions.filterIsInstance<T>()
     }
 
+    fun getDefinitions(name: String): List<AnySDLDefinition> {
+        return definitionsByName[name] ?: return emptyList()
```
gnawf, line 82: Allows for faster lookups.

Some code was doing `.definitions.filter { it.name == name }`

lib/src/main/java/graphql/nadel/engine/blueprint/NadelFastSchemaTraverser.kt
```diff
@@ -0,0 +1,272 @@
...
+import graphql.schema.GraphQLTypeReference
+import graphql.schema.GraphQLUnionType
+import graphql.schema.GraphQLUnmodifiedType
+
+/**
+ * Significantly faster than normal [graphql.schema.SchemaTraverser] as it's simpler.
+ *
+ * Validating central schema with old [graphql.schema.SchemaTraverser] takes 306ms and this impl takes 74ms.
+ *
+ * That's 4x faster.
```
gnawf, line 39: I reverted the changes in f7c095e318d15f39ba1094fdcfad1f697be80313 to show this.

I also reapplied relevant optimisations to the file that weren't related to the `SchemaTraverser` so it should be a pure comparison.

bbakerman (approved):

PR #622, opened 2024-11-18
Delete unused flag. It's set inside Nadel.Builder instead

gnawf:
I was reading over the PR again and noticed this is unused.

This flag is actually set globally for the entire Nadel engine at `graphql.nadel.Nadel.Builder#blueprintHint`


anton-vasylenko-atlassian (approved):
LGTM

PR #625, opened 2024-11-20
Safer migration to new blueprint

gnawf:


lib/src/main/java/graphql/nadel/NextgenEngine.kt
```diff
@@ -97,9 +98,16 @@ internal class NextgenEngine(
...
-        ).validateAndGenerateBlueprint(),
+        new = lazy {
+            try {
+                NadelSchemaValidation(
+                    NadelSchemas(engineSchema, services)
+                ).validateAndGenerateBlueprint()
+            } catch (e: Exception) {
+                getLogger<NextgenEngine>().error("Unable to create validated blueprint", e)
+                null
+            }
```
gnawf, line 109: Made it lazy, and added a try catch around it.

llin2 (approved):

PR #629, opened 2024-11-21
Add support for alias in partition transform

felipe-gdr:


lib/src/main/java/graphql/nadel/engine/transform/query/NFUtil.kt
```diff
@@ -17,6 +17,7 @@ object NFUtil {
         schema: GraphQLSchema,
         parentType: GraphQLOutputType,
         pathToField: NadelQueryPath,
+        aliasedPath: NadelQueryPath? = null,
```
felipe-gdr, line 20: Another, and probably safer, option would be to leave the existing `NFUtil.createField` as is and create another method to insert aliases to the NF field hierarchy later.
`NFUtil.createField(queryPathToField=field.fieldPath).insertAliases(field.queryPath)`
But I think having `aliasedPath` as an extra argument to the main function makes it easier to reuse in the future.

gnawf (approved):

PR #634, opened 2024-11-27
Improved schema traverser type safety

gnawf:
Should also address a potential issue with union traversal.

### DIFF
--- a/lib/src/main/java/graphql/nadel/engine/blueprint/NadelFastSchemaTraverser.kt
+++ b/lib/src/main/java/graphql/nadel/engine/blueprint/NadelFastSchemaTraverser.kt
@@ -1,35 +1,6 @@
 package graphql.nadel.engine.blueprint
 
-import graphql.nadel.engine.util.unwrapAll
-import graphql.schema.GraphQLAppliedDirective
-import graphql.schema.GraphQLAppliedDirectiveArgument
-import graphql.schema.GraphQLArgument
-import graphql.schema.GraphQLCompositeType
-import graphql.schema.GraphQLDirective
-import graphql.schema.GraphQLDirectiveContainer
-import graphql.schema.GraphQLEnumType
-import graphql.schema.GraphQLEnumValueDefinition
-import graphql.schema.GraphQLFieldDefinition
-import graphql.schema.GraphQLFieldsContainer
-import graphql.schema.GraphQLInputFieldsContainer
-import graphql.schema.GraphQLInputObjectField
-import graphql.schema.GraphQLInputObjectType
-import graphql.schema.GraphQLInterfaceType
-import graphql.schema.GraphQLList
-import graphql.schema.GraphQLModifiedType
-import graphql.schema.GraphQLNamedSchemaElement
-import graphql.schema.GraphQLNamedType
-import graphql.schema.GraphQLNonNull
-import graphql.schema.GraphQLNullableType
-import graphql.schema.GraphQLObjectType
-import graphql.schema.GraphQLOutputType
-import graphql.schema.GraphQLScalarType
 import graphql.schema.GraphQLSchema
-import graphql.schema.GraphQLSchemaElement
-import graphql.schema.GraphQLType
-import graphql.schema.GraphQLTypeReference
-import graphql.schema.GraphQLUnionType
-import graphql.schema.GraphQLUnmodifiedType
 
 /**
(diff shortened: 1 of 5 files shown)


lib/src/main/java/graphql/nadel/engine/blueprint/NadelFastSchemaTraverser.kt
```diff
@@ -38,242 +9,89 @@ import graphql.schema.GraphQLUnmodifiedType
...
+internal class NadelSchemaTraverser {
     fun traverse(
         schema: GraphQLSchema,
-        roots: List<String>,
-        visitor: Visitor,
+        roots: Iterable<String>,
+        visitor: NadelSchemaTraverserVisitor,
     ) {
-        val queue = roots
+        val queue: MutableList<NadelSchemaTraverserElement> = roots
```
gnawf, line 18: We aren't making any more objects than before.

The old type was `List<Pair<GraphQLNamedSchemaElement, GraphQLNamedType>>` i.e. we are replacing `Pair` with our own `NadelSchemaTraverserElement`

lib/src/main/java/graphql/nadel/engine/blueprint/NadelFastSchemaTraverser.kt
```diff
@@ -38,242 +9,89 @@ import graphql.schema.GraphQLUnmodifiedType
...
-                    }
-                    is GraphQLNamedSchemaElement -> {
-                        queueElement(parent = element, child)
-                    }
-                }
-            }
-        }
-    }
-
-    interface Visitor {
```
gnawf, line 107: Extracted to top level class

lib/src/main/java/graphql/nadel/engine/blueprint/NadelSchemaTraverserElement.kt
```diff
@@ -0,0 +1,247 @@
...
+import graphql.schema.GraphQLInputType
+import graphql.schema.GraphQLInterfaceType
+import graphql.schema.GraphQLNamedSchemaElement
+import graphql.schema.GraphQLNamedType
+import graphql.schema.GraphQLObjectType
+import graphql.schema.GraphQLOutputType
+import graphql.schema.GraphQLScalarType
+import graphql.schema.GraphQLUnionType
+
+internal sealed interface NadelSchemaTraverserElement {
```
gnawf, line 24: Model for traversals, instead of having

```kotlin
        fun visitGraphQLScalarType(
            parent: GraphQLNamedSchemaElement?,
            node: GraphQLScalarType,
        ): Boolean
```

we have

```kotlin
        fun visitGraphQLScalarType(
            element: NadelSchemaTraverserElement.Scalar
        ): Boolean
```

Where it is known that `Scalar` type doesn't have a parent.

Another e.g.

```kotlin
    data class FieldDefinition(
        val parent: GraphQLFieldsContainer,
        override val node: GraphQLFieldDefinition,
    ) : NadelSchemaTraverserElement
```

It's better modelled here that a field has a parent field container.

lib/src/main/java/graphql/nadel/validation/util/NadelGetReachableTypes.kt
```diff
@@ -190,159 +176,68 @@ private class NadelReferencedTypeVisitor(
...
-        parent: GraphQLNamedSchemaElement?,
-        node: GraphQLDirective,
+        element: NadelSchemaTraverserElement.Directive,
     ): Boolean {
         // We don't care about directives, not a Nadel runtime concern
         // GraphQL Java will do validation on them for us
         return false
     }
 
-    override fun visitGraphQLDirectiveContainer(
```
gnawf, line 282: These were never actually invoked, the fast schema traverser skips these as we don't need them.

lib/src/test/kotlin/graphql/nadel/validation/NadelUnionValidationTest.kt
```diff
@@ -304,7 +304,7 @@ class NadelUnionValidationTest {
                     type Query {
                         entity(id: ID!): Entity
                     }
-                    union Entity = Human | Dog
+                    union Entity = Human | Dog | Issue
```
gnawf, line 307: Adding this in to make the fixture more valid.

The test is to check that we run validation on union member types. It was missing `Issue` too, but that's not what we're testing.

bbakerman (approved):

PR #636, opened 2024-11-27
Treat some GraphQL Java types as pseudo sealed types

gnawf:


bbakerman (approved):
for the love of types

PR #637, opened 2024-11-27
Shadow service typename filter

gnawf:


Juliano-Prado:
Code look good to me but I missed some test showing side effect of when isShadowingEnabled.

Juliano-Prado (approved):
Code look good to me but I missed some test showing side effect of when isShadowingEnabled.

lib/src/main/java/graphql/nadel/hints/NadelServiceTypenameShadowingMismatchHint.kt
```diff
@@ -0,0 +1,33 @@
...
+            }
+
+            override fun onMismatch(
+                oldDecision: Boolean,
+                newDecision: Boolean,
+                service: Service,
+                objectTypeName: String,
+                field: ExecutableNormalizedField,
+            ) {
+            }
```
Juliano-Prado, line 30: is it right to be empty}

Juliano-Prado (commented):
Code look good to me but I missed some test showing side effect of when isShadowingEnabled.

lib/src/main/java/graphql/nadel/hints/NadelServiceTypenameShadowingMismatchHint.kt
```diff
@@ -0,0 +1,33 @@
...
+        field: ExecutableNormalizedField,
+    )
+
+    companion object {
+        internal val default = object : NadelServiceTypenameShadowingHint {
+            override fun isShadowingEnabled(): Boolean {
+                return false
+            }
+
+            override fun onMismatch(
```
Juliano-Prado, line 23: default implementation is empty, I got lost on how is setting behaviour when it is called on [NadelServiceTypeFilterTransform.kt](https://github.com/atlassian-labs/nadel/pull/637/files#diff-ef65360fa390e6d9ac0bf031a350bd2c8138040cf3664acfc1a9d98da8f1107d)

gnawf: The hint is provided by AGG. So the impl will be logging the mismatch.

PR #649, opened 2024-12-05
return empty response when subscription are running on the internal introspection service

temaEmelyan:

### DIFF
--- a/lib/src/main/java/graphql/nadel/engine/blueprint/IntrospectionService.kt
+++ b/lib/src/main/java/graphql/nadel/engine/blueprint/IntrospectionService.kt
@@ -4,12 +4,13 @@ import graphql.ExecutionInput
 import graphql.GraphQL
 import graphql.GraphqlErrorHelper.toSpecification
 import graphql.language.AstPrinter
+import graphql.language.OperationDefinition
 import graphql.nadel.NadelDefinitionRegistry
+import graphql.nadel.NadelServiceExecutionResultImpl
 import graphql.nadel.Service
 import graphql.nadel.ServiceExecution
 import graphql.nadel.ServiceExecutionParameters
 import graphql.nadel.ServiceExecutionResult
-import graphql.nadel.NadelServiceExecutionResultImpl
 import graphql.nadel.engine.util.makeFieldCoordinates
 import graphql.nadel.engine.util.toBuilder
 import graphql.nadel.engine.util.toBuilderWithoutTypes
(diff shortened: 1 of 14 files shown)


test/src/test/resources/fixtures/field removed/top-level-field-is-removed-for-a-subscription-with-namespaced-field.yml
```diff
@@ -0,0 +1,60 @@
...
+serviceCalls: [ ]
+# language=JSON
+response: |-
+  {
+    "errors": [
+      {
+        "locations": [],
+        "message": "null",
+        "extensions": {
+          "classification": "ValidationError"
```
temaEmelyan, line 53: this error is coming from a "remove field" transform that we use for tests. I activated this transform with `onCommentUpdated(id: ID): Comment @toBeDeleted` directive in the overall schema

temaEmelyan: previously we would blow up with `your data fetcher must return a publisher ...` kind of an error

lib/src/main/java/graphql/nadel/engine/blueprint/IntrospectionService.kt
```diff
@@ -40,6 +41,9 @@ open class NadelDefaultIntrospectionRunner(schema: GraphQLSchema) : ServiceExecu
         .build()
 
     override fun execute(serviceExecutionParameters: ServiceExecutionParameters): CompletableFuture<ServiceExecutionResult> {
+        if (serviceExecutionParameters.operationDefinition.operation == OperationDefinition.Operation.SUBSCRIPTION) {
+            return CompletableFuture.completedFuture(NadelServiceExecutionResultImpl())
```
russellyou, line 45: So just empty result?

russellyou (approved):

bbakerman (dismissed):

test/src/test/resources/fixtures/field removed/top-level-field-is-removed-for-a-subscription-with-namespaced-field.yml
gnawf, line 53: So this is an error from the introspection service? Why is the error message null?

temaEmelyan: nah, this error is coming from the transform, look at this comment https://github.com/atlassian-labs/nadel/pull/649/files/5774cd1580b7dcaeaecb6b6b7d46bdea52b078eb#r1870548215

gnawf: Ah I see

gnawf (approved):

PR #650, opened 2024-12-05
Return partial result on hydration exception

gnawf:


test/src/test/kotlin/graphql/nadel/tests/next/fixtures/hydration/HydrationThrowsErrorTestSnapshot.kt
```diff
@@ -0,0 +1,100 @@
...
+            |       "extensions": {
+            |         "classification": "Bye"
+            |       }
+            |     }
+            |   ],
+            |   "data": {
+            |     "issueById": {
+            |       "id": "ari:cloud:jira:19b8272f-8d25-4706-adce-8db72305e615:issue/1",
+            |       "key": "GQLGW-1",
+            |       "assignee": null
```
gnawf, line 92: So this part is `null` but before this change the entire `issueById` field was null

test/src/test/kotlin/graphql/nadel/tests/next/fixtures/hydration/statics/StaticHydrationNestedErrorTestSnapshot.kt
```diff
@@ -0,0 +1,134 @@
...
+            |     }
+            |   ],
+            |   "data": {
+            |     "businessReport_findRecentWorkByTeam": {
+            |       "__typename": "WorkConnection",
+            |       "edges": [
+            |         {
+            |           "__typename": "WorkEdge",
+            |           "cursor": "1",
+            |           "node": null
```
gnawf, line 120: Same here, before the fix `businessReport_findRecentWorkByTeam` was null.

lib/src/main/java/graphql/nadel/NextgenEngine.kt
```diff
@@ -265,13 +265,23 @@ internal class NextgenEngine(
...
+            )
+        } catch (e: Exception) {
+            when (e) {
+                is GraphQLError -> newServiceExecutionErrorResult(
+                    field = topLevelField,
+                    error = e,
+                )
+                else -> throw e
+            }
+        }
```
gnawf, line 284: @felipe-gdr I left the partitioned stuff untouched.

I think it's better to not handle it? The way I stumbled across this was I fed an invalid ARI into hydration, then we raised a `InvalidAriException`

I guess for partitioned stuff if there's an invalid ARI we should just stop the entire call?

bbakerman (approved):

PR #656, opened 2024-12-13
Delete old blueprint code

gnawf:
New blueprint is basically launched. Next's merge this some time next week.


Juliano-Prado (approved):
code wise looks good to me, although I dont have deep knowledge on the old function - such as usage of hint for example.

PR #657, opened 2024-12-13
Delete shared type rename patch as new blueprint handles it properly

gnawf:
Relies on #656


bbakerman (approved):
Will this bugger up your test migration PR ?

PR #669, opened 2025-01-28
Fix ID hydration for User types in union

gnawf:
So the issue arises when an abstract type `User` is fulfilled by one field `users(ids: [ID!]!): [User]` and its concrete implementations are referenced in a union e.g.

```graphql
type Query {
  users(ids: [ID!]!): [User]
}

type Issue {
  userId: ID @hidden
  user: IssueUser @idHydrated(idField: "userId")
}
union IssueUser = CustomerUser | AtlassianAccountUser | AppUser

interface User {
  id: ID!
}
type CustomerUser implements User @defaultHydration(field: "users", idArgument: "ids") {
  id: ID!
}
type AtlassianAccountUser implements User @defaultHydration(field: "users", idArgument: "ids") {
  id: ID!
}
type AppUser implements User @defaultHydration(field: "users", idArgument: "ids") {
  id: ID!
}
```

In the previous implementation this would result in three hydration definitions on `Issue.user` but with this fix now it's deduplicated.

There's also validation that ensures that types backed by the same field have the same `@defaultHydration` config, otherwise the hydration configuration would be ambiguous.

### DIFF
--- a/lib/src/main/java/graphql/nadel/validation/NadelIdHydrationDefinitionParser.kt
+++ b/lib/src/main/java/graphql/nadel/validation/NadelIdHydrationDefinitionParser.kt
@@ -9,6 +9,7 @@ import graphql.nadel.definition.hydration.NadelIdHydrationDefinition
 import graphql.nadel.definition.hydration.parseDefaultHydrationOrNull
 import graphql.nadel.definition.hydration.parseIdHydrationOrNull
 import graphql.nadel.engine.util.unwrapAll
+import graphql.nadel.validation.NadelValidationInterimResult.Error.Companion.asInterimError
 import graphql.schema.GraphQLFieldDefinition
 import graphql.schema.GraphQLFieldsContainer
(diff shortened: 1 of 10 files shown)


lib/src/main/java/graphql/nadel/validation/NadelIdHydrationDefinitionParser.kt
```diff
@@ -58,21 +59,30 @@ internal class NadelIdHydrationDefinitionParser {
...
+                        idHydration,
+                        unionMemberType as GraphQLObjectType,
+                    ).onErrorCast { return it }
+                }
+            }
+
+        val uniqueHydrations = hydrations
+            .groupBy { it.backingField }
+            .map { (backingField, hydrations) ->
+                hydrations.toSet().singleOrNull()
```
gnawf, line 79: The fix is here

temaEmelyan (approved):

PR #672, opened 2025-05-15
Introduce ability to cast hydration source ID to match result ID

gnawf:


lib/src/main/java/graphql/nadel/validation/hydration/NadelHydrationValidation.kt
```diff
@@ -511,6 +533,43 @@ class NadelHydrationValidation internal constructor(
...
+        } else {
+            NadelObjectIdentifierCastingStrategy.NO_CAST
+        }
+    }
+
+    context(NadelValidationContext)
+    private fun getResultIdFieldOrNull(
+        hydration: NadelHydrationDefinition,
+        resultIdFieldName: String,
+    ): GraphQLFieldDefinition? {
```
gnawf, line 554: This _should_ be non-null but it's a bit of a can of worms I'm not willing to deal with right now.

test/src/test/kotlin/graphql/nadel/tests/next/fixtures/hydration/CastHydrationLongInputToStringForMatchingTestSnapshot.kt
```diff
@@ -0,0 +1,132 @@
...
+            |       "spaces": [
+            |         {
+            |           "id": "1"
+            |         },
+            |         {
+            |           "id": "2"
+            |         },
+            |         {
+            |           "id": "10"
+            |         }
```
gnawf, line 123: Before this would've been

```
"spaces": [null, null  nul]
```

The fix means the integer input now matches the result ID (String).

bbakerman (commented):
Nice

PR #678, opened 2025-05-19
Split publish from build + test so we can more easily read logs

gnawf:
This will make it easier for us to read the Publish logs in Github UI since individual steps can be collapsed.


lib/src/main/java/graphql/nadel/definition/coordinates/NadelInputCoordinates.kt
```diff
@@ -1,3 +1,3 @@
 package graphql.nadel.definition.coordinates
 
-sealed interface NadelInputCoordinates : NadelSchemaMemberCoordinates
+sealed interface NadelInputCoordinates : NadelChildCoordinates
```
gnawf, line 3: Piggy backing off this PR to get this one liner in 😄

sbarker2 (approved):

PR #682, opened 2025-06-04
Fix coords toString

gnawf:


gnawf:
Broke because I fixed the parent sequence in #680 but this relied on the broken behavior.

PR #695, opened 2025-07-29
clean up for adding onComplete and buildContext

temaEmelyan:
clean up based on the suggestions from the previous PR
https://github.com/atlassian-labs/nadel/pull/693#issuecomment-3116415881


lib/src/main/java/graphql/nadel/engine/plan/NadelExecutionPlan.kt
```diff
@@ -19,21 +20,5 @@ data class NadelExecutionPlan(
...
         val resultTransformTimingStep: NadelInstrumentationTimingParameters.Step,
         val state: T,
-        val transformServiceExecutionContext: NadelTransformServiceExecutionContext?,
     )
-
-    /**
-     * Creates and returns a new [NadelExecutionPlan] that is a merging of `this` plan
-     * and the [other] plan.
-     */
-    fun merge(other: NadelExecutionPlan): NadelExecutionPlan {
```
temaEmelyan, line 29: I don't think we ever use it and I don't think we can do it now that we store a context by transform

gnawf (approved):

PR #697, opened 2025-08-11
Fix short circuit introspection for namespace like fields

gnawf:


lib/src/main/java/graphql/nadel/NextgenEngine.kt
```diff
@@ -479,16 +482,13 @@ internal class NextgenEngine(
     ): Boolean {
         val topLevelField = topLevelFields.singleOrNull() ?: return false
 
-        if (topLevelField.fieldName == TypeNameMetaFieldDef.name) {
+        if (topLevelField.name == TypeNameMetaFieldDef.name) {
             return true
         }
-        val operationType = service.underlyingSchema.getTypeAs<GraphQLObjectType>(topLevelField.singleObjectTypeName)
-        val topLevelFieldDefinition = operationType.getField(topLevelField.name)
-        val isNamespacedLike = topLevelFieldDefinition?.arguments?.isEmpty() == true
```
gnawf, line 487: Extracted this code into common `NamespacedUtil.isNamespaceFieldLike`

gnawf, line 487: So the bug here is that the logic we use to identify "namespaced fields" is different here and in the internal introspection service.

This leads to a bug because this code thinks more fields can be handled by the internal introspection service, but in reality it can't so it fails.

lib/src/main/java/graphql/nadel/engine/blueprint/IntrospectionService.kt
```diff
@@ -70,7 +71,7 @@ open class NadelDefaultIntrospectionRunner(schema: GraphQLSchema) : ServiceExecu
                     // This inserts data fetchers for namespaced fields so we can handle their __typename internally
                     getFieldsWithCoordinates(schema.queryType, schema.mutationType, schema.subscriptionType)
                         .filter { (_, field) ->
-                            isNamespacedField(field)
+                            isNamespacedFieldLike(field)
```
gnawf, line 74: So this uses the newly extracted code from `NextgenEngine.isOnlyTopLevelFieldTypename`

test/src/test/kotlin/graphql/nadel/tests/next/fixtures/introspection/NamespaceLikeShortCircuitsIntrospectionTestSnapshot.kt
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * This class is generated. Do NOT modify.
+ *
+ * Refer to [graphql.nadel.tests.next.UpdateTestSnapshots]
+ */
+@Suppress("unused")
+public class NamespaceLikeShortCircuitsIntrospectionTestSnapshot : TestSnapshot() {
+    override val calls: List<ExpectedServiceCall> = listOf(
+            )
```
gnawf, line 24: So here the calls are empty because the tests are executed by our internal introspection service.

test/src/test/kotlin/graphql/nadel/tests/next/fixtures/introspection/NamespaceLikeShortCircuitsIntrospectionTestSnapshot.kt
```diff
@@ -0,0 +1,50 @@
...
+     *   }
+     * }
+     * ```
+     */
+    override val result: ExpectedNadelResult = ExpectedNadelResult(
+            result = """
+            | {
+            |   "data": {
+            |     "test": {
+            |       "__typename": "TestQuery"
```
gnawf, line 42: Prior to this PR's fix, it would error out like so

```json
{
  "errors": [
    {
      "message": "Exception while fetching data (/test) : Internal error: should never happen: This data fetcher should NEVER be called from Nadel",
      "locations": [
        {
          "column": 2,
          "line": 1
        }
      ],
      "path": [
        "test"
      ],
      "extensions": {
        "classification": "DataFetchingException"
      }
    }
  ],
  "data": {
    "test": null
  }
}
```

test/src/test/kotlin/graphql/nadel/tests/next/fixtures/stub/StubNamespaceLikeMutationFieldTest.kt
gnawf, line 1: Adding some stubbed tests because that's what led to this bug.

bbakerman (approved):

PR #718, opened 2026-03-12
Schema definition transformer

gnawf:
The existing `FieldVisibilitySchemaTransformation` operates on `GraphQLSchema` and is very slow and consumes a lot of memory.

This PR adds a new capability to transform `SDLDefinition` to let us save on time and memory.

The equivalent field visibility implementation is in `NadelFieldDefinitionVisibilityTransformation`

This allows our CLI to run validation with less memory from `-Xmx850m` to `-Xmx580m` so decrease of 270MB.

It is significantly faster to transform Central Schema now.

```
Took 246ms to transform DEFINITIONS to hide lifecycle fields
Took 11638ms to transform SCHEMA to hide lifecycle fields
```

See test done in 515b78623108d113fb4f1ebf5ef57875a2e98272

### DIFF
--- a/lib/src/main/java/graphql/nadel/schema/NadelFieldDefinitionVisibilityTransformation.kt
+++ b/lib/src/main/java/graphql/nadel/schema/NadelFieldDefinitionVisibilityTransformation.kt
@@ -15,6 +15,9 @@ import graphql.schema.idl.DirectiveInfo
 import graphql.schema.idl.ScalarInfo
 
 fun interface NadelFieldDefinitionVisibilityTransformationPredicate {
+    /**
+     * @return `true` to keep field and `false` to delete
+     */
     operator fun invoke(
         parent: ImplementingTypeDefinition<*>,
         field: FieldDefinition,
--- a/lib/src/main/java/graphql/nadel/schema/NadelFieldDefinitionVisibilityTransformation.kt
+++ b/lib/src/main/java/graphql/nadel/schema/NadelFieldDefinitionVisibilityTransformation.kt
@@ -1,0 +1,237 @@
+package graphql.nadel.schema
+
+import graphql.language.FieldDefinition
+import graphql.language.ImplementingTypeDefinition
+import graphql.language.InterfaceTypeDefinition
+import graphql.language.InterfaceTypeExtensionDefinition
+import graphql.language.ObjectTypeDefinition
+import graphql.language.ObjectTypeExtensionDefinition
+import graphql.nadel.NadelOperationKind
+import graphql.nadel.engine.util.unwrapAll
+import graphql.nadel.util.AnyImplementingTypeDefinition
+import graphql.nadel.util.AnyNamedNode
+import graphql.nadel.util.AnySDLDefinition
+import graphql.schema.idl.DirectiveInfo
+import graphql.schema.idl.ScalarInfo
+
+fun interface NadelFieldDefinitionVisibilityTransformationPredicate {
+    operator fun invoke(
+        parent: ImplementingTypeDefinition<*>,
+        field: FieldDefinition,
+    ): Boolean
+}
+
+class NadelFieldDefinitionVisibilityTransformation(
+    val fieldPredicate: NadelFieldDefinitionVisibilityTransformationPredicate,
+) : NadelSchemaDefinitionTransformationHook {
+    override fun apply(definitions: List<AnySDLDefinition>): List<AnySDLDefinition> {
+        return deleteFields(definitions)
+    }
+
+    fun deleteFields(definitions: List<AnySDLDefinition>): List<AnySDLDefinition> {
+        val newDefinitions = definitions
+            .map { definition ->
+                when (definition) {
+                    is InterfaceTypeExtensionDefinition -> transformExtensionType(definition)
+                    is InterfaceTypeDefinition -> transformType(definition)
+                    is ObjectTypeExtensionDefinition -> transformExtensionType(definition)
+                    is ObjectTypeDefinition -> transformType(definition)
+                    else -> definition
+                }
+            }
+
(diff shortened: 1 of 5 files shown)


lib/src/main/java/graphql/nadel/schema/NadelFieldDefinitionVisibilityTransformation.kt
```diff
@@ -0,0 +1,237 @@
...
+import graphql.nadel.util.AnyNamedNode
+import graphql.nadel.util.AnySDLDefinition
+import graphql.schema.idl.DirectiveInfo
+import graphql.schema.idl.ScalarInfo
+
+fun interface NadelFieldDefinitionVisibilityTransformationPredicate {
+    operator fun invoke(
+        parent: ImplementingTypeDefinition<*>,
+        field: FieldDefinition,
+    ): Boolean
```
gnawf, line 24: Interface for letting consumers decide whether to keep field or not

lib/src/main/java/graphql/nadel/schema/NadelFieldDefinitionVisibilityTransformation.kt
```diff
@@ -0,0 +1,237 @@
...
+                    is InterfaceTypeExtensionDefinition -> transformExtensionType(definition)
+                    is InterfaceTypeDefinition -> transformType(definition)
+                    is ObjectTypeExtensionDefinition -> transformExtensionType(definition)
+                    is ObjectTypeDefinition -> transformType(definition)
+                    else -> definition
+                }
+            }
+
+        val observedBeforeTransform = getStronglyReferencedTypes(definitions)
+        val observedAfterTransform = getStronglyReferencedTypes(newDefinitions)
```
gnawf, line 44: This is how it works in the GraphQL Java visibility transformer. Mimicking it here so our new impl is as close to a drop in replacement as possible.

lib/src/main/java/graphql/nadel/schema/NadelFieldDefinitionVisibilityTransformation.kt
```diff
@@ -0,0 +1,237 @@
...
+    private fun filterFields(
+        parent: ImplementingTypeDefinition<*>,
+    ): List<FieldDefinition> {
+        return parent.fieldDefinitions
+            .filter { field ->
+                fieldPredicate(parent, field)
+            }
+    }
+
+    private fun getStronglyReferencedTypes(types: List<AnySDLDefinition>): Set<String> {
```
gnawf, line 113: Might belong in separate class but don't see this being reused atm.

It's a bit custom for this use case.

lib/src/main/java/graphql/nadel/schema/OverallSchemaGenerator.kt
```diff
@@ -24,25 +24,30 @@ internal class OverallSchemaGenerator {
...
+                schemaDefinitionTransformationHook.invoke(it)
+            }
+
+        val typeRegistry = createTypeRegistry(definitions)
+
+        return schemaGenerator.makeExecutableSchema(typeRegistry, runtimeWiring)
     }
 
-    private fun createTypeRegistry(serviceRegistries: List<NadelTypeDefinitionRegistry>): TypeDefinitionRegistry {
+    private fun getDefinitions(serviceRegistries: List<NadelTypeDefinitionRegistry>): List<AnySDLDefinition> {
```
gnawf, line 45: Changed this to return `List<SDLDefinition<*>>` instead of `TypeDefinitionRegistry`

Easier to work with the list than `TypeDefinitionRegistry`

We run transform on `List` then create `TypeDefinitionRegistry` afterwards.

lib/src/main/java/graphql/nadel/schema/OverallSchemaGenerator.kt
```diff
@@ -24,25 +24,30 @@ internal class OverallSchemaGenerator {
...
-            .codeRegistry(
-                NeverWiringFactory.NEVER_CODE_REGISTRY
-            )
+            .codeRegistry(NeverWiringFactory.NEVER_CODE_REGISTRY)
             .build()
 
-        return schemaGenerator.makeExecutableSchema(createTypeRegistry(serviceRegistries), runtimeWiring)
+        val definitions = getDefinitions(serviceRegistries)
+            .let {
+                schemaDefinitionTransformationHook.invoke(it)
```
gnawf, line 37: So here's where we transform the definitions.

Before

`createTypeRegistry(serivceRegistries)` → `makeExecutableSchema(typeRegistry)`

After

`getDefinitions(serviceRegistries)` → `schemaDefinitionTransformationHook(definitions)` → `createTypeRegistry(transformedDefinitions)` → `makeExecutableSchema(typeRegistry)`

lib/src/main/java/graphql/nadel/schema/NadelFieldDefinitionVisibilityTransformation.kt
```diff
@@ -0,0 +1,237 @@
...
+fun interface NadelFieldDefinitionVisibilityTransformationPredicate {
+    operator fun invoke(
+        parent: ImplementingTypeDefinition<*>,
+        field: FieldDefinition,
+    ): Boolean
+}
+
+class NadelFieldDefinitionVisibilityTransformation(
+    val fieldPredicate: NadelFieldDefinitionVisibilityTransformationPredicate,
+) : NadelSchemaDefinitionTransformationHook {
```
gnawf, line 26: Equivalent of GraphQL Java's `FieldVisibilitySchemaTransformation` but for SDL definitions.

Logic is more or less mirrored.

There are some slight differences e.g. we keep more interfaces if they implement types that are strongly referenced.

bxu3-art (approved):

PR #719, opened 2026-03-12
Add schema definition traverser

gnawf:
Clone of existing `NadelSchemaTraverser` but now for SDL definitions.

PR #720, opened 2026-03-13
Add coordinate factory method for SDL definitions

gnawf:
Depends on #719 

Allows us to read SDL definitions that are semantically invalid but syntactically valid.

### DIFF
--- a/lib/src/main/java/graphql/nadel/definition/coordinates/NadelSchemaMemberCoordinatesFactory.kt
+++ b/lib/src/main/java/graphql/nadel/definition/coordinates/NadelSchemaMemberCoordinatesFactory.kt
@@ -1,5 +1,12 @@
 package graphql.nadel.definition.coordinates
 
+import graphql.language.Document
+import graphql.language.NamedNode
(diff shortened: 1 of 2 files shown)


lib/src/main/java/graphql/nadel/definition/coordinates/NadelSchemaMemberCoordinatesFactory.kt
```diff
@@ -61,117 +88,265 @@ class NadelSchemaMemberCoordinatesFactory {
...
-                    }
-
-                    override fun visitGraphQLAppliedDirectiveArgument(
-                        element: NadelSchemaTraverserElement.AppliedDirectiveArgument,
-                    ): Boolean {
-                        coordinates.add(element.coordinates())
-                        return true
-                    }
-                }
+                NadelSchemaCoordinateCollectorTraverserVisitor(coordinates),
```
gnawf, line 91: Extracted existing code to separate class

bbakerman (approved):

PR #723, opened 2026-04-28
Support enum fields in conditional hydration

mcaliskan2:
## Summary

Allow GraphQL `enum` types as the `sourceField` in `@hydrated(when: { result: { ... } })` conditions, alongside the existing `String`/`Int`/`ID` scalars.

### What's in scope
- `equals` predicate accepts enum-typed condition fields. The supplied value must match a defined enum value (case-sensitive).
- `startsWith` / `matches` continue to require `String`/`ID` — enums are rejected with the existing predicate-specific errors.
- New `NadelHydrationConditionInvalidEnumValueError` raised when `equals` references a value not in the enum.
- `NadelHydrationResultConditionUnsupportedFieldTypeError` message updated to mention `enum`

### Implementation notes
- Internal sealed `NadelConditionFieldType` introduced so per-predicate type checks stay readable as more variants get added.

### Tests
- Validation: nullable / non-null / list / renamed enums, invalid enum value, case-sensitivity, empty string, `startsWith` + `matches` rejection.
- Integration fixtures: single conditional hydration, polymorphic (stacked `@hydrated`), and batched polymorphic


lib/src/main/java/graphql/nadel/validation/hydration/NadelHydrationConditionValidation.kt
```diff
@@ -40,6 +42,28 @@ private data class NadelHydrationConditionValidationContext(
     val condition: NadelHydrationConditionDefinition,
 )
 
+private sealed class NadelConditionFieldType {
```
gnawf, line 45: Nice

lib/src/main/java/graphql/nadel/validation/hydration/NadelHydrationConditionValidation.kt
```diff
@@ -256,16 +274,32 @@ internal class NadelHydrationConditionValidation {
...
     ): NadelValidationInterimResult<NadelHydrationCondition> {
-        if (expectedValue is String && (conditionFieldType == GraphQLString || conditionFieldType == GraphQLID)) {
+        if (expectedValue is String && (conditionFieldType == NadelConditionFieldType.StringType || conditionFieldType == NadelConditionFieldType.IdType)) {
+            return NadelHydrationCondition.StringResultEquals(
+                fieldPath = NadelQueryPath(resultCondition.pathToSourceField),
+                value = expectedValue,
+            ).asInterimSuccess()
+        } else if (expectedValue is String && conditionFieldType is NadelConditionFieldType.EnumType) {
+            if (conditionFieldType.enumType.getValue(expectedValue) == null) {
+                return NadelHydrationConditionInvalidEnumValueError(
```
gnawf, line 288: Nice

gnawf:
Looks good. Need to convert the `.yml` tests to `NadelIntegrationTest` but good otherwise. I'll do that and merge it later today.

gnawf (approved):
Approved. I updated the `.yml` tests to the new `NadelIntegrationTest`

mcaliskan2:
Thank you @gnawf 🙂

PR #724, opened 2026-04-28
Support Enum into String for hydration

AlexandreCarlton:
In GraphQL, `enum` types are serialised into `String` - as such, it is reasonable to expect that we can supply an `enum` type to a field expecting a `String` as input.

See also https://github.com/atlassian-labs/nadel/pull/723 which  lays the groundwork for conditional hydration (which would be built off in a subsequent PR).


AlexandreCarlton:
I can't seem to sign the CLA - I'm guessing because I must have already signed it to contribute to `graphql-java` and `java-dataloader`?

image

gnawf:
Looks good, I think an integration test would be good too. I'll add one and merge later today.

PR #732, opened 2026-07-02
don't expand abstract types to object fragments where possible

temaEmelyan:

### DIFF
--- a/lib/src/main/java/graphql/nadel/engine/transform/NadelNoInterfaceToObjectFragmentExpansionTransform.kt
+++ b/lib/src/main/java/graphql/nadel/engine/transform/NadelNoInterfaceToObjectFragmentExpansionTransform.kt
@@ -1,0 +1,231 @@
+package graphql.nadel.engine.transform
+
+import graphql.introspection.Introspection
+import graphql.introspection.Introspection.TypeNameMetaFieldDef
+import graphql.nadel.Service
+import graphql.nadel.ServiceExecutionHydrationDetails
+import graphql.nadel.ServiceExecutionResult
+import graphql.nadel.engine.NadelExecutionContext
+import graphql.nadel.engine.NadelServiceExecutionContext
+import graphql.nadel.engine.blueprint.IntrospectionService
+import graphql.nadel.engine.blueprint.NadelFieldInstruction
+import graphql.nadel.engine.blueprint.NadelOverallExecutionBlueprint
+import graphql.nadel.engine.transform.NadelNoInterfaceToObjectFragmentExpansionTransform.State
+import graphql.nadel.engine.transform.artificial.NadelAliasHelper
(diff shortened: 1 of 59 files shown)


lib/src/main/java/graphql/nadel/engine/transform/NadelNoInterfaceToObjectFragmentExpansionTransform.kt
```diff
@@ -0,0 +1,215 @@
...
+        if (overallField.deferredExecutions.isNotEmpty()) {
+            return null
+        }
+        val parent = overallField.parent ?: return null
+
+        val exposedOverallImplNames = computeExposedImplNamesIfRelaxable(executionBlueprint, service, overallField)
+            ?: return null
+
+        // Do nothing for fields with instructions (rename/hydration/stub/…)
+        val hasFieldInstructions = runCatching {
```
gnawf: Can we rewrite this to avoid expecting an exception and using `runCatching` to catch it.

lib/src/main/java/graphql/nadel/engine/transform/NadelNoInterfaceToObjectFragmentExpansionTransform.kt
```diff
@@ -0,0 +1,215 @@
...
+    service: Service,
+    overallField: ExecutableNormalizedField,
+): Set<String>? {
+    val parent = overallField.parent ?: return null
+
+    val engineSchema = executionBlueprint.engineSchema
+    val isTypename = overallField.fieldName == TypeNameMetaFieldDef.name
+
+    // getFieldDefinitions can throw while planning a hydration backing query; any failure => not a candidate.
+    val parentOutputTypes = runCatching {
```
gnawf: Same as before, shouldn't rely on try catch here. If the type doesn't exist in `engineSchema` then it should be handled without throwing.

lib/src/main/java/graphql/nadel/engine/transform/NadelNoInterfaceToObjectFragmentExpansionTransform.kt
```diff
@@ -0,0 +1,215 @@
...
+            .asSequence()
+            .map { unwrapAll(it.type) }
+            .toSet()
+    }.getOrNull() ?: return null
+
+    // Exposed members, gated on the field being selectable at that level (union => __typename only).
+    val exposedOverallImplNames: Set<String>
+    val parentAbstractTypeName: String
+    when (val parentType = parentOutputTypes.singleOrNull()) {
+        is GraphQLInterfaceType -> {
```
gnawf: In theory if there are multiple output interfaces and the field exists on all of them that works too.

gnawf (commented):
Looks good so far.

gnawf (approved):

PR #733, opened 2026-07-06
Batching of Root Fields in AGG

rboustany-crypto:

### DIFF
--- a/lib/src/main/java/graphql/nadel/NextgenEngine.kt
+++ b/lib/src/main/java/graphql/nadel/NextgenEngine.kt
@@ -201,18 +201,18 @@ internal class NextgenEngine(
                 val fields = fieldToService.getServicesForTopLevelFields(operation, executionHints)
                 val results = coroutineScope {
                     fields
-                        .map { (field, service) ->
+                        .map { (fields, service) ->
                             async {
                                 try {
-                                    val resolvedService = fieldToService.resolveDynamicService(field, service)
+                                    val resolvedService = fieldToService.resolveDynamicService(fields, service)
                                     executeTopLevelField(
-                                        topLevelField = field,
+                                        topLevelField = fields,
                                         service = resolvedService,
                                         executionContext = executionContext,
                                     )
                                 } catch (e: Throwable) {
                                     when (e) {
-                                        is GraphQLError -> newServiceExecutionErrorResult(field, error = e)
+                                        is GraphQLError -> newServiceExecutionErrorResult(fields.first(), error = e)
                                         else -> throw e
                                     }
                                 }
@@ -260,7 +260,7 @@ internal class NextgenEngine(
     ): ServiceExecutionResult {
         return try {
             executeTopLevelField(
-                topLevelField = topLevelField,
+                topLevelField = listOf(topLevelField),
                 service = service,
                 executionContext = executionContext.copy(
                     hydrationDetails = hydrationDetails,
@@ -283,7 +283,7 @@ internal class NextgenEngine(
         executionContext: NadelExecutionContext,
     ): ServiceExecutionResult {
         return executeTopLevelField(
-            topLevelField = topLevelField,
+            topLevelField = listOf(topLevelField),
             service = service,
             executionContext = executionContext.copy(
                 isPartitionedCall = true,
@@ -292,7 +292,7 @@ internal class NextgenEngine(
     }
 
     private suspend fun executeTopLevelField(
-        topLevelField: ExecutableNormalizedField,
+        topLevelField: List<ExecutableNormalizedField>,
         service: Service,
         executionContext: NadelExecutionContext,
     ): ServiceExecutionResult {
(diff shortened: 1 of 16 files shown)


lib/src/main/java/graphql/nadel/engine/plan/NadelExecutionPlanFactory.kt
```diff
@@ -60,14 +60,16 @@ internal class NadelExecutionPlanFactory(
         serviceExecutionContext: NadelServiceExecutionContext,
         services: Map<String, Service>,
         service: Service,
-        rootField: ExecutableNormalizedField,
+        rootField: List<ExecutableNormalizedField>,
```
gnawf: Should be `rootFields`

lib/src/main/java/graphql/nadel/engine/plan/NadelExecutionPlanFactory.kt
```diff
@@ -87,7 +89,9 @@ internal class NadelExecutionPlanFactory(
                                     executionBlueprint,
                                     services,
                                     service,
-                                    rootField,
+                                    // buildContext is memoized once per service call; the first root
+                                    // field is passed as the representative root field of the batch.
+                                    rootField.first(),
```
gnawf: I think you should pass in the `rootFields` into here.

lib/src/main/java/graphql/nadel/engine/transform/query/NadelFieldToService.kt
```diff
@@ -27,25 +27,76 @@ internal class NadelFieldToService(
...
+            val batchedFields = batchedByService[service]
+            if (batchedFields != null && topLevelField in batchedFields) {
+                if (emittedBatches.add(service)) {
+                    result += NadelFieldAndService(field = batchedFields, service = service)
+                }
             } else {
-                listOf(getServicePairFor(field = topLevelField))
+                result += NadelFieldAndService(field = listOf(topLevelField), service = service)
             }
         }
```
gnawf, line 57: Can you add an overload to `NadelBatchRootFieldsHint` so we can check enablement globally – not per service.

This allows us to feature flag the code.

So you can run

```kotlin
fun getServicesForTopLevelFields(
    query: ExecutableNormalizedOperation,
    executionHints: NadelExecutionHints,
): List<NadelFieldAndService> {
    if (executionHints.batchRootFields()) {
        // new behavior
    } else {
        // old behavior
    }
}
```

lib/src/main/java/graphql/nadel/engine/transform/query/NadelFieldToService.kt
```diff
@@ -27,25 +27,76 @@ internal class NadelFieldToService(
...
+                batchedByService.getOrPut(service) { mutableListOf() }.add(topLevelField)
+            }
+        }
+
+        // Emit entries in query order. A batched service's root fields are emitted as a single
+        // entry (one service call) at the position of the service's first batchable root field;
+        // everything else keeps its own single-field entry, exactly as before.
+        val result = mutableListOf<NadelFieldAndService>()
+        val emittedBatches = HashSet<Service>()
+        for (topLevelField in topLevelFields) {
```
gnawf: I'm not sure this second for loop is necessary? The order doesn't matter.

lib/src/main/java/graphql/nadel/NextgenEngine.kt
gnawf: I think we should pass in `operation.topLevelFields` instead of `fields` here to preserve the original order.

I checked and the code never accesses `NadelFieldAndService.service` anyway.

lib/src/main/java/graphql/nadel/NextgenEngine.kt
```diff
@@ -292,7 +292,7 @@ internal class NextgenEngine(
     }
 
     private suspend fun executeTopLevelField(
-        topLevelField: ExecutableNormalizedField,
+        topLevelField: List<ExecutableNormalizedField>,
```
gnawf: Name should be plural.

gnawf (commented):
Good start, just update `NadelFieldToService` so it is feature flagged as per my comments 😄

lib/src/main/java/graphql/nadel/engine/transform/NadelTransformJavaCompat.kt
```diff
@@ -122,7 +122,8 @@ interface NadelTransformJavaCompat<State : Any> {
                         executionBlueprint,
                         services,
                         service,
-                        rootField,
+                        // The Java compat API predates root-field batching and takes a single root field.
+                        rootFields.first(),
```
gnawf: Can you pls propagate the change down to here too, needs to be list. We can't use .first() anywhere

rboustany-crypto: Ahh I understand

lib/src/main/java/graphql/nadel/engine/transform/query/NadelFieldToService.kt
```diff
@@ -110,7 +148,7 @@ internal class NadelFieldToService(
 }
 
 data class NadelFieldAndService(
-    val field: ExecutableNormalizedField,
+    val field: List<ExecutableNormalizedField>,
```
gnawf: `fields` plural

gnawf (approved):

PR #734, opened 2026-07-08
Fix potential race conditions

gnawf:


lib/src/main/java/graphql/nadel/time/NadelInternalLatencyTrackerImpl.kt
```diff
@@ -69,15 +76,19 @@ open class NadelInternalLatencyTrackerImpl(
         private val closed = AtomicBoolean(false)
 
         init {
-            if (outstandingExternalLatencyCount.getAndIncrement() == 0) {
-                internalLatency.stop()
+            synchronized(lock) {
+                if ((++outstandingExternalLatencyCount) == 1) {
+                    internalLatency.stop()
```
gnawf, line 81: In theory stopwatch could be left in corrupt state before because the increment and stopwatch actions weren't atomic together.

lib/src/main/java/graphql/nadel/time/NadelInternalLatencyTrackerImpl.kt
```diff
@@ -47,10 +49,15 @@ open class NadelInternalLatencyTrackerImpl(
...
-            .whenComplete { _, _ ->
-                call.close()
-            }
+        try {
+            return future.get()
+                .whenComplete { _, _ ->
+                    call.close()
+                }
+        } catch (e: Throwable) {
+            call.close()
```
gnawf, line 58: In theory `future.get( )` could have thrown and the external call counter would be invalid.

PR #739, opened 2026-07-21
Support Boolean fields in conditional hydration

mcaliskan2:
## Summary

Nadel now accepts Boolean source fields for conditional hydration.

The `equals` predicate accepts `true` and `false`. The `startsWith` and `matches` predicates still require a String or ID field.

Nadel reports a validation error when the equals value is not a Boolean.

## Tests

Validation tests cover nullable and non-null Boolean fields. They also cover list fields, invalid values, and unsupported predicates.

An integration test confirms that Nadel runs the matching hydration and skips the other hydration.

### DIFF
--- a/lib/src/test/kotlin/graphql/nadel/validation/NadelHydrationWhenConditionValidationTest.kt
+++ b/lib/src/test/kotlin/graphql/nadel/validation/NadelHydrationWhenConditionValidationTest.kt
@@ -10,146 +10,150 @@ private const val argument = "$" + "argument"
 
 class NadelHydrationWhenConditionValidationTest {
     private fun booleanConditionFixture(
-        conditionFieldType: String = "Boolean",
-        underlyingConditionFieldType: String = "Boolean!",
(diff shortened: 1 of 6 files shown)


lib/src/test/kotlin/graphql/nadel/validation/NadelHydrationWhenConditionValidationTest.kt
```diff
@@ -9,6 +9,71 @@ private const val source = "$" + "source"
 private const val argument = "$" + "argument"
 
 class NadelHydrationWhenConditionValidationTest {
+    private fun booleanConditionFixture(
+        conditionFieldType: String = "Boolean",
+        underlyingConditionFieldType: String = "Boolean!",
+        predicate: String = "equals: true",
```
gnawf: Can we remove the defaults for clarity so we can see exactly what each test is working with.

And change the enum one while you're at it pls 😄

gnawf (approved):

PR #752, opened 2026-09-07
Expose all root fields for a service execution

dpathak-atl:
## Summary

- Expose the overall-schema root fields grouped into each service execution.
- Capture the fields before query transformation while preserving the existing transformed representative field.
- Keep fields isolated by service and sharding target.

## Compatibility

This is an additive metadata API. It does not change field grouping, query transformation, downstream execution, or result merging. Existing consumers of the singular `executableNormalizedField` remain unchanged.

## Testing

- Added coverage for batched and non-batched services.
- Added coverage for separate sharding targets.
- Added a renamed root field and `__typename` coverage.
- Ran `./gradlew assemble check --info` successfully.


dpathak-atl (commented):
Expose the overall-schema root fields grouped into each service execution.
Capture the fields before query transformation while preserving the existing transformed representative field.
Keep fields isolated by service and sharding target.

PR #753, opened 2026-09-10
Updated Nadel to 0.0.0-2026-08-23T10-34-47-acc4678

bbakerman:
also added so it grabs a per request query limit if provided


lib/src/main/java/graphql/nadel/Nadel.kt
```diff
@@ -168,7 +173,11 @@ class Nadel private constructor(
             executionInputRef.set(executionInput)
 
             logNotSafe.debug("Validating query: '{}'", query)
-            val errors = validate(executionInput, document, graphQLSchema, instrumentationState)
+            val errors = try {
+                validate(executionInput, document, graphQLSchema, instrumentationState)
+            } catch (e: GoodFaithIntrospectionExceeded) {
```
gnawf, line 183: Why does GraphQL Java throw here instead of return an error …

lib/src/main/java/graphql/nadel/Nadel.kt
```diff
@@ -168,7 +173,11 @@ class Nadel private constructor(
...
+                return PreparsedDocumentEntry(document, listOf(e.toBadFaithError()))
```
gnawf: This is different than below. This returns both document and the list of errors.

Below we only ever return `errors` or `document`

What's the difference?

lib/src/test/kotlin/graphql/nadel/NadelTest.kt
gnawf, line 1: Convert these to IT instead using `NadelIntegrationTest`

gnawf (changes requested):

gnawf (dismissed):

gnawf (approved):

